PR #8, opened 2015-04-14
Cleanup tests

garnett:
Cleaned some tests from boilerplate, moved tests fixtures to ```-setUp```

### DIFF
--- a/ResearchKitTests/ORKAccessibilityTests.m
+++ b/ResearchKitTests/ORKAccessibilityTests.m
@@ -39,11 +39,9 @@
 
 - (void)setUp {
     [super setUp];
-    // Put setup code here. This method is called before the invocation of each test method in the class.
 }
 
 - (void)tearDown {
-    // Put teardown code here. This method is called after the invocation of each test method in the class.
     [super tearDown];
 }
 
--- a/ResearchKitTests/ORKAccessibilityTests.m
+++ b/ResearchKitTests/ORKAccessibilityTests.m
@@ -37,14 +37,6 @@
 
 @implementation ORKAccessibilityTests
 
-- (void)setUp {
-    [super setUp];
-}
-
-- (void)tearDown {
-    [super tearDown];
-}
-
 - (void)testAccessibilityString {
     
     {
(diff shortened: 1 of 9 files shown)


ResearchKitTests/ORKAccessibilityTests.m
```diff
@@ -39,11 +39,9 @@ @implementation ORKAccessibilityTests
 
 - (void)setUp {
     [super setUp];
-    // Put setup code here. This method is called before the invocation of each test method in the class.
 }
```
Keithbsmiley: Could probably remove all these calls too, no need to call them if nothing else is being done.

modocache: +1. Empty overrides add a lot of noise with no benefit.

garnett: Agree with you, updated the pull-request.

YuanZhu-apple: Agree to remove empty overrides.

ResearchKitTests/ORKHKSampleTests.m
```diff
@@ -43,46 +46,34 @@ @implementation ORKHKSampleTests
...
     [super setUp];
-    // Put setup code here. This method is called before the invocation of each test method in the class.
-}
-
-- (void)tearDown
-{
-    // Put teardown code here. This method is called after the invocation of each test method in the class.
-    [super tearDown];
+    _d1 = [NSDate dateWithTimeIntervalSinceReferenceDate:0];
+    _d2 = [NSDate dateWithTimeInterval:10 sinceDate:_d1];
```
modocache: This should probably be in `-[ORKHKSampleTests setUp]`, not `-tearDown`. Do the tests still pass with these changes?

Also, IMHO the tests are more readable with `d1` and `d2` local to each test method, not in the `-setUp` method. To each his own, though! :point_up:

garnett: But they are in ```setUp```...

modocache: Oh, whoops! Misread the diff. My bad! :open_mouth:

jwe-apple:
Yeah, actually agree with @modocache on readability. The whitespace / boilerplate cleanup is fine, but let's revert the switch to using ivars.

jwe-apple:
Yep, now it's just whitespace removal and removing unused overrides, looks good.

PR #10, opened 2015-04-14
Added Dutch localization

BasThomas:


jwe-apple:
Hi, we already have a Dutch localization! What is this patch intended to address?

BasThomas:
It's in the more readable .strings format, which should make it easier to edit and enhace it in the future.

jwe-apple:
Actually not a bad idea, but we should do that globally instead of a file at a time. I opened https://github.com/ResearchKit/ResearchKit/issues/29 to track. I'm going to close this particular pull request.

johankool:
You missed LEARN_MORE_STUDY_SURVEY

PR #22, opened 2015-04-15
make it clear myStep is a variable

waynn:


DanKeen:
Thanks for catching that!

waynn:
:+1:

PR #23, opened 2015-04-15
Objects that conform to NSCopying should use copy instead of strong

fyaqub:
This is to prevent immutable objects from being passed their mutable counterparts and having the object's attribute changed unintentionally. 

For example, passing an `NSMutableString` object to an `NSString` object, then changing the `NSMutableString` object would result in the `NSString` object's value being changed if `strong` is used. Using `copy` would prevent that from happening.

### DIFF
--- a/ResearchKit/ActiveTasks/ORKAudioContentView.m
+++ b/ResearchKit/ActiveTasks/ORKAudioContentView.m
@@ -47,7 +47,7 @@ static const CGFloat GraphViewRedZoneHeight = 25;
 @property (nonatomic, strong) UIColor *keyColor;
 @property (nonatomic, strong) UIColor *alertColor;
 
-@property (nonatomic, copy) NSArray *values;
+@property (nonatomic, strong) NSArray *values;
 
 @property (nonatomic) CGFloat alertThreshold;
 
--- a/ResearchKit/ActiveTasks/ORKAudioContentView.m
+++ b/ResearchKit/ActiveTasks/ORKAudioContentView.m
@@ -47,7 +47,7 @@ static const CGFloat GraphViewRedZoneHeight = 25;
 @property (nonatomic, strong) UIColor *keyColor;
 @property (nonatomic, strong) UIColor *alertColor;
 
-@property (nonatomic, strong) NSArray *values;
+@property (nonatomic, copy) NSArray *values;
 
 @property (nonatomic) CGFloat alertThreshold;
 
--- a/ResearchKit/ActiveTasks/ORKHealthQuantityTypeRecorder.m
+++ b/ResearchKit/ActiveTasks/ORKHealthQuantityTypeRecorder.m
@@ -46,8 +46,6 @@
     HKQuantitySample *_lastSample;
 }
 
-@property (nonatomic, copy) NSError *recordingError;
-
 @end
 
 
--- a/ResearchKit/ActiveTasks/ORKHealthQuantityTypeRecorder.m
+++ b/ResearchKit/ActiveTasks/ORKHealthQuantityTypeRecorder.m
@@ -46,7 +46,7 @@
     HKQuantitySample *_lastSample;
 }
 
-@property (nonatomic, strong) NSError *recordingError;
+@property (nonatomic, copy) NSError *recordingError;
 
 @end
 
--- a/ResearchKit/ActiveTasks/ORKPedometerRecorder.m
+++ b/ResearchKit/ActiveTasks/ORKPedometerRecorder.m
@@ -41,7 +41,7 @@
 }
 
 @property (nonatomic, strong) CMPedometer *pedometer;
-@property (nonatomic, strong) NSError *recordingError;
+@property (nonatomic, copy) NSError *recordingError;
 
 @end
 
--- a/ResearchKit/ActiveTasks/ORKPedometerRecorder.m
+++ b/ResearchKit/ActiveTasks/ORKPedometerRecorder.m
@@ -41,7 +41,6 @@
 }
 
 @property (nonatomic, strong) CMPedometer *pedometer;
-@property (nonatomic, copy) NSError *recordingError;
 
 @end
 
--- a/ResearchKit/Common/ORKAnswerFormat.h
+++ b/ResearchKit/Common/ORKAnswerFormat.h
@@ -319,7 +319,7 @@ ORK_CLASS_AVAILABLE
  
  Note that the `detailText` property of each choice is ignored. Be sure to create localized text for each choice that is short enough to fit in a `UIPickerView` object.
  */
-@property (readonly, copy) NSArray *textChoices;
+@property (readonly, strong) NSArray *textChoices;
 
 @end
 
@@ -349,7 +349,7 @@ ORK_CLASS_AVAILABLE
  The text of the currently selected choice is displayed on screen. The text for
  each choice is spoken by VoiceOver when an image is highlighted.
  */
-@property (readonly, copy) NSArray *imageChoices;
+@property (readonly, strong) NSArray *imageChoices;
 
 @end
 
@@ -387,7 +387,7 @@ ORK_CLASS_AVAILABLE
  The text for each answer is given more prominence than the `detailText` in the row, but
  both are shown.
  */
-@property (readonly, copy) NSArray *textChoices;
+@property (readonly, strong) NSArray *textChoices;
 
 @end
 
--- a/ResearchKit/Common/ORKAnswerFormat.h
+++ b/ResearchKit/Common/ORKAnswerFormat.h
@@ -319,7 +319,7 @@ ORK_CLASS_AVAILABLE
  
  Note that the `detailText` property of each choice is ignored. Be sure to create localized text for each choice that is short enough to fit in a `UIPickerView` object.
  */
-@property (readonly, strong) NSArray *textChoices;
+@property (readonly, copy) NSArray *textChoices;
 
 @end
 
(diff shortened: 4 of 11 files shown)


DanKeen:
While this is true for `NSMutableString` and `NSString`, it's not really true for `NS{Mutable,}Array` or other collection objects. `-copy` doesn't do a deep copy, and so while you may think you are receiving a "clean" collection, any of the items in the collection are still free to mutate. Because of that I generally think it's poor form to do a copy on a collection and instead have the property declaration remind the user that the collection and its contents may be mutated elsewhere.

Can you resubmit with the collections kept as `strong`?

fyaqub:
Absolutely, will submit an update in a few minutes. Thanks!

fyaqub:
I switched `NSArray` objects back to `copy`. Let me know if you want any of the other types switched back as well.

sikhapol:
@DanKeen While it's true that each item in the collection is free to mutate, isn't it safer to at least have the non-readonly collection to be `copy`? In this case, if someone assigns an `NSMutableArray` to `values` of `ORKAudioContentView`, at least we are a little safer to use `values` without having to worry that `values` could be mutate elsewhere by the client later.
Also this might be why `viewControllers` property of `UITabBarController` is also a `copy`.

fyaqub:
I switched NSArray objects back to copy. Let me know if you want any of the other types switched back as well. Thanks!

ResearchKit/ActiveTasks/ORKAudioContentView.m
```diff
@@ -47,7 +47,7 @@ @interface ORKAudioGraphView : UIView
 @property (nonatomic, strong) UIColor *keyColor;
 @property (nonatomic, strong) UIColor *alertColor;
 
-@property (nonatomic, copy) NSArray *values;
+@property (nonatomic, strong) NSArray *values;
```
jwe-apple: I believe this was appropriately copy. Calling code passing a mutable array of NSValue could subsequently append to that array.

ResearchKit/ActiveTasks/ORKPedometerRecorder.m
```diff
@@ -41,7 +41,7 @@ @interface ORKPedometerRecorder()
 }
 
 @property (nonatomic, strong) CMPedometer *pedometer;
-@property (nonatomic, strong) NSError *recordingError;
+@property (nonatomic, copy) NSError *recordingError;
```
jwe-apple: This property appears to be unused. Might as well remove it?

ResearchKit/ActiveTasks/ORKHealthQuantityTypeRecorder.m
```diff
@@ -46,7 +46,7 @@ @interface ORKHealthQuantityTypeRecorder()
     HKQuantitySample *_lastSample;
 }
 
-@property (nonatomic, strong) NSError *recordingError;
+@property (nonatomic, copy) NSError *recordingError;
```
jwe-apple: Appears to be unused. Should remove instead.

ResearchKit/Common/ORKAnswerFormat.h
```diff
@@ -319,7 +319,7 @@ ORK_CLASS_AVAILABLE
  
  Note that the `detailText` property of each choice is ignored. Be sure to create localized text for each choice that is short enough to fit in a `UIPickerView` object.
  */
-@property (readonly, copy) NSArray *textChoices;
+@property (readonly, strong) NSArray *textChoices;
```
jwe-apple: This is intentionally copy, although it's irrelevant as API because it's a readonly property. Do not change.

ResearchKit/Common/ORKAnswerFormat.h
```diff
@@ -349,7 +349,7 @@ ORK_CLASS_AVAILABLE
  The text of the currently selected choice is displayed on screen. The text for
  each choice is spoken by VoiceOver when an image is highlighted.
  */
-@property (readonly, copy) NSArray *imageChoices;
+@property (readonly, strong) NSArray *imageChoices;
```
jwe-apple: This is intentionally copy, although it's irrelevant as API because it's a readonly property. Do not change.

ResearchKit/Common/ORKAnswerFormat.h
```diff
@@ -387,7 +387,7 @@ ORK_CLASS_AVAILABLE
  The text for each answer is given more prominence than the `detailText` in the row, but
  both are shown.
  */
-@property (readonly, copy) NSArray *textChoices;
+@property (readonly, strong) NSArray *textChoices;
```
jwe-apple: This is intentionally copy, although it's irrelevant as API because it's a readonly property. Do not change.

ResearchKit/Common/ORKCountdownLabel.m
```diff
@@ -34,8 +34,8 @@ OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 @interface ORKCountdownLabel ()
 
 
-@property (nonatomic, strong) NSString *mmString;
-@property (nonatomic, strong) NSString *ssString;
+@property (nonatomic, copy) NSString *mmString;
+@property (nonatomic, copy) NSString *ssString;
```
jwe-apple: These are fine, although technically irrelevant, since the setters are never called on these internal properties.

ResearchKit/ActiveTasks/ORKAudioContentView.m
fyaqub: My initial patch changed it from `strong` to `copy, then @DanKeen requested changing it back to `strong`. See his comment above.

ResearchKit/Common/ORKDateTimePicker.m
```diff
@@ -35,7 +35,7 @@ OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 @interface ORKDateTimePicker ()
 
 @property (nonatomic, strong) ORKAnswerFormat *answerFormat;
-@property (nonatomic, strong) NSCalendar *calendar;
+@property (nonatomic, copy) NSCalendar *calendar;
```
jwe-apple: Even though NSCalendar implements NSCopying, I don't think it's necessarily a great idea to copy one. For instance an autoupdating calendar passed in might be exactly what's intended.

fyaqub:
@jwe-apple I pushed an update with the changes you commented about.

Regarding `@property (nonatomic, strong) NSArray *values;`, you guys originally used `strong`, I changed it to `copy`, then changed it back to `strong` per @DanKeen's request.

ResearchKit/ActiveTasks/ORKAudioContentView.m
jwe-apple: I believe his comment referred to a different place, but I could be wrong. This being copy is intended and is part of the original codebase.

fyaqub: You're correct. When I made the changes based on this comments, I didn't realize that one `NSArray` property was already intentionally set to `copy` on your end.

Submitted a change to set it back to `copy` :)

ResearchKit/Common/ORKHTMLPDFWriter.m
```diff
@@ -89,8 +89,8 @@ @interface ORKHTMLPDFWriter () <UIWebViewDelegate>
 @property (nonatomic) CGSize pageSize;
 @property (nonatomic) UIEdgeInsets pageMargins;
 @property (nonatomic, strong) UIWebView *webView;
-@property (nonatomic, strong) NSData *data;
-@property (nonatomic, strong) NSError *error;
+@property (nonatomic, copy) NSData *data;
+@property (nonatomic, copy) NSError *error;
```
jwe-apple: NSError - ok.
NSData - less sure. This is an internal property of this method, which constructs mutable data and then sets it as the value of the property. Copying the mutable data is unnecessary, since we know we will never mutate it further after setting it.

Admittedly we never call the setter, since the ivar is set directly, but it's still not obvious this shouldn't just be strong.

ResearchKit/Consent/ORKVisualConsentTransitionAnimator.m
```diff
@@ -47,7 +47,7 @@ @interface ORKVisualConsentAnimationContext : NSObject
 @property (nonatomic, copy) ORKVisualConsentAnimationCompletionHandler handler;
 @property (nonatomic, copy) ORKVisualConsentAnimationCompletionHandler loadHandler;
 
-@property (nonatomic, strong) NSValue *startTime;
+@property (nonatomic, copy) NSValue *startTime;
```
jwe-apple: There's no mutable version of NSValue, so isn't this irrelevant?

fyaqub: This was more about consistency, to just use `copy` as the default for things that conform to `NSCopying`.

fyaqub:
@jwe-apple I've submitted all changes based on your comments. Let me know if there's anything else from my end. Thanks!

jwe-apple:
@fyaqub I think I may have miscommunicated. The changes in ORKAnswerFormat.h need reverting - they should be marked copy for consistency with the rest of the public API.

fyaqub:
@jwe-apple Cool, changes have been reverted in `ORKAnswerFormat.h`.

jwe-apple:
@fyaqub Thanks for your work so far, bear with us, this will likely get merged soon!

YuanZhu-apple:
@fyaqub Can you send an email to researchkit@apple.com, so we can send you a copy of "ResearchKit Contribution Terms and License Policy". Thanks!

fyaqub:
@YuanZhu-apple Just sent an email, thanks!

PR #27, opened 2015-04-15
Code of Conduct

beanieboi:
this adds a basic Code of Conduct.

thanks for open sourcing this!


YuanZhu-apple:
Thanks for the suggestion! But we will wait until we have a problem before addressing this.

PR #28, opened 2015-04-15
Added logo image of ResearchKit

codestergit:
Added Logo image of ResearchKit in  README.md


YuanZhu-apple:
We are happy not having a logo here.

PR #30, opened 2015-04-15
Replaced magic numbers in OKHTMLPDFWriter.m

franklinsch:


jwe-apple:
Good idea, but use static const instead for type safety.

franklinsch:
Good idea. I updated the file

jwe-apple:
Sorry to be a stickler, but these are passed to CGSize so should be CGFloat (different behavior 64 vs 32-bit)

franklinsch:
Woops, sorry about that! Updated.

jwe-apple:
Thanks!

Issue #3, opened 2015-04-10
Vertical scale


YuanZhu-apple:
Currently all our scales are horizontal. Some researchers want to have a vertical one.

rsanchezsaez:
Implemented.

jwe-apple:
Yep, this is done.

Issue #2, opened 2015-04-10
ORKRecorderConfiguration and ORKRecorder should have an identifier 


YuanZhu-apple:
The aim is to let recorder generated result object has a meaningful identifier which can be used to link back to the `ORKRecorderConfiguration` object.

* The identifier will be passed from `ORKRecorderConfiguration` to `ORKRecorder`.
* The identifier is like `ORKStep` identifier which is supplied by developer.

rsanchezsaez:
Hi guys,

I was going to implement this but I have some questions. Could you clarify this issue?

Currently `ORKRecorder` assigns its `ORKStep` `identifier` to the `ORKFileResult` `identifier`.

Do you need a new (independent from `ORKStep`) `identifier` property in `ORKRecorderConfiguration` and `ORKRecorder` that would override the `ORKStep` `identifier` on the `ORKFileResult`? Or would you rather add a property like `recorderIdentifier` to the `ORKFileResult` class so both the `ORKStep` and the `ORKRecorderConfiguration` links are preserved?

Thanks.

jwe-apple:
The ORKFileResult generated by a recorder should have the identifier of the recorder that generated it. The current implementation happens only because there is no per-recorder identifier.
The ORKFileResult is always reported as a child of an ORKStepResult, so there is no loss of information, similar to how ORKStepResult does not include ORKTaskResult's identifier.

ORKRecorderConfiguration definitely will need an identifier property, and appropriate changes to initializers, since it will be a readonly property to match behavior elsewhere.

If doing this, make sure you run the ORKTest app and ensure the serialization tests pass after you do. You'll likely need to add an entry for the new property in the ORKESerialization implementation, following the pattern of what's there already.

rsanchezsaez:
Cool, all is clear. Will do this shortly.

rsanchezsaez:
I'm making good progress on this: almost everything done. I'm ready to test serialization.

By `ORKTest` do you mean the *XCTest ResearchKitTests*? I cannot find anything with a name similar to `ORKESerialization`. The `ORKResultTests` file has some keyed archiving serialization tests, but the `ORKRecorderTests` file doesn't seem to.

jwe-apple:
No, there is an ORKTest test project in the repo. Look under testing, I think.

rsanchezsaez:
Oh, found it. I didn't realize there was an additional project file. Cheers!

rsanchezsaez:
Done. ;)

Issue #1, opened 2015-04-10
Add image answer format to capture image in a question step


YuanZhu-apple:
Some researchers are interested in developing medical research apps that require access to the camera and save images in high resolution.
* Create an image answer format class which could have a few attributes like:  resolution, flash light, and output format. 
* Design a question step UI interface to capture an image and display it to the user. 
* `ORKFileResult` should be used to return the image file reference.

imanabu:
We are definitely interested in participating in this implementation. We have been developing a medical grade photography application called BodyMapSnap. It captures not just the picture, but also other meta data including the body location where the photo was taken from.

jwe-apple:
In ResearchKit the collection of other metadata would happen either by incorporating the image answer format in a form item on a form step, or by incorporating an image answer format in a question step within a larger task which might incorporate those other questions.

We would love to have you contribute an implementation. The first step is probably an engineering mockup of how you propose it would look in both the form and question step scenarios, and then an API proposal for the answer format itself.

brucehappy:
I've been looking at this as well, and should be able to upload a few screenshots of a mockup for form and question steps very soon.

imanabu:
Sorry about apparent inactivity, but we have been working with other NIH researcher to see what would best way to address this. There are some medical specific questions we can ask the participants and also how the photography is integrated, and from our standpoint, what will happen to the images, where they are stored. I look forward to see the mockup as well.

jwe-apple:
Assigned to @brucehappy as further progress sounds imminent.

brucehappy:
In a form step, I've shown a few approaches to the image capture answer.  The first has the text being a section header, and the latter two have the text being within the cell, on the left, as in other cell types (I added the "INLINE CONTROLS" section to separate them from the first one).  I'm showing a few different sizes of the control in this form, as well.  I thought that the size of the control could be something set in the answer format, and then perhaps based on that size the form cell could decide dynamically whether to use a section header for the text, or include it in the cell.  I've also included a question step example too.  (The first two pictures below are of the same screen, it is just scrolled down in the second picture)

[...]

PR #33, opened 2015-04-15
Add vertical scale answer format

rsanchezsaez:
I have implemented vertical scale support ([issue #3](https://github.com/ResearchKit/ResearchKit/issues/3)).

`ORKScaleSlider` now has the `isVertical` property. When set to `YES`, it internally uses the `UIView`'s`transform` property and some minor *bounds* and *touch* trickery to draw and work in vertical.

`ORKScaleSliderView` now has vertical support through the use of the new `isVertical` method from the `ORKScaleAnswerFormatProvider`.

New classes: `ORKVerticalScaleAnswerFormat` and `ORKContinuousVerticalScaleAnswerFormat`. The corresponding example steps have been added to `ORKCatalog`'s *Scale Question*.

### DIFF
--- a/ResearchKit/Common/ORKAnswerFormat.h
+++ b/ResearchKit/Common/ORKAnswerFormat.h
@@ -44,9 +44,6 @@ typedef NS_ENUM(NSInteger, ORKQuestionType) {
     
     /// The scale question type asks participants to place a mark at an appropriate position on a continuous or discrete line.
     ORKQuestionTypeScale,
-    
-    /// The vertical scale question type asks participants to place a mark at an appropriate position on a continuous or discrete vertical line.
-    ORKQuestionTypeVerticalScale,
 
     /// In a single choice question, the participant can pick only one predefined option.
     ORKQuestionTypeSingleChoice,
--- a/ResearchKit/Common/ORKAnswerFormat.h
+++ b/ResearchKit/Common/ORKAnswerFormat.h
@@ -141,10 +141,20 @@ ORK_CLASS_AVAILABLE
                                                    step:(NSInteger)step
                                            defaultValue:(NSInteger)defaultValue;
 
++ (ORKScaleAnswerFormat *)verticalScaleAnswerFormatWithMaxValue:(NSInteger)scaleMax
+                                                       minValue:(NSInteger)scaleMin
+                                                           step:(NSInteger)step
+                                                   defaultValue:(NSInteger)defaultValue;
+
 + (ORKContinuousScaleAnswerFormat *)continuousScaleAnswerFormatWithMaxValue:(double)scaleMax
                                                                    minValue:(double)scaleMin
                                                                defaultValue:(double)defaultValue
                                                       maximumFractionDigits:(NSInteger)maximumFractionDigits;
+
++ (ORKContinuousScaleAnswerFormat *)continuousVerticalScaleAnswerFormatWithMaxValue:(double)scaleMax
+                                                                           minValue:(double)scaleMin
+                                                                       defaultValue:(double)defaultValue
+                                                              maximumFractionDigits:(NSInteger)maximumFractionDigits;
 
 + (ORKBooleanAnswerFormat *)booleanAnswerFormat;
 
@@ -179,8 +189,8 @@ ORK_CLASS_AVAILABLE
 + (ORKTimeIntervalAnswerFormat *)timeIntervalAnswerFormat;
 + (ORKTimeIntervalAnswerFormat *)timeIntervalAnswerFormatWithDefaultInterval:(NSTimeInterval)defaultInterval step:(NSInteger)step;
 
-/// @name Validation
 
+/// @name Validation
 
 /**
  Validates the parameters of the answer format to ensure that they can be displayed.
@@ -211,7 +221,7 @@ ORK_CLASS_AVAILABLE
  
  @param maximumValue   The upper bound of the scale.
  @param minimumValue   The lower bound of the scale.
- @param step  The size of each discrete offset on the scale.
+ @param step   The size of each discrete offset on the scale.
  @param defaultValue   The default value of the scale. If this value is out of range, the slider is displayed without a default value.
  @return An initialized scale answer format.
  */
@@ -229,7 +239,6 @@ ORK_CLASS_AVAILABLE
  The lower bound of the scale. (read-only)
  */
 @property (readonly) NSInteger minimum;
-
 
 /**
  The size of each discrete offset on the scale. (read-only)
@@ -250,6 +259,21 @@ ORK_CLASS_AVAILABLE
 
 @end
 
+
+
+/**
+ The `ORKVerticalScaleAnswerFormat `class represents an answer format that includes a vertical slider control.
+ 
+ The vertical scale answer format produces an `ORKScaleQuestionResult` object that contains an integer whose
+ value is between the scale's minimum and maximum values, and represents one of the quantized step values.
+ */
+ORK_CLASS_AVAILABLE
+@interface ORKVerticalScaleAnswerFormat : ORKScaleAnswerFormat
+
+@end
+
+
+
 /**
  The `ORKContinuousScaleAnswerFormat` class represents an answer format that lets participants
  select a value on a continuous scale.
@@ -259,7 +283,17 @@ ORK_CLASS_AVAILABLE
 ORK_CLASS_AVAILABLE
 @interface ORKContinuousScaleAnswerFormat : ORKAnswerFormat
 
-
+/**
+ Returns an initialized continous scale answer format using the specified values.
+ 
+ This method is the designated initializer.
+ 
+ @param maximumValue   The upper bound of the scale.
+ @param minimumValue   The lower bound of the scale.
+ @param defaultValue   The default value of the scale. If this value is out of range, the slider is displayed without a default value.
+ @param maximumFractionDigits    The maximum number of fractional digits to display.
+ @return An initialized scale answer format.
+ */
 - (instancetype)initWithMaximumValue:(double)maximumValue
                         minimumValue:(double)minimumValue
                         defaultValue:(double)defaultValue
@@ -288,6 +322,21 @@ ORK_CLASS_AVAILABLE
 @property (readonly) NSInteger maximumFractionDigits;
 
 @end
+
+
+
+/**
+ The `ORKContinuousVerticalScaleAnswerFormat` class represents an answer format that lets participants
+ select a value on a continuous vertical scale.
+ 
+ The continuous vertical scale answer format produces an `ORKScaleQuestionResult` object that has a real-number value.
+ */
+ORK_CLASS_AVAILABLE
+@interface ORKContinuousVerticalScaleAnswerFormat : ORKContinuousScaleAnswerFormat
+
+@end
+
+
 
 /**
  The `ORKValuePickerAnswerFormat` class represents an answer format that lets participants use a value picker 
@@ -323,6 +372,8 @@ ORK_CLASS_AVAILABLE
 
 @end
 
+
+
 /**
  The `ORKImageChoiceAnswerFormat` class represents an answer format that lets participants choose one image from a fixed set of images in a single choice question.
  
@@ -352,6 +403,8 @@ ORK_CLASS_AVAILABLE
 @property (readonly, copy) NSArray *imageChoices;
 
 @end
+
+
 
 /**
  The `ORKTextChoiceAnswerFormat` class represents an answer format that lets participants choose from a fixed set of text choices in a multiple or single choice question.
--- a/ResearchKit/Common/ORKAnswerFormat.h
+++ b/ResearchKit/Common/ORKAnswerFormat.h
@@ -136,25 +136,17 @@ ORK_CLASS_AVAILABLE
 
 /// @name Factory methods
 
-+ (ORKScaleAnswerFormat *)scaleAnswerFormatWithMaxValue:(NSInteger)scaleMax
-                                               minValue:(NSInteger)scaleMin
-                                                   step:(NSInteger)step
-                                           defaultValue:(NSInteger)defaultValue;
++ (ORKScaleAnswerFormat *)scaleAnswerFormatWithMaximumValue:(NSInteger)scaleMaximum
+                                               minimumValue:(NSInteger)scaleMinimum
+                                               defaultValue:(NSInteger)defaultValue
+                                                       step:(NSInteger)step
+                                                   vertical:(BOOL)verticalFlag;
 
-+ (ORKScaleAnswerFormat *)verticalScaleAnswerFormatWithMaxValue:(NSInteger)scaleMax
-                                                       minValue:(NSInteger)scaleMin
-                                                           step:(NSInteger)step
-                                                   defaultValue:(NSInteger)defaultValue;
-
-+ (ORKContinuousScaleAnswerFormat *)continuousScaleAnswerFormatWithMaxValue:(double)scaleMax
-                                                                   minValue:(double)scaleMin
-                                                               defaultValue:(double)defaultValue
-                                                      maximumFractionDigits:(NSInteger)maximumFractionDigits;
-
-+ (ORKContinuousScaleAnswerFormat *)continuousVerticalScaleAnswerFormatWithMaxValue:(double)scaleMax
-                                                                           minValue:(double)scaleMin
-                                                                       defaultValue:(double)defaultValue
-                                                              maximumFractionDigits:(NSInteger)maximumFractionDigits;
++ (ORKContinuousScaleAnswerFormat *)continuousScaleAnswerFormatWithMaximumValue:(double)scaleMaximum
+                                                                   minimumValue:(double)scaleMinimum
+                                                                   defaultValue:(double)defaultValue
+                                                          maximumFractionDigits:(NSInteger)maximumFractionDigits
+                                                                       vertical:(BOOL)verticalFlag;
 
 + (ORKBooleanAnswerFormat *)booleanAnswerFormat;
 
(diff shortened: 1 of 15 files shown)


rsanchezsaez:
I tried to visually layout vertical scales so the scrolling is minimum on *iPhone 4S* sizes, but this produces somewhat short vertical scales. Maybe we should layout to *iPhone 5* sizes and just have *iPhone 4S* users scrolling a bit.

Also, I'm not happy by the fact I had to increase the `ORKSurveyAnswerCellForScale`'s `suggestedCellHeightForView:` (shared by both vertical and horizontal scales) so it fits the vertical ones.

Ideally, the height wouldn't increase for the horizontal scale cell (and vertical ones could be made even taller). But I couldn't figure a simple and sound way of doing this without a major refactor (since the `ORKSurveyAnswerCellForScale` class is inferred from `ORKQuestionTypeScale`, and it doesn't make much sense to add an additional `ORKQuestionTypeVerticalScale` value just for the vertical scales). Any suggestion regarding this issue?

ResearchKit/Common/ORKAnswerFormat.m
```diff
@@ -251,11 +251,16 @@ @implementation ORKAnswerFormat
 + (ORKScaleAnswerFormat *)scaleAnswerFormatWithMaxValue:(NSInteger)scaleMax
                                               minValue:(NSInteger)scaleMin
                                                   step:(NSInteger)step
-                                          defaultValue:(NSInteger)defaultValue
-{
+                                          defaultValue:(NSInteger)defaultValue {
```
3lvis: Better be consistent with the location of the `{` (put it in the next line)

ResearchKit/Common/ORKAnswerFormat.m
```diff
@@ -251,11 +251,16 @@ @implementation ORKAnswerFormat
...
     return [[ORKScaleAnswerFormat alloc] initWithMaximumValue:scaleMax minimumValue:scaleMin step:step defaultValue:defaultValue];
 }
 
++ (ORKScaleAnswerFormat *)verticalScaleAnswerFormatWithMaxValue:(NSInteger)scaleMax
+                                                       minValue:(NSInteger)scaleMin
+                                                           step:(NSInteger)step
+                                                   defaultValue:(NSInteger)defaultValue {
```
3lvis: Better be consistent with the location of the `{` (put it in the next line)

ResearchKit/Common/ORKAnswerFormat.h
```diff
@@ -250,6 +259,21 @@ ORK_CLASS_AVAILABLE
...
+/**
+ The `ORKVerticalScaleAnswerFormat `class represents an answer format that includes a vertical slider control.
+ 
+ The vertical scale answer format produces an `ORKScaleQuestionResult` object that contains an integer whose
+ value is between the scale's minimum and maximum values, and represents one of the quantized step values.
+ */
+ORK_CLASS_AVAILABLE
+@interface ORKVerticalScaleAnswerFormat : ORKScaleAnswerFormat
+
+@end
```
YuanZhu-apple: Would it be much simpler to just add a `vertical` boolean property on  `ORKScaleAnswerFormat` and `ORKContinuousScaleAnswerFormat `?

ResearchKit/Common/ORKSurveyAnswerCellForScale.m
```diff
@@ -119,9 +119,7 @@ - (IBAction)sliderValueChanged:(id)sender {
 }
 
 + (CGFloat)suggestedCellHeightForView:(UIView *)view {
-    return 140.0;
+    return 244.0;
 }
```
YuanZhu-apple: I think better to add a second `vertical` parameter to this method, so it can return different height for different layout.

YuanZhu-apple:
Congratulations on the FIRST new feature PR for ResearchKit!

To the visually layout of vertical scale, scale bar should has an adequate height to assist user to select on the bar easier.
* Value labels should be move to the left to save space for scale bar
* It's OK to have a fixed height across different screen sizes, but it is better to make the page is not scrollable on iPhone 6 and iPhone 6 plus (page scrolling on iphone 4s and iphone 5 is fine). Try this magic cell height: `358.0`.

I did an engineering mockup:
mockup

jwe-apple:
One more comment - to accept this pull request, it will also need to work properly with VoiceOver enabled. The existing horizontal control has VoiceOver support, so this should be straightforward.

ResearchKit/Common/ORKSurveyAnswerCellForScale.m
jwe-apple: Yuan - not sure quite how that would work! This is a class method on the base class, and vertical/horizontal is very specific to this class. It might be easiest to just create a subclass of ORKSurveyAnswerCellForScale to cover the vertical case.

rsanchezsaez:
Hi guys,

thanks for the feedback. I'll look into the design changes. Definitely the end labels on the side of the slider is a good idea that saves some vertical space.

I tried displaying `_valueLabel` to the left of the vertical scale before but it didn't look quite right (all the rest of the answer formats have a nice vertical-axis hierarchical organisation to their UI elements, and I thought this setup kind of broke it). I'll try again and report back.

Regarding the height of `ORKSurveyAnswerCellForScale`, I had the problem that @jwe-apple mentions: it's quite tricky to send parameters to the `suggestedCellHeightForView:` method without a big refactoring. Creating a `ORKSurveyAnswerCellForVerticalScale` subclass may be the best solution. Note that I'll have to add a (otherwise needless) `ORKQuestionTypeVerticalScale` value for this to work.

Regarding adding a `isVertical` property to `ORKScaleAnswerFormat` and ` ORKContinuousScaleAnswerFormat` instead of adding the `ORKVerticalScaleAnswerFormat` and `ORKContinuousVerticalScaleAnswerFormat` classes, I thought the the additional classes in my solution were neater from the framework user perspective (and would allow for more encapsulated visual customization in the subclasses). But tell me if the adding property makes more sense and I'll modify that, this should be pretty straight forward.

jwe-apple:
We've found the answer formats tend to be pretty static and think a property makes sense. Probably read-only, but could be convinced otherwise.

rsanchezsaez:
A read-only property passed on the answer format initializers makes sense. I'll make the changes soon.

Do you prefer `isVertical` or `vertical` for the property names (for `ORKScaleSlider`, `ORKScaleAnswerFormatProvider`, `ORKScaleAnswerFormat`, and `ORKContinuousScaleAnswerFormat`)? Looking at past usages in other Apple code: *iOS*' `UITextSelectionRect` uses `isVertical`, but *OS X*'s `NSSplitView` uses `vertical`.

[Coding Guidelines for Cocoa](https://developer.apple.com/library/mac/documentation/Cocoa/Conceptual/CodingGuidelines/Articles/NamingIvarsAndTypes.html#//apple_ref/doc/uid/20001284-BAJGIIJE) naming conventions seem to recommend using the `vertical` name with a custom specified `isVertical` get accessor as in:

    @property (assign, getter=isEditable) BOOL editable;

jwe-apple:
Yep, the latter

rsanchezsaez:
I have added commits implementing the requested feedback.

Regarding visual layout, I optimized it for iPhone 6 and up. I moved the range labels to the left side of the scale. I left the `valueLabel` at the top center-y of the view, as I feel it looks nicer this way. I think that with the default iPhone 6 cell height, there's enough room for the vertical scale. If you disagree, let me know and I can move the label to center-x left-y of the view.

I think that VoiceOver should work straight away as well (I checked with the Accessibility inspector). I cannot test on an actual device right now, can anybody confirm that it works?

Thanks.

rsanchezsaez:
Here are some iPhone 6 sized screenshots:

vertical scale #1

---

vertical scale #2

---

continuous vertical scale

3lvis:
Kicking ass @rsanchezsaez!

:clap:

avocade:
This is such a great read! And even greater to have Apple engineers on Github—never thought I'd see the day, but I'm glad I was wrong to doubt.

YuanZhu-apple:
One more thing, `CurrentThumbImage` should be rotated to let its drop shadow point downward.

rsanchezsaez:
Excellent catch! Issue solved.

ResearchKit/Common/ORKAnswerFormat.h
```diff
@@ -136,15 +139,17 @@ ORK_CLASS_AVAILABLE
...
 
-+ (ORKContinuousScaleAnswerFormat *)continuousScaleAnswerFormatWithMaxValue:(double)scaleMax
-                                                                   minValue:(double)scaleMin
-                                                               defaultValue:(double)defaultValue
-                                                      maximumFractionDigits:(NSInteger)maximumFractionDigits;
++ (ORKContinuousScaleAnswerFormat *)continuousScaleAnswerFormatWithMaximumValue:(double)scaleMaximum
+                                                                   minimumValue:(double)scaleMinimum
+                                                                   defaultValue:(double)defaultValue
+                                                          maximumFractionDigits:(NSInteger)maximumFractionDigits
+                                                                       vertical:(BOOL)verticalFlag;
```
YuanZhu-apple: Personally, I prefer `vertical` over `verticalFlag`.

YuanZhu-apple:
`ORKTest App`(under `Testing` directory) also needs to be updated. Nice to add few vertical scale testing steps in `scale` task.

ios simulator screen shot apr 20 2015 12 41 58 pm

rsanchezsaez:
Added two vertical scale tests to the `scale` task in `ORKTest`.

Also added `vertical` property to `NSCoding`, and added it to the serialization tests. Finally, renamed some remaining `verticalFlag` arguments to `vertical`.

jwe-apple:
Thanks, will review shortly.

ResearchKit/Common/ORKQuestionStepViewController.m
```diff
@@ -784,8 +785,11 @@ - (CGFloat)tableView:(UITableView *)tableView heightForRowAtIndexPath:(NSIndexPa
             height = [ORKSurveyAnswerCellForScale suggestedCellHeightForView:tableView];
         }
             break;
+        case ORKQuestionTypeVerticalScale:{
+            height = [ORKSurveyAnswerCellForVerticalScale suggestedCellHeightForView:tableView];
+        }
+            break;
```
jwe-apple: Rather than adding this question type, I'd prefer to change the logic for ORKQuestionTypeScale to interrogate the impliedAnswerFormat of the question step, cast appropriately, and check the vertical property. It might be best to add a method to questionStep like isFormatVertical, which makes the appropriate check. This would be in line with the above code for ORKQuestionTypeSingle/MultipleChoice.

ResearchKit/Common/ORKAnswerFormat.h
```diff
@@ -45,6 +45,9 @@ typedef NS_ENUM(NSInteger, ORKQuestionType) {
     /// The scale question type asks participants to place a mark at an appropriate position on a continuous or discrete line.
     ORKQuestionTypeScale,
     
+    /// The vertical scale question type asks participants to place a mark at an appropriate position on a continuous or discrete vertical line.
+    ORKQuestionTypeVerticalScale,
```
jwe-apple: Would prefer not to add this as a separate question type. See further comment below.

ResearchKit/Common/ORKQuestionStepViewController.m
```diff
@@ -563,6 +563,7 @@ - (ORKSurveyAnswerCell *)answerCellForTableView:(UITableView *)tableView
     static dispatch_once_t onceToken;
     dispatch_once(&onceToken, ^{
         typeAndCellMapping = @{@(ORKQuestionTypeScale): [ORKSurveyAnswerCellForScale class],
+                               @(ORKQuestionTypeVerticalScale): [ORKSurveyAnswerCellForVerticalScale class],
```
jwe-apple: Instead of doing this, you can do a special case correction in the code below this to handle the vertical case.

ResearchKit/Common/ORKScaleSlider.m
```diff
@@ -96,10 +164,9 @@ - (void)updateValueForTouchAtPoint:(CGPoint)touchPoint {
 }
 
 static CGFloat kLineWidth = 1.0;
-- (void)drawRect:(CGRect)rect
-{
-    CGRect trackRect = [self trackRectForBounds:[self bounds]];
-    CGFloat centerY = [self bounds].size.height / 2.0;
+- (void)drawRect:(CGRect)rect {
+    CGRect trackRect = [self trackRectForBounds:self.bounds];
+    CGFloat centerY = self.bounds.size.height / 2.0;
```
jwe-apple: This is ok, but we do prefer the bounds property calls to be written as method calls to make the fact that they're method calls explicit. If I were going to make a change here, for instance, I'd factor out the call to the bounds method and then re-use the struct for both the following calculations.

ResearchKit/Common/ORKScaleSlider.m
```diff
@@ -60,17 +63,78 @@ - (id)initWithFrame:(CGRect)frame
...
+    }
+}
+
+- (UIView *)thumbImageSubview {
+    UIView *thumbImageSubview = nil;
+    CGRect trackRect = [self trackRectForBounds:self.bounds];
+    CGRect thumbRect = [self thumbRectForBounds:self.bounds trackRect:trackRect value:self.value];
+    for (UIView *subview in self.subviews) {
+        if (CGRectEqualToRect(thumbRect, subview.frame)) {
+            thumbImageSubview = subview;
```
jwe-apple: This is ugly and error-prone. It would be better to have and explicitly set a thumb image asset. I'm willing to accept in the short term, though. Please add a comment to this effect and open an issue for fixing in future.

ResearchKit/Common/ORKScaleSlider.m
rsanchezsaez: Ah, I see. I replaced it to a *property getter call* because of `UIView` public documentation defining it as a *property*. I'll change it back.

ResearchKit/Common/ORKScaleSlider.m
```diff
@@ -60,17 +63,78 @@ - (id)initWithFrame:(CGRect)frame
...
+                                                        multiplier:1.0
+                                                          constant:0.0]];
+    }
+}
+
+- (UIView *)hitTest:(CGPoint)point withEvent:(UIEvent *)event {
+    UIView *view = nil;
+    if (_vertical) {
+        // In vertical mode, we need to ignore the touch area for the needed extra width
+        const CGFloat desiredSliderWidth = 36.0;
```
jwe-apple: Can we widen this a little, perhaps to 44?

ResearchKit/Common/ORKScaleSlider.m
jwe-apple: It is a property, but ObjC property accesses are still method calls, and they're still relatively expensive compared to accessing (for example) a member of a struct.

ResearchKit/Common/ORKQuestionStepViewController.m
rsanchezsaez: Yeah, I never felt good about adding a new `ORKQuestionTypeVerticalScale` value. I didn't realize `ORKQuestionStep` had this helper methods. I will improve it.

ResearchKit/Common/ORKScaleSlider.m
rsanchezsaez: I agree that it's hackish and not very future proof, but in my opinion the usage of the actual `thumbRect` for identifying the subview makes it quite safe. Doesn't it?

Anyway, yeah, using an explicit thumb would be better, but I didn't wan't to bother you guys for a thumb design. I'll add a comment.

ResearchKit/Common/ORKScaleSlider.m
rsanchezsaez: I see, thanks for the clarification. :)

rsanchezsaez:
All feedback issues taken care of.

jwe-apple:
With that one naming change I'll merge this.

rsanchezsaez:
Done.

jwe-apple:
Thanks, great job on this!

rsanchezsaez:
Thanks for the thoughtful comments all along. :+1:

rsanchezsaez:
Hey, I think I made a mistake in `ORKQuestionStep`. One needs to do:

    return (ORKDynamicCast([self impliedAnswerFormat], ORKScaleAnswerFormat).vertical ||
            ORKDynamicCast([self impliedAnswerFormat], ORKContinuousScaleAnswerFormat).vertical);

(the property name) instead of:

    return (ORKDynamicCast([self impliedAnswerFormat], ORKScaleAnswerFormat).isVertical ||
            ORKDynamicCast([self impliedAnswerFormat], ORKContinuousScaleAnswerFormat).isVertical);

(the getter name).

If you clean the project and recompile, *Xcode* is not happy. I've fixed it in the latest commit in [the corresponding branch](https://github.com/rsanchezsaez/ResearchKit/tree/rsanchezsaez-verticalscale). I think you can just do a fast-forward merge to keep *GitHub* clean, but tell me if you want me to submit a separate pull request.

I spotted it when trying to compile the *merged master* branch. I'm not sure why *Xcode* was happily compiling on my separate branch.

rsanchezsaez:
Oh, I figured why it doesn't compile now. [Pull request #48](https://github.com/ResearchKit/ResearchKit/pull/48) (Fix double evaluation in ORKDynamicCast macro) makes the compiler to see the general `id` type, rather than the casted type, coming from the  `ORKDynamicCast()` macro. Not sure if [PR#48](https://github.com/ResearchKit/ResearchKit/pull/48) needs to be re-evaluated or not.

Still, the fix to my code still applies, I think.

rsanchezsaez:
Actually, my previous fix does not compiler either. Explicitly calling the getter works:

    return ([ORKDynamicCast([self impliedAnswerFormat], ORKScaleAnswerFormat) isVertical] ||
            [ORKDynamicCast([self impliedAnswerFormat], ORKContinuousScaleAnswerFormat) isVertical]);

I've overwritten my fixing commit to reflect this. Although I think we may want to go back to an `ORKDynamicCast()` implementation that returns a properly casted object (and then I'd prefer to just use the `.vertical` property syntax).

Sorry about the confusion.

jwe-apple:
Don't panic. I think the ORKDynamicCast() just needs extra parens. I'll fix it up.

jwe-apple:
Just pushed my fix to master, thanks!

rsanchezsaez:
Perfect.

drekryan:
@jwe-apple :+1: Thanks for that fix!

PR #39, opened 2015-04-16
Format property declarations

sikhapol:
I looked at the property declarations and figure that the 80-85% of them are in this format:
`@property (nonatomic, strong/weak/copy/assign/unsafe-unretained, readonly/readwrite, nullable/nonnull, getter=..., setter=...) Type *object;`

So, I reformatted (manually) the lines that are not in the format above to be consistent with the rest.


YuanZhu-apple:
Consistency+1!

Issue #29, opened 2015-04-15
Convert all non-English Localizable.strings to human-readable format


jwe-apple:
Will make it easier to review changes.

rsanchezsaez:
Done. See [Pull Request #58](https://github.com/ResearchKit/ResearchKit/pull/58) for comments.

jwe-apple:
Yep, thanks. This is done!

PR #40, opened 2015-04-16
[WIP] Convert all non-English Localizable.strings to human-readable format 

3lvis:
Hey guys, I started working on this task #29. It's a quite extensive one so I was wondering about something that I did first. 

**Sorting English localizations**, is that ok with you? I find localizations easy to maintain when they are structured in this way, it's easier to find where to put your localized string if you use a sorting method.

The sorting is done per section, sections that are already sorted haven't been touched.

### DIFF
--- a/ResearchKit/Localized/en.lproj/Localizable.strings
+++ b/ResearchKit/Localized/en.lproj/Localizable.strings
@@ -1,21 +1,21 @@
 /*
  Copyright (c) 2015, Apple Inc. All rights reserved.
-
+ 
  Redistribution and use in source and binary forms, with or without modification,
  are permitted provided that the following conditions are met:
-
+ 
  1.  Redistributions of source code must retain the above copyright notice, this
  list of conditions and the following disclaimer.
-
+ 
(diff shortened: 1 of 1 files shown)


jwe-apple:
1) Would rather not reorder the English at this point, sorry. It would make it harder to track changes versus the last localization run.

2) The non-English languages should ideally be converted with an automated script rather than by hand. We have an internal process for updating these at release time, and we will need to apply the script each time the localizations go through that process. If the script has a stable sort as output, that would be preferable, since it will make for cleaner diffs.

3lvis:
Understood! Will see what I can do :smile:

rsanchezsaez:
Hey @3lvis, I went ahead and implemented this ([PR#58](https://github.com/ResearchKit/ResearchKit/pull/58)). I hope you don't mind!

Best,
Ricardo

3lvis:
@rsanchezsaez Haha, don't worry. I see you're kicking ass!

rsanchezsaez:
:)

Issue #4, opened 2015-04-10
Scale answer format support in form items


YuanZhu-apple:
So you can have an array of scale questions on a form.

* Form items can have section headers, each scale should occupy one section.
* Question text will be displayed as section header

jwe-apple:
Completed!

PR #41, opened 2015-04-16
Scale answer support in form items (issue #4)

brucehappy:
This is a very basic implementation of issue #4.  There is obviously some refactoring that should be considered for Survey and Form table cells as there is much they have in common.  I didn't tackle that, but instead added just the bare necessities for a scale answer as a forum item.  When PR #33 is merged, the height function will need to change in my new scale form cell class, as well as in the survey version to take into account vertical orientations.

I tested this by modifying the ORKCatalog app to include two new form items in the Form example: an integer and a decimal slider.  I can include the change to the ORKCatalog app in this PR if desired.


rsanchezsaez:
Once [pull request #33](https://github.com/ResearchKit/ResearchKit/pull/33) is merged (I need to do some changes based on feedback), I suggest to add a check on your implementation to throw an exception in case you try to use a vertical scale within a form (which in my opinion doesn't make much sense).

I suggest you proactively include your addition to the `ORKCatalog` app. It makes it easier for everybody to see how it looks. ;-)

Also, note that this must support several scales on the same form. I suggest that your changes to ORKCatalog reflect this. Rather than two new *form steps*, I'd just add one additional *form step* with several discrete and continuous rulers in it.

Best!

brucehappy:
If preventing people from using vertical scale slider form items in a form step is required, I can certainly add that restriction, though, in general, I'd rather not impose restrictions based on stylistic or aesthetic considerations.

As for the ORKCatalog app, my test added two slider form items to the existing form step.  I was hesitant to include this in the PR simply because it was clear that the ORKCatalog app was not an attempt to demonstrate all possible configurations.  For example, the existing form step only contains one numeric form item and one time based form item, though multiple other form item types are supported (image picker, grouped choices, etc).  I have now added the change to the ORKCatalog app to the PR.

jwe-apple:
Actually we'd like to keep ORKCatalog clean. More exhaustive development test cases, especially for forms, should go in the ORKTest objective-C app.

brucehappy:
I'll revert the last commit that modified the ORKCatalog app, and instead add the sliders to the kitchen sink section form step in the ORKTest app.  Thanks.

brucehappy:
Ok, ORKCatalog has been reverted, and ORKTest has two sliders added to its kitchen sink form step.  I've also merged master.  Let me know if there's anything else needed.

rsanchezsaez:
Sorry about suggesting that adding more cases to the catalog was a good idea. :)

ResearchKit/Common/ORKFormItemCell.h
```diff
@@ -83,4 +83,8 @@ NS_ASSUME_NONNULL_BEGIN
 @interface ORKFormItemPickerCell : ORKFormItemTextFieldBasedCell
 @end
 
+@interface ORKFormItemScaleCell : ORKFormItemCell
++ (CGFloat)suggestedCellHeightInTableView:(nullable UITableView *)tableView;
```
YuanZhu-apple: Should use auto layout like other `ORKFormItemCell`, Instead of a hard code height.

YuanZhu-apple:
Also , can you make sure the scale cell has proper vertical margins.

image

brucehappy:
I've removed the hardcoded cell height for scales.  I had mimicked the behavior I saw in Survey scale cells when I used the hardcoded height, so I removed the hardcoded height there as well.  The vertical margins should also be correct as a result of this change.
screen shot 2015-04-18 at 4 03 34 am

YuanZhu-apple:
Please wait #33 to be merged first. 
And can you send an email to researchkit@apple.com, so we can send you `ResearchKit Contribution Terms and License Policy`.

brucehappy:
Sure, I'll wait for #33 to be merged, and then I'll go back and add whatever changes necessary to support the vertical scale.  I've responded to the policy email.  And I correct in my understanding that I need to modify every file in my commits that have more than 10 lines added/changed/removed to include a copyright notice?

jwe-apple:
Yes. i.e., ORKFormItemCell.m will need a copyright line for you.

jwe-apple:
#33 has been merged.

brucehappy:
Ok, great.  I'll work on getting this PR back into shape.

rsanchezsaez:
I spotted a small problem with my code, see my latest comment at [pull request #33](https://github.com/ResearchKit/ResearchKit/pull/33). It's a very quick fix.

jwe-apple:
Build's fixed, go ahead!

brucehappy:
Ended up spending my time on issue #1 today.  I'll jump back on cleaning up this PR tomorrow.

brucehappy:
This PR should be ready for review.  Master merged (after it merged #33) and copyrights added.

Auto layout and I went 20 rounds, and, in the end, I sorta won.  @YuanZhu-apple had asked for hardcoded height value to be removed for the slider cell in the form step, and instead use auto layout.  I did that for the horizontal cells without issue last week.  The vertical cells included with PR #33 were another story due to the circular nature of making the height the same as the width, and having the width fill via auto layout.  In the end I made the width available to the slider via a delegate for use as an intrinsic height when it is in vertical mode.  If there is another preferred method that also works through orientation changes, table cell reuse, both form and question step use, etc, I would really like to hear about it.

ResearchKit/Common/ORKQuestionStepViewController.m
```diff
@@ -553,8 +553,6 @@ - (ORKSurveyAnswerCell *)answerCellForTableView:(UITableView *)tableView {
         class = [ORKSurveyAnswerCellForTextField class];
     } else if ([[self.questionStep impliedAnswerFormat] isKindOfClass:[ORKValuePickerAnswerFormat class]]) {
         class = [ORKSurveyAnswerCellForPicker class];
-    } else if ([self.questionStep isFormatVerticalScale]) {
-        class = [ORKSurveyAnswerCellForVerticalScale class];
```
jwe-apple: This looks like a bad merge?

ResearchKit/Common/ORKQuestionStepViewController.m
```diff
@@ -741,15 +739,6 @@ - (CGFloat)tableView:(UITableView *)tableView heightForRowAtIndexPath:(NSIndexPa
...
             break;
-        case ORKQuestionTypeScale:{
-            if ([self.questionStep isFormatVerticalScale]) {
-                height = [ORKSurveyAnswerCellForVerticalScale suggestedCellHeightForView:tableView];
-            }
-            else {
-                height = [ORKSurveyAnswerCellForScale suggestedCellHeightForView:tableView];
-            }
-        }
-            break;
```
jwe-apple: This looks like a bad merge?

ResearchKit/Common/ORKQuestionStep_Internal.h
```diff
@@ -39,7 +39,6 @@ NS_ASSUME_NONNULL_BEGIN
 - (BOOL)isFormatChoiceWithImageOptions;
 - (BOOL)isFormatFitsChoiceCells;
 - (BOOL)isFormatTextfield;
-- (BOOL)isFormatVerticalScale;
```
jwe-apple: ditto?

ResearchKit/Common/ORKScaleSlider.m
```diff
@@ -99,26 +100,22 @@ - (UIView *)thumbImageSubview {
...
-                                                          constant:0.0]];
-    }
+- (CGSize)intrinsicContentSize {
+    CGSize s = [super intrinsicContentSize];
+    // If we have a layout width provided by our delegate and we are vertical, use the provided
+    // width for the instrinsic content height, and leave the intrinsic content width alone.
+    // The intrinsic content width is typically -1, which will allow the slider to fill the
+    // available width in the superview.
+    CGFloat sliderLayoutWidth = self.delegate ? self.delegate.sliderLayoutWidth : 0;
+    if(_vertical && sliderLayoutWidth>0)
```
jwe-apple: Spacing: we want a space before and after the '>' operator

ResearchKit/Common/ORKQuestionStepViewController.m
brucehappy: No, these lines were removed on purpose as they are only there in order to eventually provide the hardcoded cell height, which is no longer needed.

ResearchKit/Common/ORKScaleSlider.m
brucehappy: Fixed

ResearchKit/Common/ORKQuestionStepViewController.m
brucehappy: No, see comment above.  This is the removal of the hardcoded height provided by the suggestedCellHeightForView calls, which was replaced by auto layout on the cell.

ResearchKit/Common/ORKQuestionStep_Internal.h
brucehappy: No, see comment above.

ResearchKit/Common/ORKScaleSliderView.m
```diff
@@ -128,79 +159,68 @@ - (void)updateConstraints {
     
     NSDictionary *views = NSDictionaryOfVariableBindings(_slider,_leftRangeLabel,_rightRangeLabel, _valueLabel);
     
-    if ([_formatProvider isVertical])
-    {
+    if ([_formatProvider isVertical]) {
```
jwe-apple: The existing code was not great in this respect, and it's quite likely I missed it in a previous review, but it seems more likely that we should be setting up the constraints *once* at setup time rather than in updateConstraints. If you do it in updateConstraints you have to be quite careful to deal with old constraints, etc.

We know in advance whether we're vertical or not, and these constraints are totally fixed -- they're not dynamic.

ResearchKit/Common/ORKQuestionStepViewController.m
jwe-apple: Ah, sorry, spotted that later!

ResearchKit/Common/ORKScaleSliderView.m
```diff
@@ -117,6 +128,26 @@ - (IBAction)sliderValueChanged:(id)sender {
...
+    // Use the delegate expected width, if available, or if we have no delegate, then use our own width.
+    // Subtract the left and right margins from the width, to give the appropriate slider width.
+    return (self.delegate ? self.delegate.sliderLayoutWidth : self.previousLayoutWidth)-(kVerticalSliderHorizontalMargin*2);
+}
+
+-(void)layoutSubviews {
+    [super layoutSubviews];
+    // This is currently only used by vertical sliders, but keep track of the width to implement the Protocol anyway
+    CGFloat currentLayoutWidth = self.delegate ? self.delegate.sliderLayoutWidth : self.bounds.size.width;
+    if(self.previousLayoutWidth!=currentLayoutWidth) {
```
jwe-apple: Spacing around the operator.

ResearchKit/Common/ORKScaleSliderView.m
```diff
@@ -117,6 +128,26 @@ - (IBAction)sliderValueChanged:(id)sender {
...
+-(CGFloat)sliderLayoutWidth {
+    // Use the delegate expected width, if available, or if we have no delegate, then use our own width.
+    // Subtract the left and right margins from the width, to give the appropriate slider width.
+    return (self.delegate ? self.delegate.sliderLayoutWidth : self.previousLayoutWidth)-(kVerticalSliderHorizontalMargin*2);
+}
+
+-(void)layoutSubviews {
+    [super layoutSubviews];
+    // This is currently only used by vertical sliders, but keep track of the width to implement the Protocol anyway
+    CGFloat currentLayoutWidth = self.delegate ? self.delegate.sliderLayoutWidth : self.bounds.size.width;
```
jwe-apple: Why is this code in layoutSubviews? Perhaps the delegate relationship ought to be the other way round (so we get told when the required width changes)?

ResearchKit/Common/ORKScaleSliderView.m
```diff
@@ -128,79 +159,68 @@ - (void)updateConstraints {
...
+                                                         attribute:NSLayoutAttributeCenterX
+                                                        multiplier:1.0
+                                                          constant:-kSideLabelMargin]];
         
-        [self addConstraints:[NSLayoutConstraint constraintsWithVisualFormat:@"H:|-kMargin-[_leftRangeLabel]-kMargin-[_slider]-kMargin-[_rightRangeLabel(==_leftRangeLabel)]-kMargin-|"
-                                                                     options:NSLayoutFormatAlignAllCenterY|NSLayoutFormatDirectionLeadingToTrailing
-                                                                     metrics:@{@"kMargin": @(kMargin)}
-                                                                       views:views]];
+        [self addConstraint:[NSLayoutConstraint constraintWithItem:self.rightRangeLabel
+                                                        attribute:NSLayoutAttributeCenterY
```
jwe-apple: Why is this in terms of centerY rather than top? I feel like I'm missing something here!

ResearchKit/Common/ORKScaleSliderView.m
brucehappy: I kept the original updateConstraints there when I changed the constraints. You are correct that they are fixed constraints and updateConstraints is not needed.  I've eliminated it, and moved the code up into the initializer.

ResearchKit/Common/ORKScaleSliderView.m
brucehappy: Fixed

ResearchKit/Common/ORKScaleSliderView.m
brucehappy: This is a slightly modified version of what was in #33.  The labels for the min and max in vertical mode get centered vertically relative to the top and bottom points of the slider line, thus the use of CenterY for the labels and Top and Bottom for the slider.

jwe-apple: Got it. And this is ok because the margin is 20 pt and we don't think the font will ever be bigger than 40pt line height, so won't extend beyond the bounds of the view.

jwe-apple:
ok to test

rsanchezsaez:
I suggest that next time you keep *merge commits* and *implementation commits* separate.

[This commit](https://github.com/brucehappy/ResearchKit/commit/9eb1d445bb19b86997e2b79fb633b07ff24d9b81) has both a merge and some of your changes, so it's quite difficult to review within *GitHub*. ;)

jwe-apple:
@rsanchezsaez agree, but I don't mind if it happens in this PR or another follow-on one.

I'll test this shortly, but code looks ready.

jwe-apple:
Looks good to me, sending for accessibility review.

brucehappy:
@rsanchezsaez Yeah, you are right, the commit had a merge and also changes to #33 in it.  I suppose I could have done the merge and committed, even if it had half auto layout support and half not, and then another commit for fixing the auto layout of vertical sliders.  Will keep in mind next time, thanks.

jwe-apple:
@brucehappy Thanks, still waiting on review.

jwe-apple:
Ok, this one's good!

rsanchezsaez:
@brucehappy I just tested this for a bit and it looks great, congrats! Good call on not forbidding vertical scales on the form, they look good. It was silly to suggest that.

brucehappy:
Thanks @rsanchezsaez !  Glad to work with you on this.

jwe-apple:
Thanks guys. Amazing progress!

PR #48, opened 2015-04-17
Fixed possible double evaluation in the ORKDynamicCast macro

wczekalski:
#45 Can be closed when this is merged. Kudos to @steipete for finding this issue.

### DIFF
--- a/ResearchKit/Common/ORKDefines_Private.h
+++ b/ResearchKit/Common/ORKDefines_Private.h
@@ -37,8 +37,9 @@ ORK_EXTERN NSBundle *ORKBundle() ORK_AVAILABLE_DECL;
 #define ORKLocalizedString(key, comment) \
 [ORKBundle() localizedStringForKey:(key) value:@"" table:nil]
 
-#define ORKDynamicCast(x, c) ((c *) ([x isKindOfClass:[c class]] ? x : nil))
+ORK_EXTERN id _ORKDynamicCast(id x, Class objClass) ORK_AVAILABLE_DECL;
 
+#define ORKDynamicCast(x, c) (c *) _ORKDynamicCast(x, [c class])
 
 ORK_EXTERN NSString *ORKTimeOfDayStringFromComponents(NSDateComponents *dateComponents) ORK_AVAILABLE_DECL;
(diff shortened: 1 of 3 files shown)


ResearchKit/Common/ORKDefines_Private.h
```diff
@@ -37,8 +37,9 @@ ORK_EXTERN NSBundle *ORKBundle() ORK_AVAILABLE_DECL;
 #define ORKLocalizedString(key, comment) \
 [ORKBundle() localizedStringForKey:(key) value:@"" table:nil]
 
-#define ORKDynamicCast(x, c) ((c *) ([x isKindOfClass:[c class]] ? x : nil))
+ORK_EXTERN id _ORKDynamicCast(id x, Class objClass) ORK_AVAILABLE_DECL;
 
+#define ORKDynamicCast(x, c) (c *) _ORKDynamicCast(x, [c class])
```
jwe-apple: Let's move this to ORKHelpers.h

We don't want this to live in API (and anything _Private.h is exposed in the private headers and so is developer accessible, even if it is not Swift accessible).

ResearchKit/Common/ORKDefines_Private.h
```diff
@@ -37,8 +37,9 @@ ORK_EXTERN NSBundle *ORKBundle() ORK_AVAILABLE_DECL;
 #define ORKLocalizedString(key, comment) \
 [ORKBundle() localizedStringForKey:(key) value:@"" table:nil]
 
-#define ORKDynamicCast(x, c) ((c *) ([x isKindOfClass:[c class]] ? x : nil))
+ORK_EXTERN id _ORKDynamicCast(id x, Class objClass) ORK_AVAILABLE_DECL;
```
jwe-apple: Let's rename to put the underscore at the end. (i.e. ORKDynamicCast_)

And do not mark ORK_EXTERN because we don't want the symbol exposed when built as a framework. And drop ORK_AVAILABLE_DECL, which is not used in non-public headers.

It’s technically against the rules to use names that start with an underscore followed by a capital letter (C11 7.1.3p1)

jwe-apple:
Looks good.

PR #51, opened 2015-04-17
More tests

garnett:
I have added test for some classes. Please feel free to comment style/naming/etc. because it's hard to get it right without complete guide

### DIFF
--- a/ResearchKit.xcodeproj/project.pbxproj
+++ b/ResearchKit.xcodeproj/project.pbxproj
@@ -23,6 +23,7 @@
 
 /* Begin PBXBuildFile section */
 		2EBFE11D1AE1B32D00CB8254 /* UIViewORKAccessibilityTests.m in Sources */ = {isa = PBXBuildFile; fileRef = 2EBFE11C1AE1B32D00CB8254 /* UIViewORKAccessibilityTests.m */; };
+		2EBFE1201AE1B74100CB8254 /* ORKVoiceEngineTests.m in Sources */ = {isa = PBXBuildFile; fileRef = 2EBFE11F1AE1B74100CB8254 /* ORKVoiceEngineTests.m */; };
 		618DA04E1A93D0D600E63AA8 /* ORKAccessibility.h in Headers */ = {isa = PBXBuildFile; fileRef = 618DA0481A93D0D600E63AA8 /* ORKAccessibility.h */; };
 		618DA0501A93D0D600E63AA8 /* ORKAccessibilityFunctions.h in Headers */ = {isa = PBXBuildFile; fileRef = 618DA0491A93D0D600E63AA8 /* ORKAccessibilityFunctions.h */; };
 		618DA0521A93D0D600E63AA8 /* ORKAccessibilityFunctions.m in Sources */ = {isa = PBXBuildFile; fileRef = 618DA04A1A93D0D600E63AA8 /* ORKAccessibilityFunctions.m */; };
@@ -365,6 +366,8 @@
 
 /* Begin PBXFileReference section */
 		2EBFE11C1AE1B32D00CB8254 /* UIViewORKAccessibilityTests.m */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.objc; path = UIViewORKAccessibilityTests.m; sourceTree = "<group>"; };
+		2EBFE11E1AE1B68800CB8254 /* ORKVoiceEngine_Internal.h */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.h; path = ORKVoiceEngine_Internal.h; sourceTree = "<group>"; };
+		2EBFE11F1AE1B74100CB8254 /* ORKVoiceEngineTests.m */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.objc; path = ORKVoiceEngineTests.m; sourceTree = "<group>"; };
 		618DA0481A93D0D600E63AA8 /* ORKAccessibility.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; path = ORKAccessibility.h; sourceTree = "<group>"; };
 		618DA0491A93D0D600E63AA8 /* ORKAccessibilityFunctions.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; path = ORKAccessibilityFunctions.h; sourceTree = "<group>"; };
 		618DA04A1A93D0D600E63AA8 /* ORKAccessibilityFunctions.m */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.objc; path = ORKAccessibilityFunctions.m; sourceTree = "<group>"; };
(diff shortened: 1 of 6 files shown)


ResearchKit.xcodeproj/project.pbxproj
```diff
@@ -1914,7 +1922,9 @@
 				86CC8EBA1AC09383001CCD89 /* ORKResultTests.m in Sources */,
 				86CC8EB81AC09383001CCD89 /* ORKHKSampleTests.m in Sources */,
 				86CC8EB51AC09383001CCD89 /* ORKConsentTests.m in Sources */,
+				2EBFE11D1AE1B32D00CB8254 /* UIViewORKAccessibilityTests.m in Sources */,
 				86CC8EB41AC09383001CCD89 /* ORKChoiceAnswerFormatHelperTests.m in Sources */,
```
YuanZhu-apple: Please use ORK as the prefix for class and file.

ResearchKit/ActiveTasks/ORKVoiceEngine_Internal.h
```diff
@@ -0,0 +1,43 @@
...
+ OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+
+
+#import "ORKVoiceEngine.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
```
YuanZhu-apple: No need to include nullability in this file,  since this header is marked with `Internal`, Swift project cannot see this API.

ResearchKit/ActiveTasks/ORKVoiceEngine_Internal.h
```diff
@@ -0,0 +1,43 @@
...
+@interface ORKVoiceEngine ()
+
+@property (nonatomic, strong, nullable) AVSpeechSynthesizer *speechSynthesizer;
+
```
YuanZhu-apple: Mark this readonly, so developer will not able to change it by accident. 

To make the unit test work, you can subclass `ORKVoiceEngine` and override  `speechSynthesizer` getter method to provide a mock version.

ResearchKitTests/ORKVoiceEngineTests.m
```diff
@@ -0,0 +1,132 @@
...
+ CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#import <XCTest/XCTest.h>
+#import "ORKVoiceEngine_Internal.h"
+
+@interface ORKMockSpeechSynthesizer : AVSpeechSynthesizer {
+    BOOL _speaking;
```
sstigler: Could you change this to be a property? Once you do that you can get rid of the instance variable declaration here, because Xcode will automatically synthesize it for you.

ResearchKitTests/ORKVoiceEngineTests.m
```diff
@@ -0,0 +1,132 @@
...
+@property (nonatomic, readonly) BOOL didSpeakText;
+@property (nonatomic, readonly) NSString *speech;
+@property (nonatomic, readonly) AVSpeechBoundary stopBoundary;
+
+@end
+
+@implementation ORKMockSpeechSynthesizer
+
+- (void)injectSpeaking:(BOOL)speaking {
+    _speaking = speaking;
```
sstigler: In general could you please try to use the property setters and getters whenever possible? This will be possible once you make `_speaking` a property.

ResearchKitTests/ORKVoiceEngineTests.m
```diff
@@ -0,0 +1,132 @@
...
+}
+
+- (BOOL)isSpeaking {
+    return _speaking;
+}
+
+- (BOOL)stopSpeakingAtBoundary:(AVSpeechBoundary)boundary {
+    _didStopSpeaking = YES;
+    _stopBoundary = boundary;
```
sstigler: Same thing here about preferring the property getters and setters over modifying the instance variables directly.

ResearchKitTests/UIViewORKAccessibilityTests.m
```diff
@@ -0,0 +1,69 @@
...
+
+@interface UIViewORKAccessibilityTests : XCTestCase
+
+@end
+
+@implementation UIViewORKAccessibilityTests
+
+- (void)testOrkSuperViewOfType {
+    
+    {   // Returns nil when superview class is nil
```
sstigler: Why do you have all the curly brackets around the different chunks of code here?

sstigler: A good alternative would probably be to split this method up into one method for each assertion.

ResearchKit/ActiveTasks/ORKVoiceEngine_Internal.h
garnett: That's interesting, never heard of that. Could you please provide any reference?
By the way, all other ```internal``` headers of this project contain nullability annotations

ResearchKit.xcodeproj/project.pbxproj
garnett: Done

ResearchKit/ActiveTasks/ORKVoiceEngine_Internal.h
garnett: Done

ResearchKitTests/ORKVoiceEngineTests.m
garnett: Done

ResearchKitTests/ORKVoiceEngineTests.m
garnett: Done

ResearchKitTests/ORKVoiceEngineTests.m
garnett: This is done because public interface of ```ORKMockSpeechSynthesizer``` has all it's properties readonly, I am not sure if doing them readwrite is the best options, since ```readonly``` clarifies the intent of these variables - to be read by assertions, not set manually.
@YuanZhu-apple what do you think?

ResearchKitTests/UIViewORKAccessibilityTests.m
garnett: @sstigler it's done like in other test of ResearchKit - to split almost identical testing scenarios inside one ```test...``` method. But I agree that here 3 separate methods will be more suitable

ResearchKit/ActiveTasks/ORKVoiceEngine_Internal.h
YuanZhu-apple: Oops, I was wrong, having nullability is fine in this file. 
But I don't think `speechSynthesizer` is `nullable`.

ResearchKitTests/ORKVoiceEngineTests.m
YuanZhu-apple: I think `readonly` is fine, since these values are not intended to be modified from outside.

jwe-apple:
I'm ready to accept this, except the copyright message needs updating to include your copyright on the files you have added, per the terms of the contributor agreement I've emailed you separately. Please also reply by email to agree to the contributor agreement.

jwe-apple:
Thanks, merged!

PR #54, opened 2015-04-17
[Podspec] Fixed tag naming

jhersh:
Fixed an inconsistency between the podspec, which was looking for a tag named `v1.0.0`, and the tag's actual name `1.0.0`.


jwe-apple:
Well spotted.

PR #57, opened 2015-04-18
Adds Swift code snippets for Getting Started

mharper:


README.md
```diff
@@ -154,10 +173,20 @@ taskViewController.delegate = self;
 [self presentViewController:taskViewController animated:YES completion:nil];
 ```
 
+Swift
+
+```swift
+let taskViewController = ORKTaskViewController(task: task, taskRunUUID: nil)
+taskViewController.delegate = self
+self.presentViewController(taskViewController, animated: true, completion: nil)
```
YuanZhu-apple: To be consistent with our style, these should just get the “self” stripped off, so it’s
```
presentViewController(taskViewController, animated: true, completion: nil)

File: README.md [hunk 2]
```diff
@@ -171,6 +200,20 @@ which you must implement in order to handle the completion of the task:
...
+```swift
+func taskViewController(taskViewController: ORKTaskViewController, 
+                didFinishWithReason reason: ORKTaskViewControllerFinishReason, 
+                                     error: NSError?) {
+  let taskResult = taskViewController.result
+  // You could do something with the result here.
+
+  // Then, dismiss the task view controller.
+  self.dismissViewControllerAnimated(true, completion: nil)
+}
```
YuanZhu-apple: Same here:
```
dismissViewControllerAnimated(true, completion: nil)

PR #58, opened 2015-04-19
Convert all non-English Localizable.strings to human-readable format

rsanchezsaez:
Deals with [Issue #29](https://github.com/ResearchKit/ResearchKit/issues/29).

Adds `plist-to-strings.py` script.

It can be used to convert binary `.strings` files (binary `.plist`) to plaintext `.strings` files. It takes a master plaintext `.strings` file for the key sort order in the converted file (as requested in the comments of [Pull Request #40](https://github.com/ResearchKit/ResearchKit/pull/40)). See comments on top of the script source for details about its usage.

All non-English `.strings` files have been converted using the script.

Careful reviewing needed to make sure all the converted `.strings` files remain as they should, as encoding and escaped characters handling is tricky. Please also review correct conversion of left-to-right language files (on first glance they seem to be fine).


3lvis:
The files still show up as binary in GitHub.

https://github.com/ResearchKit/ResearchKit/pull/58/files

rsanchezsaez:
@3lvis I think this is due to GitHub's *diff* not showing differences between a binary file and a text file (it wouldn't make much sense). If you click the `View` button next to each file, you'll be able to see the text version.

I think that subsequent modifications to the `.strings` files should show the plaintext diffs correctly.

jwe-apple:
I've not verified these individually, but I have spot checked Chinese and German. The script also looks good, and copyrights are correctly applied. Merging.

PR #59, opened 2015-04-19
add german localization

idev:
translated en Localizable.Strings into german


YuanZhu-apple:
We will convert all files in one batch.

PR #64, opened 2015-04-20
replacing (id) with (instancetype) in initWithCoder: methods

pbodsk:


YuanZhu-apple:
We are following the lead of [NSCoding Protocol](https://developer.apple.com/library/mac/documentation/Cocoa/Reference/Foundation/Protocols/NSCoding_Protocol/index.html#//apple_ref/occ/intfm/NSCoding/initWithCoder) to have return type id.

pbodsk:
OK. I just noticed that some of the `initWithCoder:` methods were returning `instancetype` and others were returning `id`. For instance all the `initWithCoder:` methods in `ORKAnswerFormat` returns `instancetype`, so I thought I'd align the various `initWithCoder:` methods.

Should they all be returning `id` instead then?

jwe-apple:
I don't think it matters - use of instancetype only really makes a difference for factory (class) methods; the type is implicit for init methods. And of course NSCoder's initializer never gets called directly anyway.

But I appreciate the push for consistency so I'm going to accept this regardless.

pbodsk:
Great. Thanks :+1:

PR #65, opened 2015-04-20
Use ORKTapCountLabel and dynamic text in ORKCountdownView

ilyannn:
There are two similar label classes: `ORKTapCountLabel` and (private) `ORKCountDownViewLabel`.

The latter seems to be redundant: both create a thin label of size 56 for normal user-selected dynamic text size. This commit gets rid of `ORKCountDownViewLabel` and updates the calculation of progress indicator metrics. 

Although `ORKTapCountLabel` can change size dynamically, this would make no sense for `ORKCountdownView`, since the latter is only shown on the screens where user is actively engaged.

I tested the view via the *Fitness Check Active Task* screen; dynamic text sizes now seem to work.
 
Line-by-line version of the comments in the [source commit](https://github.com/ilyannn/ResearchKit/commit/b5d169fc41efef5373c6466d7faaa4acc060d8d7).

### DIFF
--- a/ResearchKit/ActiveTasks/ORKCountdownStepViewController.m
+++ b/ResearchKit/ActiveTasks/ORKCountdownStepViewController.m
@@ -37,24 +37,15 @@
 #import "ORKResult.h"
 #import "ORKLabel.h"
 #import "ORKSubheadlineLabel.h"
+#import "ORKTapCountLabel.h"
 #import "ORKHelpers.h"
 #import "ORKAccessibility.h"
 #import "ORKActiveStepView.h"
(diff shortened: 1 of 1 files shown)


jwe-apple:
Fair point, but the magic numbers etc come from visual specs. Although they seem the same, the specs these two came from were different, and could potentially change independently in future (as they have in the past).

ilyannn:
Hmm, where can we talk about changing the specs or take a look at them? It doesn't seem to make sense that there are two labels with different behaviour under user dynamic size selection.

Issue #68, opened 2015-04-21
Visual consent screens should be scrollable


jwe-apple:
Currently, scrolling is disabled on visual consent screens.

When the consent content is taller than the screen, we should allow vertical scroll (a scroll view is already present, so enabling scrolling itself will be easy).

The difficulty is that the animation between consecutive screens needs to work, and the animation itself is played in a view that is overlaid on top of the view of the VC managing the transition. When a transition is required, we should scroll to the top first, then navigate. Scrolling to the top first is only necessary for forward navigation, since animations are only applied on forward navigation and not on backward navigation in visual consent.

If you have an idea for a different work-around, happy to hear it.

rsanchezsaez:
Initial stab at [pull request #79](https://github.com/ResearchKit/ResearchKit/pull/79).

jwe-apple:
Seems I was a bit trigger happy on this one. A couple things need fixing:

a) The visual consent step's scroll view needs to be registered with the task view controller, so we get the hairline effect when scrolling. See the `registeredScrollView` property.

b) The video animation often goes on longer than the view controller transition. If you attempt to scroll during this period, the video animation view probably needs to scroll too, tracking the position of the scroll view.

jwe-apple:
I would revert the change, but since you deleted your branch already let's just carry on from here.

rsanchezsaez:
Regarding *b)*, scrolling is currently disabled until the video animation finishes. Scrolling is enabled at the same time as the back button is updated (notice how the back button on the second consent screen doesn't appear until the video animation finishes).

Do you prefer scrolling to be enabled while the video animation has not finished yet (with video view tracking, etc.)? If yes, should we advance updating of the back button to the same place?

jwe-apple:
Regarding *b)*, in my testing I observed scrolling being enabled when the animation view was still visible, so there must be a bug here.

I also (once) got stuck with scrolling disabled, but haven't been able to reproduce again to isolate the issue.

Let's try just making the animation view track the content offset, and see how we go.

I'd prefer not to change the back button behaviour, since that has been through QA.

rsanchezsaez:
Regarding *b)*, you were right, there was a bug. If you went back to a previous scene, when you forward to the next scene again the scroll remains unlocked even when the animation video is still playing.

[...]

PR #79, opened 2015-04-21
Add scroll to visual content screens

rsanchezsaez:
Stab at doing [Issue #68](https://github.com/ResearchKit/ResearchKit/issues/68).

Adds `ORKConsentSceneViewController_Internal.h` file with convenience methods to enable/disable scrolling, scroll to top, and set the `scrollView` `delegate`.

Modifies `ORKVisualConsentStepViewController` so:
- `ORKConsentSceneViewController` scrolling is enabled only after overlay animation completes (`updatePageIndex` method)
- Current `ORKConsentSceneViewController` scrolls to top in animated fashion before transitioning to next page
- Makes sure next `ORKConsentSceneViewController` is always scrolled to top (guards against going forward, scrolling downward, going back, and going forward again)
- Disregards scrolling when going back

I made the sample *lorem ipsilum* text in `ORKCatalog`'s *Consent task* longer: scrolling can be tested on iPhone 4 and iPhone 5 screens.

### DIFF
--- a/ResearchKit/Consent/ORKConsentSceneViewController.m
+++ b/ResearchKit/Consent/ORKConsentSceneViewController.m
@@ -1,6 +1,7 @@
 /*
  Copyright (c) 2015, Apple Inc. All rights reserved.
- 
+ Copyright (c) 2015, Ricardo Sánchez-Sáez.
+
  Redistribution and use in source and binary forms, with or without modification,
  are permitted provided that the following conditions are met:
  
--- a/ResearchKit/Consent/ORKConsentSceneViewController.m
+++ b/ResearchKit/Consent/ORKConsentSceneViewController.m
@@ -174,6 +174,26 @@ static NSString *localizedLearnMoreForType(ORKConsentSectionType sectionType) {
     }
 }
 
+- (void)setScrollEnabled:(BOOL)enabled {
+    ORKConsentSceneView *consentSceneView = (ORKConsentSceneView *)self.view;
+    consentSceneView.scrollEnabled = enabled;
+}
+
+- (void)scrollToTopAnimated:(BOOL)animated {
+    ORKConsentSceneView *consentSceneView = (ORKConsentSceneView *)self.view;
+    CGPoint targetContentOffset = CGPointMake(consentSceneView.contentOffset.x, 0);
+    [consentSceneView setContentOffset:targetContentOffset animated:animated];
+}
+
+- (BOOL)isScrolledToTop {
+    ORKConsentSceneView *consentSceneView = (ORKConsentSceneView *)self.view;
+    return consentSceneView.contentOffset.y == 0;
+}
+
+- (void)setScrollViewDelegate:(id<UIScrollViewDelegate>)delegate {
+    ORKConsentSceneView *consentSceneView = (ORKConsentSceneView *)self.view;
+    consentSceneView.delegate = delegate;
+}
 
 #pragma mark - Action
 
--- a/ResearchKit/Consent/ORKConsentSceneViewController.m
+++ b/ResearchKit/Consent/ORKConsentSceneViewController.m
@@ -174,25 +174,31 @@ static NSString *localizedLearnMoreForType(ORKConsentSectionType sectionType) {
     }
 }
 
+- (BOOL)scrollEnabled {
+    ORKConsentSceneView *consentSceneView = (ORKConsentSceneView *)self.view;
+    return consentSceneView.scrollEnabled;
+}
+
 - (void)setScrollEnabled:(BOOL)enabled {
     ORKConsentSceneView *consentSceneView = (ORKConsentSceneView *)self.view;
     consentSceneView.scrollEnabled = enabled;
 }
 
-- (void)scrollToTopAnimated:(BOOL)animated {
+- (void)scrollToTopAnimated:(BOOL)animated completion:(void (^)(BOOL finished))completion {
     ORKConsentSceneView *consentSceneView = (ORKConsentSceneView *)self.view;
-    CGPoint targetContentOffset = CGPointMake(consentSceneView.contentOffset.x, 0);
-    [consentSceneView setContentOffset:targetContentOffset animated:animated];
-}
-
-- (BOOL)isScrolledToTop {
-    ORKConsentSceneView *consentSceneView = (ORKConsentSceneView *)self.view;
-    return consentSceneView.contentOffset.y == 0;
-}
-
-- (void)setScrollViewDelegate:(id<UIScrollViewDelegate>)delegate {
-    ORKConsentSceneView *consentSceneView = (ORKConsentSceneView *)self.view;
-    consentSceneView.delegate = delegate;
+    CGRect targetBounds = consentSceneView.bounds;
+    targetBounds.origin.y = 0;
+    if (animated) {
+        [UIView animateWithDuration:0.2 animations:^{
+            consentSceneView.bounds = targetBounds;
+        } completion:completion];
+    } else {
+        consentSceneView.bounds = targetBounds;
+        if (completion) {
+            completion(YES);
+        }
+            
+    }
 }
 
 #pragma mark - Action
--- a/ResearchKit/Consent/ORKConsentSceneViewController_Internal.h
+++ b/ResearchKit/Consent/ORKConsentSceneViewController_Internal.h
@@ -32,15 +32,11 @@
 
 #import "ORKStepViewController.h"
 
-NS_ASSUME_NONNULL_BEGIN
-
 @interface ORKConsentSceneViewController ()
 
-- (void)setScrollEnabled:(BOOL)enabled;
-- (void)scrollToTopAnimated:(BOOL)animated;
-- (BOOL)isScrolledToTop;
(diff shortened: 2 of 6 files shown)


ResearchKit/Consent/ORKVisualConsentStepViewController.m
```diff
@@ -251,10 +252,30 @@ - (IBAction)goToPreviousPage {
...
+    }
+    else
+    {
+        [currentConsentSceneViewController setScrollViewDelegate:self];
+        [currentConsentSceneViewController scrollToTopAnimated:YES];
+    }
 }
 
+- (void)scrollViewDidEndScrollingAnimation:(UIScrollView *)scrollView {
+    self.delegate = nil;
```
jwe-apple: Why do we clear out delegate?

ResearchKit/Consent/ORKVisualConsentStepViewController.m
```diff
@@ -251,10 +252,30 @@ - (IBAction)goToPreviousPage {
...
+    ORKConsentSceneViewController *currentConsentSceneViewController = [self viewControllerForIndex:[self currentIndex]];
+    [currentConsentSceneViewController setScrollEnabled:NO];
+    if ([currentConsentSceneViewController isScrolledToTop])
+    {
+        [self showNextViewController];
+    }
+    else
+    {
+        [currentConsentSceneViewController setScrollViewDelegate:self];
+        [currentConsentSceneViewController scrollToTopAnimated:YES];
```
jwe-apple: This is hacky. Might be better to change the API to the consent scene to animate with a completion, and do a direct animation on the bounds rather than setContentOffset:animated:. Then you'll get a block callback on the animation completion, whether the animation actually finishes or not, which is also less fragile.

samples/ORKCatalog/ORKCatalog/TaskListRow.swift
```diff
@@ -764,13 +764,13 @@ enum TaskListRow: Int, Printable {
...
             
-            consentSection.summary = self.loremIpsumShortText
+            consentSection.summary = self.loremIpsumLongText
             
             if contentSectionType == .Overview {
                 consentSection.htmlContent = htmlContentString
             }
             else {
-                consentSection.content = self.loremIpsumLongText
+                consentSection.content = self.loremIpsumVeryLongText
```
jwe-apple: I'd prefer you did this testing in ORKTest rather than ORKCatalog. ORKCatalog is meant to be representative of how it should actually appear, rather than show worst case.

ResearchKit/Consent/ORKVisualConsentStepViewController.m
```diff
@@ -285,6 +306,8 @@ - (void)updatePageIndex {
     [self updateBackButton];
     [self setScrollEnabled:NO];
```
jwe-apple: Do we still need this call?

ResearchKit/Consent/ORKConsentSceneViewController_Internal.h
```diff
@@ -0,0 +1,46 @@
...
+#import "ORKStepViewController.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface ORKConsentSceneViewController ()
+
+- (void)setScrollEnabled:(BOOL)enabled;
+- (void)scrollToTopAnimated:(BOOL)animated;
+- (BOOL)isScrolledToTop;
+- (void)setScrollViewDelegate:(id<UIScrollViewDelegate>)delegate;
```
jwe-apple: Assuming you still need this, which I doubt, I'd prefer it were a property.

ResearchKit/Consent/ORKConsentSceneViewController_Internal.h
```diff
@@ -0,0 +1,46 @@
...
+
+
+
+#import "ORKStepViewController.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface ORKConsentSceneViewController ()
+
+- (void)setScrollEnabled:(BOOL)enabled;
```
jwe-apple: Also prefer this as a property, if we still need it.

jwe-apple:
Speedy as always!

ResearchKit/Consent/ORKVisualConsentStepViewController.m
rsanchezsaez: It's just defensive coding in case the user goes backward and forward again. Not clearing it should not be a problem, but the `delegate` has no use after the scrolling animation ends, and it's re-set before the `scrollToTopAnimated:` method is called anyway.

ResearchKit/Consent/ORKVisualConsentStepViewController.m
rsanchezsaez: I will have a look at this tomorrow.

samples/ORKCatalog/ORKCatalog/TaskListRow.swift
rsanchezsaez: Will do. :)

ResearchKit/Consent/ORKVisualConsentStepViewController.m
jwe-apple: I *think* this is the wrong delegate to clear.

rsanchezsaez: Oops, you are right. ;)

It won't be needed when we switch to the block callback approach.

rsanchezsaez:
Feedback taken care of. The animate bounds block callback is much nicer, we avoid all the delegate shenanigans. ;)

ResearchKit/Consent/ORKVisualConsentStepViewController.m
```diff
@@ -47,7 +48,7 @@ OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 #import "ORKContinueButton.h"
 #import "ORKAccessibility.h"
 
-@interface ORKVisualConsentStepViewController () <UIPageViewControllerDelegate>
+@interface ORKVisualConsentStepViewController () <UIPageViewControllerDelegate, UIScrollViewDelegate>
```
jwe-apple: UIScrollViewDelegate not needed.

ResearchKit/Consent/ORKVisualConsentStepViewController.m
```diff
@@ -251,10 +252,20 @@ - (IBAction)goToPreviousPage {
 }
 
 - (IBAction)next {
-    [self showViewController:[self viewControllerForIndex:[self currentIndex]+1] forward:YES animated:YES];
-    ORKAccessibilityPostNotificationAfterDelay(UIAccessibilityScreenChangedNotification, nil, 0.5);
+    ORKConsentSceneViewController *currentConsentSceneViewController = [self viewControllerForIndex:[self currentIndex]];
+    [currentConsentSceneViewController setScrollEnabled:NO];
+    [currentConsentSceneViewController scrollToTopAnimated:YES completion:^(BOOL finished) {
+        [self showNextViewController];
```
jwe-apple: My one hesitation is here - couldn't something interrupt this animation? If the animation is not finished, then it seems like the transition has been interrupted, and we should not transition, and we re-enable animation the current view controller. I'm thinking of someone tapping the Cancel button, which could interrupt the animation, for example.

ResearchKit/Consent/ORKVisualConsentStepViewController.m
```diff
@@ -251,10 +252,20 @@ - (IBAction)goToPreviousPage {
...
+    }];
 }
 
+- (void)showNextViewController {
+    ORKConsentSceneViewController *nextConsentSceneViewController = [self viewControllerForIndex:[self currentIndex]+1];
+    [nextConsentSceneViewController scrollToTopAnimated:NO completion:^(BOOL finished) {
+        [self showViewController:nextConsentSceneViewController forward:YES animated:YES];
```
jwe-apple: A comment is appropriate here to indicate that finished is always YES.

jwe-apple:
Good work, almost there. I'll test in the morning and if all is well, merge.

ResearchKit/Consent/ORKConsentSceneViewController.m
```diff
@@ -174,6 +174,32 @@ - (void)setLearnMoreButtonTitle:(NSString *)learnMoreButtonTitle {
     }
 }
 
+- (BOOL)scrollEnabled {
```
jwe-apple: Add your copyright to this file, please!

ResearchKit/Consent/ORKVisualConsentStepViewController.m
rsanchezsaez: Upon testing, it seems that pressing *Cancel* doesn't interrupt the animation+transition, it just continues under the alert view. Anyway, I implemented what you suggested, just to be safe.

rsanchezsaez:
All done!

jwe-apple:
r=me

rsanchezsaez:
I restored the branch, in case you can still revert the merge. I'll look into the newfound issues (https://github.com/ResearchKit/ResearchKit/issues/68).

jwe-apple:
Not going to revert, but thanks anyway ;-)

Issue #70, opened 2015-04-21
Remove ork_firstResponderView and track current cell directly


jwe-apple:
On review of the recent ork_firstResponderView change, we feel it's not necessary to have this hack to get the first responder. Instead, the table view container should track the currently active cell via a delegate relationship.

rsanchezsaez:
Note: relevant change introduced in [merge #32](https://github.com/ResearchKit/ResearchKit/pull/32).

rsanchezsaez:
Attempt to fix this at [pull request #83](https://github.com/ResearchKit/ResearchKit/pull/83). See the description for details.

jwe-apple:
@rsanchezsaez merge #32 didn't introduce it, it just renamed it.

jwe-apple:
Change verified. Closing.

PR #83, opened 2015-04-22
Track ORKTableContainerView's current first responder cell via delegate

rsanchezsaez:
Stab at fixing [issue #70](https://github.com/ResearchKit/ResearchKit/issues/70).

---

Adds `ORKTableContainerViewDelegate` protocol with a single

    - (UITableViewCell *)currentFirstResponderCellForTableContainerView:(ORKTableContainerView *)tableContainerView

method.

`ORKFormStepViewController` implements it to provide the current first responder `ORKFormItemCell` to its child table container view.

---

On the other hand, `ORKQuestionStepViewController` doesn't implement this method and leaves its child  table container view's `delegate` unset. I think finding the current first responder is not needed for full page table view cells.

---

Also: remove no longer used 'ork_currentFirstResponderView' hack method.


ResearchKit/Common/ORKFormStepViewController.m
```diff
@@ -977,6 +981,7 @@ - (void)tableView:(UITableView *)tableView willDisplayCell:(UITableViewCell *)ce
 
 - (void)formItemCellDidBecomeFirstResponder:(ORKFormItemCell *)cell
 {
+    _currentFirstResponderCell = cell;
```
jwe-apple: I *think* this should also be cleared when the cell resigns first responder.

rsanchezsaez:
I went ahead and implemented proper `_currentFirstResponderCell` clearing on *cell resign first responder*. It feels more future-proof this way.

jwe-apple:
Yep, looks good.

Issue #69, opened 2015-04-21
ORKTest should include a test of repeated identifiers in a task


jwe-apple:
If a task repeats the same question multiple times with the same identifier, there should be just one reported result from the task, that should include the most recent value entered.

The intermediate results reported during the task should also be accurate for each time the same identifier was presented.

We have had reports that this does not work correctly; this task is to investigate, add a test case for it in ORKTest, and fix the issue (if any).

rsanchezsaez:
Currently, if a task repeats the same *question* step multiple times using the same *identifier* the user enters an infinite loop. This is due to `ORKOrderedTask`'s `indexOfStep:` returning the index of the first duplicate step when we reach the second step. This in turn affects both `stepAfterStep:withResult:` and `progressOfCurrentStep:withResult:`.

How do you think `ORKOrderedTask` should handle this? I can think of three options:
- `ORKOrderedTask` keeps the current step index as an ivar.
- `ORKStep` has an auto-generated internal UUID for differentiating repeated steps.
- `indexOfStep:` uses the `ORKStep` object reference instead of `ORKStep`'s `isEqual:`.

Or am I misunderstanding the problem completely?

---

Also, what should happen if two completely unrelated questions on the same task happen to have the same identifier? Currently, it fails the `NSParameterAssert()` (validating the previous step answer) when you try to navigate to the different step with the repeated identifier.

Maybe we can check earlier: `ORKOrderedTask` could check that if two steps have the same *identifier* they should exactly represent the same question.

jwe-apple:
I don't think ORKOrderedTask should support this at all. It would be reasonable for ORKOrderedTask not to pass verification if it has two steps with the same identifier.

This is for a custom ORKTask, which due to how the user has answered decides to re-ask a question.

rsanchezsaez:
Is there such a custom `ORKTask` in any of the open sourced apps? Any additional details about how is this not working correctly?

jwe-apple:
In the open sourced apps, the issue came up in the consent quiz, which is a custom task that replays the quiz from the beginning if you make a mistake.

That said, to reproduce and experiment it would be easier to just create a custom task that replays the same question step indefinitely.

rsanchezsaez:
I've been able to reproduce the issue in *ShareTheJourney*: once you fail the quiz, it cannot be completed successfully upon restart.

I'll look into this.

jwe-apple:
Thanks!

[...]

PR #95, opened 2015-04-22
Fix single text choice cell should not deselect when re-selecting the same answer

rsanchezsaez:
Fixes [issue #69](https://github.com/ResearchKit/ResearchKit/issues/69).

*ResearchKit* tests updated accordingly. Tested on *Share the Journey* by failing the quiz the first time and getting it correct the second time.


jwe-apple:
Yep, this would explain the regression. ok to test

jwe-apple:
Just a tools issue. This is actually ok. Merging.

Issue #75, opened 2015-04-21
[iPad] Pickers are not aligned properly on iPad


jwe-apple:
It happens on all kinds of pickers: `date`, `time`, `timeInterval`, and `choice`.
The pickers should align to the center of the screen.

![screen shot 2015-04-21 at 15 53 49](https://cloud.githubusercontent.com/assets/11575859/7264260/a5597b0a-e83e-11e4-8340-cc3832b7427f.png)

rsanchezsaez:
Pull request submitted.

PR #98, opened 2015-04-23
Center pickers on iPad

rsanchezsaez:
Fixes [issue #75](https://github.com/ResearchKit/ResearchKit/issues/75).

### DIFF
--- a/ResearchKit/Common/ORKSurveyAnswerCellForPicker.m
+++ b/ResearchKit/Common/ORKSurveyAnswerCellForPicker.m
@@ -66,20 +66,36 @@
         
         [_tempPicker removeFromSuperview];
         _tempPicker = nil;
+        
+        [self setNeedsUpdateConstraints];
     }
 }
 
 - (void)layoutSubviews {
     [super layoutSubviews];
     
-    if (self.picker) {
-        CGSize sz = [_picker.pickerView sizeThatFits:(CGSize){self.bounds.size.width,CGFLOAT_MAX}];
-        _picker.pickerView.frame = (CGRect){{0,0},sz};
+    if (_picker) {
+        CGSize pickerSize = [_picker.pickerView sizeThatFits:(CGSize){self.bounds.size.width,CGFLOAT_MAX}];
+        _picker.pickerView.frame = (CGRect){{0,0}, pickerSize};
     }
     
     if (_tempPicker) {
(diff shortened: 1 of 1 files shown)


rsanchezsaez:
I think you are right regarding duplicate constraints. If that's the case, there are other *ResearchKit* classes potentially doing `updateConstraints:` wrong that should be fixed: `ORKActiveStepTimerView`, `ORKScaleSliderView`, `ORKSurveyAnswerCellForNumber`, `ORKSurveyAnswerCellForScale`, `ORKSurveyAnswerCellForTextField`, `ORKTextFieldView`.

Tell me if this is a problem and I'll create and issue.

For the rest of the classes that implement `updateConstraints:`, there are what seem two equivalent methods for removing all current constraints.

Some use `removeConstraints:`:
- `ORKActiveStepQuantityView`, `ORKQuantityPairView`, `ORKFitnessContentView`, `ORKSurveyAnswerCellForText`, `ORKWalkingContentView`.

Other use `deactivateConstraints:` and then *nil* the `_constants` ivar:
- `ORKAudioContentView`, `ORKConsentSignatureWrapperView`, `ORKTableContainerView`, `ORKTappingContentView`, `ORKNavigationContainerView`, `ORKSpatialSpanMemoryContentView`, `ORKVerticalContainerView`.

Which of these two is the preferred method?

Yet, there are some classes that use *ivars* to track and handle the constraints manually added: `ORKFormItemTextFieldBasedCell`, `ORKStepHeaderView`, `ORKContinueButton`. Not sure if these views cannot just remove and re-add all constraints (which I think would be coherent, simpler and more maintainable) due to having some constraints being added elsewhere.

rsanchezsaez:
For what it's worth, the `updateConstraints:` documentation states:

> You must not invalidate any constraints as part of your constraint update phase. You also must not invoke a layout or drawing phase as part of constraint updating.

Does *invalidate* means *deactivate*? That makes me wonder about the `deactivateConstraints:` way.

Also:

> *IMPORTANT*
> Call [super updateConstraints] as the final step in your implementation.

And several `updateConstraints:` implementations do this wrong by calling it before anything else.

rsanchezsaez:
I fixed this in what I felt was the most sensible way. Looking forward to your comments on the issues I mentioned.

jwe-apple:
Your fix works for me.

PR #101, opened 2015-04-24
Tweak code sample

evanmiller:
* firstResult is a property of the ORKStepResult

* Remove extraneous space from the if statement


jwe-apple:
Yep, ok.

PR #103, opened 2015-04-24
Fix small grammatical issue

evanmiller:
Use a semicolon instead of a comma.

Also add a newline to the end of the file.


jwe-apple:
Thanks

PR #104, opened 2015-04-24
Remove unnecessary casts

rsanchezsaez:
I have removed some seemingly unnecessary casts (and added an assert for an indirect assumption-based `NSDictionary` -> `NSMutableDictionary` cast). Removing these produces no warnings (nor static analyzer warnings), and all tests continue to pass.

If these were intentional, feel free to just close this pull request instead.

### DIFF
--- a/ResearchKit/ActiveTasks/HKSample+ORKJSONDictionary.m
+++ b/ResearchKit/ActiveTasks/HKSample+ORKJSONDictionary.m
@@ -44,16 +44,9 @@ static NSString *const kHKCorrelatedObjectsKey = @"objects";
 // static NSString *const kHKSourceIdentifierKey = @"sourceBundleIdentifier";
 
 
-@interface HKSample ()
-
-- (NSMutableDictionary *)ork_JSONMutableDictionaryWithOptions:(ORKSampleJSONOptions)options unit:(HKUnit *)unit;
-
-@end
-
-
 @implementation HKSample (ORKJSONDictionary)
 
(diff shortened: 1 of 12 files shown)


jwe-apple:
Noisy test. This is fine.

ResearchKit/ActiveTasks/HKSample+ORKJSONDictionary.m
```diff
@@ -142,6 +142,7 @@ @implementation HKCorrelation (ORKJSONDictionary)
 
 - (NSDictionary *)ork_JSONDictionaryWithOptions:(ORKSampleJSONOptions)options sampleTypes:(NSArray *)sampleTypes units:(NSArray *)units {
     NSMutableDictionary *mdict = (NSMutableDictionary *)[self ork_JSONDictionaryWithOptions:options unit:nil];
+    NSAssert([mdict isKindOfClass:[NSMutableDictionary class]], @"");
```
jwe-apple: I'd rather this assert need stayed in, *or* if we changed the base class to provide a differently named implementation that explicitly returns a mutable dictionary.

jwe-apple:
Just the one comment - otherwise this looks fine. I think the extra annotations probably were required at an earlier stage when our nullability annotation was incomplete.

rsanchezsaez:
Fixed by adding `ork_JSONMutableDictionaryWithOptions:` method.

jwe-apple:
Otherwise, great.

rsanchezsaez:
Last comment fixed.

Removed anonymous category (the *mutable dictionary* method is only used internally so I didn't add it to the header).

jwe-apple:
Looks good. Test failure is just noise.

PR #109, opened 2015-04-25
Use ORKStrongify() and ORKWeakify() macros to avoid retain cycles in blocks

rsanchezsaez:
I replaced all `__weak typeof(self) weakSelf = self`and `typeof(self) strongSelf = weakSelf` declarations for block use by the new `ORKStrongify()` and `ORKWeakify()` macros.

I feel this is less verbose and less error prone.

### DIFF
--- a/ResearchKit/ActiveTasks/ORKActiveStepTimer.m
+++ b/ResearchKit/ActiveTasks/ORKActiveStepTimer.m
@@ -29,6 +29,7 @@
  */
 
 #import "ORKActiveStepTimer.h"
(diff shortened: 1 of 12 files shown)


jwe-apple:
Actually, in this case I strongly (heh) prefer the verbosity. In ARC memory management it's important to see exactly what's going on and not hide it in a macro.

Also really dislike global macros that internally declare variables in the current scope, since (for instance) you can't use them twice in the same scope.

rsanchezsaez:
Fair enough.

PR #111, opened 2015-04-26
Added Signature Section to htmlForMobile even if htmlReviewContent is used

ekhall:
When creating the PDF in the Consent Section, the signature fields are not appended if htmlReviewContent is used. This appends the signature fields to the htmlReviewContent prior to PDF creation.

### DIFF
--- a/ResearchKit/Consent/ORKConsentDocument.m
+++ b/ResearchKit/Consent/ORKConsentDocument.m
@@ -41,7 +41,8 @@
 #import "ORKDefines_Private.h"
 
 
-@implementation ORKConsentDocument {
+@implementation ORKConsentDocument
+{
     NSMutableArray *_signatures;
 }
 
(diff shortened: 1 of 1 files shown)


jwe-apple:
This was by design, due to the requirements of the initial shipping apps. If adding this feature, it would need to be possible to turn it off, perhaps through configuration on the `ORKConsentReviewStep`. Also prefer not to have this code in two places.

Note also that we're standardizing on having the curly brackets on the line. Some of these were corrected in a recent patch and their reversion in this PR may have been a merge conflict? If you submit a new PR, please make sure your diff is clean!

ekhall:
You're right - shoddy PR. Done in a hurry by a non-programmer trying to get his own app to work :)

jwe-apple:
We still want to hear things like this, even if you don't have time to code it yourself! I'll open an issue to track.

Issue #24, opened 2015-04-15
Uncoordinated curly brackets style


dEn13L:
https://github.com/ResearchKit/ResearchKit/blob/master/ResearchKit/ActiveTasks/ORKRecorder.m

![1](https://cloud.githubusercontent.com/assets/4187743/7151515/05a0bd80-e34d-11e4-8685-cc91914e147c.png)
![2](https://cloud.githubusercontent.com/assets/4187743/7151516/05a24b64-e34d-11e4-9990-13e1d5d57f95.png)

kindraywind:
Also spaces between lines and unimplemented method :\

alistra:
Making a `clang-format` pre-commit hook can fix it for long term

haaakon:
I've mad a clang-format file that tries to adhere to Apples code style. Could use that in this project
https://github.com/haaakon/Apple-clang-format

jwe-apple:
I applied the change in #35.

PR #112, opened 2015-04-27
Homogenize whitespace [wip]

rsanchezsaez:
In the spirit of [issue #24](https://github.com/ResearchKit/ResearchKit/issues/24), I'm homogenizing whitespace.

The general rules I'm following:
- Two spaces between imports; between different interfaces; between different implementations; between interface and implementation
- One space between methods or properties
- No space before first line or after last line of method (except if method signature can be confounded with first line)

I did the *Task*, *Step*, and *Answer Format*. Have a look and let me know if you want me to continue. I think it's worth it because it's early yet.


jwe-apple:
I'm going to accept this; it looks benign. But it does make blame tracing for bugs more difficult. I'd prefer this only happen as part of an existing change, when you're already heavily modifying things in the same area.

rsanchezsaez:
Since you merged this, if you don't mind I'll today finish the whitespace-conform on other half of the project, so you can merge straight away. Having only half it's disquieting! ;)

There haven't been many new features merged in yet, so I figure the disruption to the line blaming will only be minor.

I'll make sure conflicts are resolved for my other pull request as well.

PR #113, opened 2015-04-27
Audio predefined task: audio step should not show too loud if outputdir is nil and developer ignored the error

coxy1989:


ResearchKit/ActiveTasks/ORKAudioStepViewController.m
```diff
@@ -75,10 +72,16 @@ - (void)viewDidLoad {
 
 - (void)viewDidAppear:(BOOL)animated {
     [super viewDidAppear:animated];
-    
     [self start];
 }
 
+- (void)prepareStep {
+    if (_audioRecorder.outputDirectory == nil) {
```
jwe-apple: Errm, but you don't have the audio recorder at this point.

Check self.outputDirectory ?

coxy1989:
facepalm.

hmm outputDirectory is nil in subclasses of ORKActiveStepViewController in prepareStep and therefore recordersDidChange when the subclass is initialised in ORKTaskViewController’s viewControllerForStep: method...

- (ORKStepViewController *)viewControllerForStep:(ORKStep *)step {

— some code — 

stepViewController = [[stepViewControllerClass alloc] initWithStep:step result:result]; (calls prepareStep)

— some more code — 

 stepViewController.outputDirectory = self.outputDirectory;
}

I moved the check and exception to audioRecorder's start method because it's closest to the manifestation of the problem and agnostic to precisely when the output directory is set upstream.

jwe-apple:
Ok. This works - I agree that the outputDirectory being nil on a step with an audio recorder is definitely a programmer error.

On reflection, though, this still doesn't completely address the original issue, because there are multiple ways that audio recording can fail apart from the directory not being set. A UI adjustment in the audio step is also needed in the case that the audio recorder generates an error, to ensure we don't display "Too Loud" or any bogus audio volume indications. Could you have a look and see if you can address that?

Since this is a pre-defined active task, a user visible alert could be appropriate.

coxy1989:
How do you feel about that as a fix to part one (preventing bogus audio indications).

Re: a user visible alert, I see two possibilities

1) UIAlertView from the AudioStepViewController

2) New  (something like.. - (void)displayAlertWithError: (NSError *)error) method on the audioContentView to update the existing alert label that currently shows "Too Loud" with a generic error message.

Both obviously require maintaining copy for the messages. Option 2 would likely need new error types also to distinguish between "Too Loud" and generic errors and display the correct string.

jwe-apple:
Actually I'm happy to accept this without the alert view. Let's take this change without the alerting, and open a new PR for displaying an alert. I think re-using the "Too Loud" label would make sense, rather than an alert view. The extra string key is fine.

PR #114, opened 2015-04-27
Add new “tone audiometry” active task

vtourraine:
This pull-request adds a completely new active task: tone audiometry.

> “Pure tone audiometry (PTA) is the key hearing test used to identify hearing threshold levels of an individual, enabling determination of the degree, type and configuration of a hearing loss.”
> [Wikipedia](http://en.wikipedia.org/wiki/Pure_tone_audiometry)

Practically speaking, this task generates a series of “pure” sinusoid sounds, with different frequencies and on different channels (left or right), starting from the minimum volume, and being gradually increased until the participant presses a button. The current sound amplitude at this point is recorded, and associated with its frequency and channel, constitutes the step result.

You can test this new task directly with the `ORKCatalog` application.

We also wrote a [blog post](http://blog.shazino.com/articles/dev/researchkit-new-active-task/) explaining how to create a new active task. We hope other people might find it useful.

We’re really looking forward to hearing (pun intended) your feedback:

- Significance of the sound amplitude: we track the audio amplitude, in [Audio Unit](https://developer.apple.com/library/ios/documentation/MusicAudio/Conceptual/AudioUnitHostingGuide_iOS/ConstructingAudioUnitApps/ConstructingAudioUnitApps.html#//apple_ref/doc/uid/TP40009492-CH16-SW4) terms, going from 0 to 1. Audio tests using consumer devices like the iPhone are always hard because the actual audio output is basically impossible to control (user volume adjustment, earphones being used, etc). This is probably the most challenging aspect of this task, but we think it could still be worth using.
- Lack of custom artwork: we simply re-used the artworks from the original audio task.
- Limited localization: only English for now.
- No dashboard interface yet (AppCore): audiograms could be very useful.

Task in action:

ios simulator screen shot apr 27 2015 5 00 41 pm

Results displayed in the `ORKCatalog` app:

ios simulator screen shot apr 27 2015 5 00 53 pm


jwe-apple:
@vtourraine Please send an email to researchkit@apple.com so we can get your contributor agreement sorted.

ResearchKit/ActiveTasks/ORKAudioGenerator.h
```diff
@@ -0,0 +1,77 @@
+/*
+ Copyright (c) 2015, Apple Inc. All rights reserved.
```
jwe-apple: Copyrights need to be adjusted to the copyright of the actual copyright holder.

ResearchKit/ActiveTasks/ORKAudioGenerator.m
```diff
@@ -0,0 +1,225 @@
...
+    ok = [audioSession setCategory:AVAudioSessionCategoryPlayback error:&setCategoryError];
+    NSAssert1(ok, @"Audio error %@", setCategoryError);
+    [[NSNotificationCenter defaultCenter] addObserver:self
+                                             selector:@selector(handleInterruption:)
+                                                 name:AVAudioSessionInterruptionNotification
+                                               object:audioSession];
+}
+
+- (void)createToneUnit {
+    // Configure the search parameters to find the default playback output unit
```
jwe-apple: What's the original source of this code? I've found this exact comment all over the web, so this code has probably come from an example.

ResearchKit/ActiveTasks/ORKToneAudiometryStepViewController.m
```diff
@@ -0,0 +1,210 @@
...
+    CGFloat progress = 0.001 + (CGFloat)self.currentTestIndex/(self.testingFrequencies.count * 2);
+    [self.toneAudiometryContentView setProgress:progress
+                                        caption:(channel == ORKAudioChannelLeft) ? [NSString stringWithFormat:ORKLocalizedString(@"TONE_LABEL_%@_LEFT", nil), frequency] : [NSString stringWithFormat:ORKLocalizedString(@"TONE_LABEL_%@_RIGHT", nil), frequency]
+                                       animated:YES];
+
+    [self.audioGenerator playSoundAtFrequency:frequency.doubleValue
+                                    onChannel:channel
+                               fadeInDuration:SoundDuration];
+
+    [self performSelector:@selector(testExpired) withObject:nil afterDelay:SoundDuration];
```
jwe-apple: Rather than performSelector:afterDelay:, (which in any case would need a +cancelPreviousPerformSelector: somewhere to match in case of early dealloc), let's use a dispatch_after with a weak reference.

ResearchKit/ActiveTasks/ORKToneAudiometryStepViewController.m
```diff
@@ -0,0 +1,210 @@
...
+- (ORKToneAudiometryStep *)toneAudiometryStep {
+    return (ORKToneAudiometryStep *)self.step;
+}
+
+- (void)startCurrentTest {
+    const NSTimeInterval SoundDuration = self.toneAudiometryStep.toneDuration;
+
+    NSUInteger frequencyIndex = self.currentTestIndex/2;
+    if (frequencyIndex >= self.testingFrequencies.count) {
+        return;
```
jwe-apple: If you actually hit this case in testing, it would be indicative of a bug, correct? Should this be an assert?

ResearchKit/ActiveTasks/ORKToneAudiometryStepViewController.m
```diff
@@ -0,0 +1,210 @@
...
+    }
+
+    NSNumber *frequency = self.testingFrequencies[frequencyIndex];
+    ORKAudioChannel channel = (self.currentTestIndex%2 == 0) ? ORKAudioChannelLeft : ORKAudioChannelRight;
```
jwe-apple: Coding standard is to have spaces around operators like '%' or '/'. Personally I like having parentheses around such operators if they yield arguments to ==, for clarity.

ResearchKit/Common/ORKOrderedTask.h
```diff
@@ -284,6 +284,35 @@ typedef NS_OPTIONS(NSUInteger, ORKPredefinedTaskOption) {
...
 
+/**
+ Returns a predefined task that tests tone audiometry.
+
+ In a tone audiometry task, the participant is asked to listen to some tones
+ with different audio frequencies, playing on different channels (left and right),
+ with the volume being progressively increased until the participant taps a button.
+
+ A tone audiometry task can be used to measure properties of the user's hearing,
+ based on its reaction to a wide range of frequencies.
```
jwe-apple: s/its/their/

samples/ORKCatalog/ORKCatalog/TaskListRow.swift
```diff
@@ -289,6 +294,9 @@ enum TaskListRow: Int, Printable {
             
             case .Form:
                 return formTask
+
+            case .ToneAudiometry:
```
jwe-apple: Since this is sample code, it's helpful to retain the same order everywhere. If you feel the right place is after audio, put this case just below audio.

samples/ORKCatalog/ORKCatalog/TaskListRow.swift
```diff
@@ -134,6 +135,9 @@ enum TaskListRow: Int, Printable {
 
             case .Form:
                 return NSLocalizedString("Form", comment: "")
+            
+            case .ToneAudiometry:
```
jwe-apple: Same comment here, order is important.

samples/ORKCatalog/ORKCatalog/TaskListRow.swift
```diff
@@ -56,6 +56,7 @@ enum TaskListRow: Int, Printable {
     case Survey
     case Consent
     case Form
+    case ToneAudiometry
```
jwe-apple: Order is important - place after Survey.

jwe-apple:
I'm also expecting a patch to Localizable.strings in English with all the new keys you expect.

jwe-apple:
`AUTOGENERATED MESSAGE by pull-request-mailer`
This pull request has been converted to an email thread on the mailing list researchkit-dev@lists.apple.com. Discussion continues there.

**Please do not post comments to this pull request.** The email thread will not get notified.

jwe-apple:
A little more feedback, commenting on your RFC items:

> Significance of the sound amplitude: we track the audio amplitude, in Audio Unit terms, going from 0 to 1. Audio tests using consumer devices like the iPhone are always hard because the actual audio output is basically impossible to control (user volume adjustment, earphones being used, etc). This is probably the most challenging aspect of this task, but we think it could still be worth using.

Totally agree, we love the idea. The direct approach seems like a good first start. Wonder if tests involving masking, where the relative volume of a masking stimulus and a target stimulus is defines the condition, might be better suited to environments without a good baseline. Have you looked into these?

> Lack of custom artwork: we simply re-used the artworks from the original audio task.

Fine start.

> Limited localization: only English for now.

This will get taken care of during our next convergence round, but you're welcome to provide a first pass at French if you like!

> No dashboard interface yet (AppCore): audiograms could be very useful.

Please open an issue for this!

ResearchKit/ActiveTasks/ORKToneAudiometryStepViewController.m
vtourraine: I agree, but in this case I need to easily cancel the previous invocation when the participant presses the button (I’m not aware of any simple solution to cancel a `dispatch_after`). So I’ve updated the `dealloc` implementation instead. If you have a better solution, please let me know.

jwe-apple: If using dispatch_after, just maintain an ivar to track whether action is required. On user cancel, reset the ivar. Assuming all work is done on the main queue, this is pretty simple.

Generally with ARC, prefer not to use performSelector: variants if we can avoid it, as it's error prone.

vtourraine: I’ve updated the code with a local variable instead of an ivar, because this method is being called repeatedly, so we need to track somehow how to cancel the matching invocation. I hope it make sense.

jwe-apple:
ok to test

ResearchKit/Common/ORKResult.m
```diff
@@ -178,6 +178,99 @@ - (NSString *)description {
 @end
 
 
+@implementation ORKToneAudiometryResult
+
+- (void)encodeWithCoder:(NSCoder *)aCoder {
```
jwe-apple: Ah. This is missing JSON serialization support for both result and step in ORKESerialization. In any case, you need to run the tests attached to the ORKTest testing app.

ResearchKit/Localized/en.lproj/Localizable.strings
```diff
@@ -166,6 +166,15 @@
 "AUDIO_INTENDED_USE" = "This activity evaluates your voice by recording it with the microphone at the bottom of your phone.";
 "AUDIO_TOO_LOUD_LABEL" = "Too Loud";
 
+/* Tone audiometry active task. */
+"TONE_AUDIOMETRY_TASK_TITLE" = "Tone Audiometry";
+"TONE_AUDIOMETRY_INTENDED_USE" = "This activity evaluates your hearing by testing it with different types of sounds.";
+"TONE_AUDIOMETRY_INTRO_TEXT" = "Put your headphones on, and turn up the volume to a sensible level. Tap the button every time your start hearing a sound.";
```
jwe-apple: last "your" should be "you"

ResearchKit/Localized/en.lproj/Localizable.strings
```diff
@@ -166,6 +166,15 @@
...
+"TONE_AUDIOMETRY_CALL_TO_ACTION" = "Tap Get Started to begin.";
+"TONE_AUDIOMETRY_INSTRUCTION" = "Tap the button every time your start hearing a sound.";
```
jwe-apple: "your" should be "you"

samples/ORKCatalog/ORKCatalog/ResultTableViewProviders.swift
```diff
@@ -470,6 +473,62 @@ class TappingIntervalResultTableViewProvider: ResultTableViewProvider {
     }
 }
 
+/// Table view provider specific to an `ORKToneAudiometryResult` instance.
+class ToneAudiometryResultTableViewProvider: ResultTableViewProvider {
```
jwe-apple: I appreciate having this here in ORKCatalog, but we also need this case added to ORKTest. Localization testing uses the ORKTest app only.

ResearchKit/ActiveTasks/ORKToneAudiometryContentView.m
```diff
@@ -0,0 +1,184 @@
...
+    [self setNeedsUpdateConstraints];
+}
+
+- (void)updateConstraints {
+    if ([_constraints count]) {
+        [NSLayoutConstraint deactivateConstraints:_constraints];
+        _constraints = nil;
+    }
+
+    ORKScreenType screenType = _screenType;
```
jwe-apple: This constraint code does not actually need to be dynamic. Move to a -setupConstraints method, called in the initializer, that creates and activates the constraints. We're gradually changing this elsewhere too, to avoid unnecessary layout re-calculation, but let's fix it here now.

vtourraine:
Alright, thanks again for the feedback. I’ve added tests for the new step and sample/result classes. 

GitHub tells me that the internal Jenkins integration is failing (it was already failing before I added these new tests). I cannot see the logs, is there something I can do about that?

jwe-apple:
Yeah, you can ignore that. It's mainly just assistance for me and I'll let you know if it turns out not to be noise.

jwe-apple:
This is almost there! Looking good.

jwe-apple:
So, I tried this out on a few devices and:

1- What do I do if I can't hear a sound? Unclear from a usability perspective.
2- On the devices I tested with, I can't actually hear the 125 Hz - are you able to hear it?
3- It feels like this test won't convey much information, because rate of increase in volume is too non-linear in perception.

What are your thoughts?

vtourraine:
1. The usability could be vastly improved, there’s no doubt about that. If the user doesn’t press the button for a particular sound, the controller moves on to the next channel/frequency, and doesn’t record any amplitude value (we could also record a `nil` value, I can see pros/cons with both approaches). On a usability level, what would you suggest?
2. In good conditions, I can; but it’s true that —according to Wikipedia™— audiometry often tests “between 250 hertz (Hz) and 8 kHz” ([source](http://en.wikipedia.org/wiki/Pure_tone_audiometry)), so I think I’ll remove the 125 Hz value. Could be a step parameter.
3. Good point, the increase is linear, but the perception isn’t. I’ll update the math.

ResearchKit/ActiveTasks/ORKAudioGenerator.m
```diff
@@ -0,0 +1,244 @@
...
+@interface ORKAudioGenerator () {
+    @public
+    AudioComponentInstance _toneUnit;
+
+@public
+    double frequency;
+    double theta;
+    ORKAudioChannel activeChannel;
+    double fadeInFactor;
+    NSTimeInterval fadeInDuration;
```
jwe-apple: Let's rename these ivars to start with underscores. Sorry I missed this earlier!

jwe-apple:
I've reviewed this again. The audio level increase is perceptually more as expected. A couple things I noticed:

- The progress bar is being used to indicate overall progress through the task, rather than showing a timer for this particular audio setting. I think using it as a timer for the current stimulus, as it is used elsewhere, would be more consistent. If you feel it's important to show progress through the task overall, perhaps an extra line that says "Stimulus 5/10" or similar, that increments after each stimulus completes?

- I wonder about calibration. You say in the instructions to use a "sensible" volume. I wonder if this task needs an extra step in which you play a pre-set volume at a known frequency and ask the user to adjust their volume setting until it is *just* perceptible. Then the overall result would therefore be more correctly calibrated (see next point). I'd suggest perhaps 1kHz at volume 0.25, in stereo?

- I found that when set to maximum volume, all the frequencies 1kHz and higher were immediately audible from the start. Clearly this does not provide informative results. I found that in this configuration all the threshold outputs were < 0.05.

Further regarding the calibration. Another possibility is to use a QUEST-like algorithm, with binary inputs (can hear / can't hear) to determine the perceptual threshold, rather than relying on response time. This might provide more reliable data but take a little longer. I don't suggest we do this now - still interested in accepting this PR with the above comments addressed.

vtourraine:
- Progress bar: actually, I think the participant should _not_ see any progress indicator for each frequency. We don’t want him to second-guess his hearing because he can see that he’s past 50%, for instance. That said, we could switch from a progress bar to a label (“Stimulus n/N”). I think the bar is fine as it is, but if you prefer a label, I don’t have any objection.

- Calibration: asking the participant to calibrate/configure the session himself is obviously problematic, but not a dead-end. I’ve switched from “sensible” to “comfortable” volume (slightly less ambiguous, I think), and added a custom instruction step/view controller to play a “normal” tone, as you suggested. On a related note, I’ve added a `outputVolume` property to the result class, to save the device volume for the test. I think that will make the results a lot more interesting, no matter what the user select for the calibration.

- Most people have no problem hearing sounds above 1 kHz, so yes, the results are less interesting (which is a good sign: your hearing is good), but that matches what audiometry tests usually evaluate. Ultimately, we could modify the fade-in function based on the frequency, or even let the app developer customize it depending on the type of hearing problems he’s studying, but I’d rather push this simpler approach first.

Thanks again for the feedback. I really think it’s getting to a good shape, and hopefully we’ll manage to ship it with ResearchKit 1.1, to actually get people to use it and give us more feedback.

jwe-apple:
Thanks for the update. Will get formal UI review for this early next week.

jwe-apple:
Tested this and it seems to do what it says, but I'm going to wait for UI review before accepting, due to concerns around the timer, and around playing a tone without user control during the second instruction step.

vtourraine:
Sure, thanks for taking the time to review this.

rsanchezsaez:
Hey @vtourraine, congratulations on the pull request.

I've been drafting a [ResearchKit Coding Style guide](https://github.com/rsanchezsaez/ResearchKit/wiki/ResearchKit-Coding-Style). I'd be great if you could have a look and provide any relevant feedback you may have (via [Issue #174](https://github.com/ResearchKit/ResearchKit/issues/174) or the mailing list).

vtourraine:
Thanks, @rsanchezsaez. I’ve seen your draft, and I largely agree with the current proposition. This PR already complies with it. I’ll keep an eye on the discussion, but for now, it looks good to me.

jwe-apple:
I'm going to merge this so we have it in the convergence branch.

jwe-apple:
Merging now so it's in the convergence branch for localization.

rsanchezsaez:
I think that some unresolved merge conflicts were committed when merging. I've fixed them as part of [pull request #182](https://github.com/ResearchKit/ResearchKit/pull/182).

vtourraine:
Thanks @jwe-apple, glad to see this pull-request merged. I’m looking forward to the localization, and any feedback from people interested in this active task. In the meantime, I’ve created an issue in AppCore for an audiogram dashboard item, in order to visualize the results from the audiometry.

PR #116, opened 2015-04-27
Complete whitespace homogenization

rsanchezsaez:
Second half of [pull request #112](https://github.com/ResearchKit/ResearchKit/pull/112). This should be pretty much it.

I may have missed some stuff, but it’s definitely much more homogenous than before. I also did some *naming conventions* renames.

I’ll resolve conflicts on my other pull request. If you want me to resolve conflicts on any of the pending pull request by others, I can do so as well.


jwe-apple:
Apart from the two nit picks where I prefer C-style comments, looks good.

rsanchezsaez:
Reverted.

I think I may have changed some other multi-line comments to single-line ones elsewhere, but I think all the rest were inside a method body, so those should not be as much of a problem

jwe-apple:
Ok - merging (against my better judgment!)

Issue #122, opened 2015-04-28
taskViewController(_:shouldPresentStep:) crashes in Swift


getaaron:
In this delegate method:

    func taskViewController(taskViewController: ORKTaskViewController, shouldPresentStep step: ORKStep) -> Bool {
        // use step
    }

`step` is guaranteed by the method signature to be non-nil, and crashes when using `step` if it's not. But this method has many code paths that can return `nil`:

    - (ORKStep *)stepAfterStep:(ORKStep *)step withResult:(ORKTaskResult *)result {
        NSArray *steps = _steps;
        
        if (steps.count <= 0) {
            return nil;
        }
        
        ORKStep *currentStep = step;
        ORKStep *nextStep = nil;
        
        if (currentStep == nil) {
            nextStep = steps[0];
        } else {
            NSUInteger index = [self indexOfStep:step];
            
            if (NSNotFound != index && index != (steps.count-1)) {
                nextStep = steps[index+1];
            }
        }
        return nextStep;
    }

One possible workaround would be to wrap `[ORKTaskViewController -shouldPresentStep:]` in an `if (step) {…}`.

jwe-apple:
Agree, that's the correct fix. `stepAfterStep:withResult:` is correctly annotated as nullable input and nullable result. `shouldPresentStep:` should not be called with a nil `step`.

jwe-apple:
Fixed.

PR #123, opened 2015-04-28
Issue #122: `shouldPresentStep:` method should correctly handle nil step

jwe-apple:
Adjust internal `shouldPresentStep:` method not to call the delegate
on a nil step, and to return NO if the step is nil. When the task has
completed, a nil step indicates this, so handle that case explicitly
rather than depending on a semantically odd `YES` from `shouldPresentStep:`.

Testing: verified correct behavior for forward and backward navigation at beginning and end of tasks in `ORKTest` in simulator.


jwe-apple:
Addresses #122. Test failure is spurious.

jwe-apple:
Whitespace fixed. pushing.

Issue #118, opened 2015-04-27
Fitness task step images change position after transition


rsanchezsaez:
When transitioning to a an `ORKFitnessStep`'s from a previous one. The image (if any) moves to a higher position after the transition finishes. This causes a visual glitch.

This happens on *iPhone 5* and *6* screen sizes. *iPhone 4* screen size is not affected. It happens on *stable* as well, so it isn't a recent regression.

jwe-apple:
Sounds like insufficient constraints. What iOS version are you testing on?

rsanchezsaez:
iOS 8.3. It's 100% reproducible on device and simulator (*ORKCatalog*).

rsanchezsaez:
Looks fixed on the *iPhone 6* simulator.

jwe-apple:
Thanks!

PR #125, opened 2015-04-28
Issue #118: Adjust fitness quantity pair layout

jwe-apple:
This fixes a regression in quantity pair layout. The left view of
the quantity pair, the distance view, formerly always contained
content, which fixed the vertical height of the view. At some point
that content was no longer set, handling the no-pedometer-distance
case. The primary fix here is to ensure that a title and zero
quantity are present in the distance view, even if distance is
not available (as in the simulator).

Secondary fixes fix constraints to make them static, since they
do not need to be dynamically updated. We also introduce new baseline
constraints to better enforce alignment between the left and right
views.

Tested distance-only, heart-rate-only, both, and neither visible on
iPhone 6 on iOS 8.3. Tested with and without a value for heart rate,
with both visible.

Addresses issue #118.

### DIFF
--- a/ResearchKit/ActiveTasks/ORKActiveStepQuantityView.m
+++ b/ResearchKit/ActiveTasks/ORKActiveStepQuantityView.m
@@ -50,6 +50,8 @@
     ORKQuantityLabel *_valueLabel;
     ORKTintedImageView *_imageView;
     UIView *_valueHolder;
+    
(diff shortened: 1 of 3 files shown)


Erin-Mounts:
Looks good. r=me

jwe-apple:
Thanks Erin!

PR #127, opened 2015-04-28
Removing duplicate imports.

richardrauser:


jwe-apple:
ok to test

jwe-apple:
Builds fine. Thanks!

PR #129, opened 2015-04-28
Naming conventions: 'ident' is not an acceptable abbreviation

rsanchezsaez:
`ident` is not an [acceptable abbreviation](https://developer.apple.com/library/mac/documentation/Cocoa/Conceptual/CodingGuidelines/Articles/APIAbbreviations.html#//apple_ref/doc/uid/20001285-BCIHCGAE).


jwe-apple:
Haha. I'm pretty sure the acceptable abbreviations rules apply only to API, not to the internals of methods, but agree this is prettier.

jwe-apple:
But I'm not willing to work for it! Bring it up to date and I'll merge it ;-)

rsanchezsaez:
Haha, updated. :)

I'd say we should try to use acceptable abbreviations and naming conventions inside the framework as well. Given that it's open source some people may look at the code and mimic its style.

PR #130, opened 2015-04-28
LaunchScreen: fix nav bar height and table view content inset

rsanchezsaez:
I fixed *ORKCatalog*'s `LaunchScreen.xib` so it matches the main view controller.

I just added a height constraint of `64` to the navigation bar, and pinned the top of the table to it (so the separator vertical positions are in sync with the live view controller content inset). Tell me if you think there's a better approach instead.


jwe-apple:
Looks good!

jwe-apple:
(Sorry for the delay, just needed a chance to test)

rsanchezsaez:
:)

Issue #87, opened 2015-04-22
Audio predefined task: audio step should not show too loud if outputdir is nil and developer ignored the error


jwe-apple:
[rdar://problem/20503028](rdar://problem/20503028)

Currently, if the output directory is nil, we report an error to the developer via the task view controller delegate. If the developer does not handle the error, the task continues, and the audio recorder setup fails, which results in the audio showing as "too loud".

Instead, the audio step view should display some kind of error state to indicate that there was an error setting up audio. This should be a useful prompt to the developer to do some error handling - though we hope users will never see it.

coxy1989:
Something like

_audioRecorder.outputDirectory == nil ? [self functionWhichCallsSuspendAndShowsUIAlertView] : [self start];

in ORKAudioStepViewController's ViewDidAppear?

jwe-apple:
Probably without the alert view, because if we have an alert view it needs copy which we need to localize and maintain. But yes, roughly that.

PR #136, opened 2015-05-01
Alert the user when an audioRecorder generates an error

coxy1989:
#87

### DIFF
--- a/ResearchKit/ActiveTasks/ORKAudioContentView.m
+++ b/ResearchKit/ActiveTasks/ORKAudioContentView.m
@@ -362,8 +362,8 @@ static const CGFloat kValueLineMargin = 1.5;
 
 - (void)updateAlertLabelHidden {
     NSNumber *sample = [_samples lastObject];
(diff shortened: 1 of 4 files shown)


ResearchKit/ActiveTasks/ORKAudioContentView.m
```diff
@@ -356,7 +362,7 @@ - (void)updateGraphSamples {
 
 - (void)updateAlertLabelHidden {
     NSNumber *sample = [_samples lastObject];
-    BOOL hide = _finished || !([sample doubleValue] > _alertThreshold);
+    BOOL hide = _finished || !_failed || (sample && !([sample doubleValue] > _alertThreshold));
```
jwe-apple: This condition is not right, and also hard to parse. I'd prefer:

BOOL show = (! _finished && ([sample doubleValue] > _alertThreshold)) || _failed;
_alertLabel.hidden = !show;

jwe-apple:
Awesome, thanks!

PR #137, opened 2015-05-01
Vertical scale improvement

rsanchezsaez:
I implemented a very small optimization to the vertical slider. See first commit for details.

Also, I noticed that scrolling was enabled for vertical scale question steps on iPhone 6, due to the scale view being just 5 pixels taller than the no-scroll maximum. So I increased the `kVerticalSliderHorizontalMargin` a tiny bit so it fits without scrolling. @brucehappy: do you think this is a sensible modification?


jwe-apple:
This looks ok to me, including the hack. It might also be nice to avoid scrolling on the iPhone 5, could make the correction value a metric?

jwe-apple:
Checked - false negative from the build.

brucehappy:
Why not increase kVerticalSliderHorizontalMargin in ORKScaleSliderView from 48 to 50 or 52?  Since its already a magic number itself, seems like changing it might be a better solution than adding a new magic number into the ORKScaleSlider.  And then to point that @jwe-apple is making, we could just make the kVerticalSliderHorizontalMargin value a metric in the Skin.  Thoughts?

rsanchezsaez:
@brucehappy: Good idea, I did just that.

Corresponding value to avoid scrolling on iPhone 5 added (scrolling is only avoided on single-line questions, multi-line ones will scroll). Because of this, I replaced the English localization `"Your question goes here."` by `"Your question here."` so it fits within a single-line on an iPhone 5 screen.

jwe-apple:
Yep, ok.

Issue #138, opened 2015-05-02
Improved Survey Skip Logic


hershman:
For the Asthma Health app, we ran into a couple limitations in implementing the daily survey.  There are two specific features related to survey skip logic we'd love in order to improve the experience for our users:

1. __Cross Survey Skip Logic__: We're using the daily survey to asses if users took their controller medication.  Users who aren't prescribed controller medication shouldn't see these questions.  A question from the medication survey could be used to detect these users, but it's not clear how we can link the medication response in the context of the daily survey.

2. __Multi-Response Skip Logic__: In the daily survey we have a question about symptoms during the day and another about night. If the user answers yes to either of these, we would like to ask a follow-up question, but we're not sure how to access the responses to multiple questions when defining a skip rule.

rsanchezsaez:
Regarding *Multi-Response Skip Logic*, that will be possible when we finish the *ordered task with navigation rules* implementation ([issue #86](https://github.com/ResearchKit/ResearchKit/issues/86)). We're working on it already.

On *Cross Survey Skip Logic*, I'm not sure if *ResearchKit* should handle this. When you finish a survey, *ResearchKit* provides you with the results, and the app is responsible for handling and/or storing them. *ResearchKit* doesn't relate independently run tasks. Maybe this part could be an *AppCore* feature request? (@jwe-apple: please correct me if I'm wrong).

jwe-apple:
@rsanchezsaez Cross Survey Skip Logic sounds like something we should think about including in the navigation rules task already in progress. Two ways it could be achieved:

1) Providing an override point where a subclass could use a parameter to control.

2) Providing a way to inject extra identifiers into the namespace for predicates, so that a predicate can operate against an additional result set, perhaps from another task entirely.

I think I prefer the latter but an experiment in writing an example is probably merited.

rsanchezsaez:
I'll see what I can do.

jwe-apple:
I believe both have been addressed by PR #151

Issue #86, opened 2015-04-22
Extend ORKOrderedTask to include navigation rules


jwe-apple:
Either subclass `ORKOrderedTask`, or extend it, to include navigation rules.

A navigation rule would be something which can be asked if it matches the current step. If it matches the current step, then if provided with the results of the task to date, it should provide the step identifier for the next step. A concrete navigation rule class might be one checks if the result is equal to a particular question result. Although a block might be nicer, it would not be possible to serialize, so we prefer direct equality checks for now. A future generalization might be the use of predicates.

Any navigation rule classes should be serializable with `NSSecureCoding`, and serialization support in `ORKESerialization` in the ORKTest project would also be appropriate.

This task is to propose suitable API, implement, and ideally write unit tests.

rsanchezsaez:
I'm working on this while the *iPad margins pull request* is reviewed.

jwe-apple:
Sounds good. Please email me a draft of API for review as soon as you have something to look at. Can likely save you implementation time if we get the API close to right the first time.

jwe-apple:
(and also I can circulate it more widely here)

rsanchezsaez:
Some ideas for the API:

    @interface ORKStepNavigationRule : NSObject

    - (instancetype)initWithSourceStepIdentifier:(NSString *)sourceStepIdentifier;
    - (NSString *)nextStepIdentifierWithTaskResult:(ORKTaskResult *)taskResult;

    @property (nonatomic, copy, readonly) NSString *sourceStepIdentifier;


    @end


    @interface ORKStepResultEvaluator : NSObject

    - (instancetype)initWithStepResultIdentifier:(NSString *)stepResultIdentifier;
    - (BOOL)evaluateWithTaskResult:(ORKTaskResult *)taskResult;

    @end


    @interface ORKStepBooleanResultEvaluator : ORKStepResultEvaluator

    - (instancetype)initWithStepResultIdentifier:(NSString *)stepResultIdentifier expectedAnswer:(BOOL)expectedAnswer;

    @end


    @interface ORKResultEvaluatorStepNavigationRule : ORKStepNavigationRule

    - (instancetype)initWithSourceStepIdentifier:(NSString *)sourceStepIdentifier
                           successStepIdentifier:(NSString *)successStepIdentifier
                           failureStepIdentifier:(NSString *)failureStepIdentifier
                                      evaluators:(NSArray *)evaluators;

    @end

[...]

PR #151, opened 2015-05-07
Add navigable ordered task and step navigation rules

rsanchezsaez:
New feature dealing with issues #86 and #138.

See the issue pages, the header documentation, and the *researchkit-dev* mailing list thread (subject: '*ResearchKit: navigation rules*') for implementation details.

For now, this has one follow-on issue: Implement an *NSPredicate JSON serializer* (#152).


nskboy:
Can't find public link to `researchkit-dev` discussion on this topic. Could you please help?

jwe-apple:
http://lists.apple.com/archives/researchkit-dev/2015/May/index.html

jwe-apple:
Sadly it's not the whole discussion since we only moved to the list later on.

ResearchKit/Common/ORKStepNavigationRule.m
```diff
@@ -0,0 +1,460 @@
...
+static void ORKValidateResultIdentifiersUnique(NSArray *results, NSString* exceptionReason) {
+    NSArray *uniqueIdentifiers = [results valueForKeyPath:@"@distinctUnionOfObjects.identifier"];
+    BOOL itemsHaveNonUniqueIdentifiers = ([results count] != [uniqueIdentifiers count]);
+    if (itemsHaveNonUniqueIdentifiers) {
+        @throw [NSException exceptionWithName:NSGenericException reason:exceptionReason userInfo:nil];
+    }
+}
+
+- (NSString *)identifierForDestinationStepWithTaskResult:(ORKTaskResult *)taskResult {
+    NSMutableArray *leafResults = [[NSMutableArray alloc] initWithArray:[[self class] getLeafResultsWithTaskResult:taskResult]];
```
nskboy: Starting regular method with `get` is not a good idea unless it's a getter

ResearchKit/Common/ORKStepNavigationRule.h
```diff
@@ -0,0 +1,371 @@
...
+ 
+ @param aDecoder    Coder from which to initialize the step navigation rule.
+ @return A new predicate step navigation rule.
+ */
+- (instancetype)initWithCoder:(NSCoder *)aDecoder NS_DESIGNATED_INITIALIZER;
+
+/**
+ An optional array of additional task results.
+ 
+ The predicate step navigation rule will use the child step results within these tasks, in addition to the current task results, to math the result predicates. You must ensure that all the step identifiers within the additional task results are unique, and that they are different from the step identifiers within the current task result.
```
nskboy: Could you please provide a use case where `-additionalTaskResults` is useful?

rsanchezsaez: It's useful if you want to add conditional navigation logic to your task, based on the results on one or more previously completed task. E.g., if the participant answers an "About you" task the first day of study, you could then have other tasks with common parts and specific questions for different participant parameters.

This is not currently used in the `ORKTest`'s interactive example, but you can see a small non-interactive example in the `testPredicateStepNavigationRule` method of `ORKTaskTest`.

nskboy: Oh, great! Please document this use case in header file.

ResearchKit/Common/ORKStepNavigationRule.m
rsanchezsaez: Very good point, method renamed.

jwe-apple:
This is still out for API review, but it's a pretty busy time here. I'll take a detailed look myself tomorrow.

rsanchezsaez:
No problem.

Testing/ORKTest/ORKTest/ORKESerialization.m
```diff
@@ -295,541 +296,570 @@ @implementation ORKESerializer
...
+            PROPERTY(signature, ORKConsentSignature, NSObject, NO, nil, nil)
+            })),
+   ENTRY(ORKCollectionResult,
+         nil,
+         (@{
+            PROPERTY(results, ORKResult, NSArray, YES, nil, nil)
+            })),
+   ENTRY(ORKTaskResult,
+         ^id(NSDictionary *dict, ORKESerializationPropertyGetter getter) {
+             NSLog(@"blah");
```
jwe-apple: Drop the log

Testing/ORKTest/ORKTest/ORKESerialization.m
```diff
@@ -233,6 +233,7 @@ static id propFromDict(NSDictionary *dict, NSString *propName) {
                 NSCAssert(convertedValue != nil, @"Could not convert to object of class %@", propertyClass);
                 outputDictionary[key] = convertedValue;
             }
+            output = outputDictionary;
```
jwe-apple: oopS!

jwe-apple:
Ok, I've conducted some further review. Here are some comments!

Navigable ordered task feedback

* Please reinstate manual line wraps in headerdoc comments
* Don't need to describe ORKNavigableOrderedTask in the doc for ORKOrderedTask. There is a reference to the subclass in the generated documentation anyway for readers that are interested.
* ORKNavigableOrderedTask: should this have a method -(ORKStepNavigationRule *)stepNavigationRuleForTriggerStepIdentifier: ?
* In comment for setNavigationRule:forTriggerStepIdentifier:, move the second sentence into the second paragraph. The first paragraph should be a one sentence abstract.
* Let's move ORKNavigableOrderedTask into its own file.
* Method name "updateStepIdentifierStackWithSourceIdenfier" - spelling
* Progress for ORKNavigableOrderedTask: I'd consider showing k / N, using the index of the step in the task. It might skip or move backward as the user progresses, but this is still fundamentally an ordered task.
* ORKNavigableOrderedTask has become stateful here. If one is re-used in a second task, it should cleanly reset itself. You should handle cleanly the case where sourceIdentifier (spelling of var name too!) is nil, but the step identifier stack is non-nil.
* Wonder if the task state should be separated from the navigable task. So, you ask the navigable task object for a TaskState object, which itself implements the protocol and owns a reference to the navigable task. And *that* is what you attach to the task view controller. It just seems odd having this navigable task object which is stateful, when none of our other objects are stateful in the same way.
* Documentation of `ORKResultPredicate` abstract is ok, but it needs more description beyond that. *Why* build a predicate in the first place? This should refer to ORKNavigationRule, for instance.
* Don't use a headerdoc comment for unavailable methods. We don't want them to appear in the documentation.
* For all these predicates, how does one specify an integer which should be ignored? Or an NSTimeInterval?
* On ORKStepNavigationRule, remove the headerdoc on new and init
* By limiting to only one navigation rule per source step identifier, you're preventing ternary decision trees. I would consider allowing an array of navigation rules per step identifier, or perhaps better, having a composite navigation rule which combines an array of navigation rules, which can each have an opportunity to match in turn.
* The documentation for additionalTaskResults needs to say why I should use this, i.e. document the fact that it's intended for allowing a task to have different behavior depending on the results of a previous or related task the user has already completed.
* Any ideas for how we could relax the limitation that the task results have step identifiers that don't overlap the step identifiers in this task? We've thought of task identifiers as namespacing step identifiers in the past, so limitation would be unfortunate.
* The ORKDirectStepNavigationRule should expose its destination identifier as a read-only property.
* ORKResultPredicate probably needs its own header
* predicateForTimeIntervalQuestionResultWithIdentifier: this should use NSTimeInterval, and should have both minimum and maximum values.
* predicateForTimeOfDayQuestionResultWithIdentifier - this feels clunky. I'd almost rather just leave it out.

jwe-apple:
Overall this is looking really good, and some of the above, if you agree they sound useful, could certainly be opened as follow-on issues while allowing this to be merged.

rsanchezsaez:
Thanks for the feedback, I'll work on it during the weekend.

rsanchezsaez:
I agree with most of the feedback. I think we may as well take as many of them as possible on this PR (the simple ones for sure).

Some comments on a few of the items:

> - Progress for ORKNavigableOrderedTask: I'd consider showing k / N, using the index of the step in the task. It might skip or move backward as the user progresses, but this is still fundamentally an ordered task.

Hmm, I think I don't agree on this one. For example, see `ORKTest`'s example navigable task. The number of total steps is considerably higher than the actual number of steps the participant would see, no matter what path is taken. So the total progress number would always be wrong. The current step will also jump around like crazy. I think this would just be noise for the participant, rather than helpful.

> Wonder if the task state should be separated from the navigable task. So, you ask the navigable task object for a TaskState object, which itself implements the protocol and owns a reference to the navigable task. And that is what you attach to the task view controller. It just seems odd having this navigable task object which is stateful, when none of our other objects are stateful in the same way.

I agree with your concern about it being the only stateful logic object in the framework. However, I don't fully get how having a separate `TaskState<ORKTask>` object improves things. Given that `ORKNavigableOrderedTask` inherits from `ORKOrderedTask` and thus implements `<ORKTask>` as well, why would using a `TaskState<ORKTask>` object on the task view controller be better? Maybe I'm misunderstanding something.

> For all these predicates, how does one specify an integer which should be ignored? Or an NSTimeInterval?

How do you mean? If you want to ignore a step, you just don't add a predicate for it, no?

> By limiting to only one navigation rule per source step identifier, you're preventing ternary decision trees. I would consider allowing an array of navigation rules per step identifier, or perhaps better, having a composite navigation rule which combines an array of navigation rules, which can each have an opportunity to match in turn.

But isn't this use case dealt with `ORKPredicateStepNavigationRule` being able to have several `resultPredicates` with corresponding `matchingStepIdentifiers`? The predicates will be checked in order until one of them is matched.

> Any ideas for how we could relax the limitation that the task results have step identifiers that don't overlap the step identifiers in this task? We've thought of task identifiers as namespacing step identifiers in the past, so limitation would be unfortunate.

Good point, I'll look into that.

> predicateForTimeOfDayQuestionResultWithIdentifier - this feels clunky. I'd almost rather just leave it out.

Yeah, there are too many `NSDateComponents` for covering all of them. But I did this because of this comment in `ORKTimeOfDayQuestionResult`'s *HeaderDoc*: `/** Typically only hour, minute, and AM/PM data are of interest. */`. Couldn't this method be useful anyway? If the developer needs more, she can just define further methods or manually create more predicates.

---

The rest of comments I fully agree. I'll start working on the appropriate changes.

jwe-apple:
We talked about this offline, and are on the same page for changes to come.

rsanchezsaez:
> On ORKStepNavigationRule, remove the headerdoc on new and init

Should I do the same on `ORKRecorderConfiguration`? It has a similar comment.

rsanchezsaez:
> Any ideas for how we could relax the limitation that the task results have step identifiers that don't overlap the step identifiers in this task? We've thought of task identifiers as namespacing step identifiers in the past, so limitation would be unfortunate.

I've managed to rework how predicates are built, and they now must have a *task result identifier* and a *question result identifier*. This supports question results with the same identifier as long as they belong to different task results. I like the code better now, because the new predicates match against the vanilla `ORKTaskResult` structures, with no flattening needed. You can cross match question result from different `ORKTaskResult` as part of the same predicate freely, which is nice.

See the [corresponding commit] (https://github.com/rsanchezsaez/ResearchKit/commit/977dab78dc0121ba7536a4d87324a8272281675c) for additional commentary. 

---

With this I think I took care of all your feedback, with the exception of this:

> Wonder if the task state should be separated from the navigable task. So, you ask the navigable task object for a TaskState object, which itself implements the protocol and owns a reference to the navigable task. And that is what you attach to the task view controller. It just seems odd having this navigable task object which is stateful, when none of our other objects are stateful in the same way.

As we discussed, `ORKNavigableOrderedTask` should gracefully reset its navigation stack if it's reused, so I don't think it's much of a problem having it like this. Also, do note that although the stack is currently being saved in *key archiving* and *JSON serialization*, it's not not being used in the `isEqual:` and `hash` methods.

But if you have better idea to try and keep things more stateless, I'm all ears. I don't yet see how a separate *task state object* or a *task generator* could make things cleaner.

jwe-apple:
I've not been through your most recent commit in detail yet. One quick question that popped up on first viewing - do you think it would be nicer for the common case if task identifier could be nullable, and a nil task identifier could match the current task's result?

rsanchezsaez:
> I've not been through your most recent commit in detail yet. One quick question that popped up on first viewing - do you think it would be nicer for the common case if task identifier could be nullable, and a nil task identifier could match the current task's result?

Yeah, I initially thought so too. But the problem is that the task identifier needs to be embed into the predicate when you build it, and you don't know the current task identifier at this time (you can potentially build predicates anywhere, but you only know the current task result identifier on the `stepAfterStep:withResult:` method).

One potential solution to achieve *task identifier nullability* would be to turn `ORKResultPredicate` into an instantiable class with a `taskIdentifier` property, which could then be nil-checked by `ORKPredicateStepNavigationRule` and matched accordingly. But this would make compounding result predicates from different tasks much more cumbersome (or not possible at all).

jwe-apple:
@rsanchezsaez I'd imagined using NSPredicate's `-predicateWithSubstitutionVariables:` on the predicates to substitute the task identifier. This works right down through predicates. So you'd use `$CURRENT_TASK_IDENTIFIER` right down through all the predicates when passed a nil predicate, and then pass `@{@"CURRENT_TASK_IDENTIFIER" : task.identifier }` . I've not tried it but the documentation indicates you can do this on a root compound predicate, and it gets applied through the entire expression hierarchy.

See [https://developer.apple.com/library/ios/documentation/Cocoa/Conceptual/Predicates/Articles/pCreating.html](Creating Predicates), search for `predicateWithSubstitutionVariables:`.

rsanchezsaez:
Oh, that's a a good idea. I'll look into it.

rsanchezsaez:
I made `taskIdentifier` it nullable. I also added `ORKResultPredicate` convenience methods without the `taskIdentifier` argument, and added the corresponding Unit Tests. See the commit message for more details.

jwe-apple:
```
/Users/Shared/Jenkins/Home/workspace/ResearchKit-GitHub-PullRequests/ResearchKit/ResearchKit/Common/ORKStepNavigationRule.m:182:45: error: use of undeclared identifier 'ORKTaskIdentifierResultPredicateVariableName'
                    substitutionVariables:@{ORKTaskIdentifierResultPredicateVariableName: taskResult.identifier}]) {
                                            ^
1 error generated.
```

jwe-apple:
ok to test

rsanchezsaez:
Sorry for the botched commit. Fixed.

ResearchKitTests/ORKTaskTests.m
```diff
@@ -140,9 +154,158 @@ - (void)setUpOrderedTask {
...
+    _stepNavigationRules[OtherSymptomStepIdentifier] = [directRule copy];
+    
+    directRule = [[ORKDirectStepNavigationRule alloc] initWithDestinationStepIdentifier:nil];
+    [_navigableOrderedTask setNavigationRule:directRule forTriggerStepIdentifier:EndStepIdentifier];
+    _stepNavigationRules[EndStepIdentifier] = [directRule copy];
+}
+
+typedef NS_OPTIONS(NSUInteger, TestsTaskResultOptions) {
+    TestsTaskResultOptionSymptomHeadache    = 1 << 0,
+    TestsTaskResultOptionSymptomDiziness    = 2 << 0,
```
Erin-Mounts: nitpick: two z's in Dizziness (seems to be used with one z throughout the file)

Erin-Mounts:
ok I didn't spot anything new (other than the one nitpick) so r=me

ResearchKitTests/ORKTaskTests.m
rsanchezsaez: Thanks! Fixed.

rsanchezsaez:
Replaced `nil` value by `ORKNullStepIdentifier` to signal that a step navigation rule should finish the ongoing task.

jwe-apple:
r=me

PR #156, opened 2015-05-07
Add optional NSNumberFormatter to scale & continuous scale

scott-ce:
This adds a numberFormatter property to ORKScaleAnswerFormat and ORKContinuousAnswerFormat. Our motivation for this is to enable a continuous slider that displays a percentage, though it could be used for currency or other number formats.

Because this property is optional and likely to be used infrequently, I did not add it to the initializers.

This PR also includes a small change to ORKCatalog to display the second step for Scale Question as a continuous percentage, so you can view the appearance there.

### DIFF
--- a/ResearchKit/Common/ORKAnswerFormat.h
+++ b/ResearchKit/Common/ORKAnswerFormat.h
@@ -84,6 +84,14 @@ typedef NS_ENUM(NSInteger, ORKChoiceAnswerStyle) {
     ORKChoiceAnswerStyleMultipleChoice
 } ORK_ENUM_AVAILABLE;
 
+/// An enumeration of the format styles available for scale answers.
+typedef NS_ENUM(NSInteger, ORKNumberFormattingStyle) {
+    /// The default decimal style.
+    ORKNumberFormattingStyleDefault,
+    
+    /// Percent style.
+    ORKNumberFormattingStylePercent
+} ORK_ENUM_AVAILABLE;
 
 @class ORKScaleAnswerFormat;
 @class ORKContinuousScaleAnswerFormat;
@@ -270,6 +278,11 @@ ORK_CLASS_AVAILABLE
  */
 @property (readonly, getter=isVertical) BOOL vertical;
 
+/**
+ Number formatter applied to the minimum, maximum, and slider values. Can be overridden by subclasses.
+ */
+@property (readonly) NSNumberFormatter *numberFormatter;
+
 @end
 
 
@@ -344,6 +357,16 @@ ORK_CLASS_AVAILABLE
  A Boolean value indicating whether the scale is oriented vertically. (read-only)
  */
 @property (readonly, getter=isVertical) BOOL vertical;
+
+/**
+ Formatting style applied to the minimum, maximum, and slider values.
+ */
+@property ORKNumberFormattingStyle numberStyle;
+
+/**
+ Number formatter applied to the minimum, maximum, and slider values. Can be overridden by subclasses.
+ */
+@property (readonly) NSNumberFormatter *numberFormatter;
 
 @end
 
--- a/ResearchKit/Common/ORKAnswerFormat.m
+++ b/ResearchKit/Common/ORKAnswerFormat.m
@@ -1,5 +1,6 @@
 /*
  Copyright (c) 2015, Apple Inc. All rights reserved.
+ Copyright (c) 2015, Scott Guelich.
  
  Redistribution and use in source and binary forms, with or without modification,
  are permitted provided that the following conditions are met:
@@ -59,6 +60,10 @@ NSString *ORKQuestionTypeString(ORKQuestionType questionType) {
             SQT_CASE(TimeInterval);
     }
 #undef SQT_CASE
+}
+
+NSNumberFormatterStyle ORKNumberFormattingStyleConvert(ORKNumberFormattingStyle style) {
+    return style == ORKNumberFormattingStylePercent ? NSNumberFormatterPercentStyle : NSNumberFormatterDecimalStyle;
 }
 
 @implementation ORKAnswerDefaultSource {
@@ -1180,7 +1185,9 @@ static NSArray *ork_processTextChoices(NSArray *textChoices) {
 
 #pragma mark - ORKScaleAnswerFormat
 
-@implementation ORKScaleAnswerFormat
+@implementation ORKScaleAnswerFormat {
+    NSNumberFormatter *_numberFormatter;
+}
 
 - (Class)questionResultClass {
     return [ORKScaleQuestionResult class];
@@ -1231,7 +1238,17 @@ static NSArray *ork_processTextChoices(NSArray *textChoices) {
     return @(integer);
 }
 - (NSString *)localizedStringForNumber:(NSNumber *)number {
-    return [NSNumberFormatter localizedStringFromNumber:number numberStyle:NSNumberFormatterDecimalStyle];
+    return [self.numberFormatter stringFromNumber:number];
+}
+
+- (NSNumberFormatter *)numberFormatter {
+    if (! _numberFormatter) {
+        _numberFormatter = [[NSNumberFormatter alloc] init];
+        _numberFormatter.numberStyle = NSNumberFormatterDecimalStyle;
+        _numberFormatter.locale = [NSLocale autoupdatingCurrentLocale];
+        _numberFormatter.maximumFractionDigits = 0;
+    }
+    return _numberFormatter;
 }
 
 - (NSInteger)numberOfSteps {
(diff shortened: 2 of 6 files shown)


jwe-apple:
So, you've missed the JSON serialization in ORKESerialization, which is not part of the primary framework but is in the ORKTest test app.

Serializing a number formatter to JSON in a way that is machine-generatable is non-trivial.

I see the appeal of doing this, but one of the goals of the answer format design has been to keep answer formats etc to a structure that *can* be completely serialized to JSON. The only exception to this so far has been for images.

In the numeric answer formatters we used the `ORKNumericAnswerStyle` parameter to select between a few predefined number formatters. I think I'd prefer to follow a similar pattern here.

scott-ce:
So I did!

Wish that were documented outside ORKTest… because I've also written a full JSON to RK deserialization library that I was planning to release soon. Added live UI updating as well, which simplifies editing JSON definition files. Will take a closer look at your implementation tomorrow.

Regarding NSNumberFormatter, in our JSON implementation I just defined string constants corresponding to NSNumberFormatterStyle values, which is similar to the ORKNumericAnswerStyle approach. Figured it might be nice to allow folks to take full advantage of number formatters in code if they wanted, but provide the basics via JSON. But if you want full concordance with JSON I'll reduce it down to a few presets.

jwe-apple:
I think better to reduce to a few presets.

Do have a look at our implementation; see what you think. It's not in the framework proper because we weren't confident it was stable or clean enough.

scott-ce:
I updated the PR to use a style enum instead. Let me know what you think and if you prefer alternate names.

Regarding the JSON serialization, I like your compact DSL approach and declarative way devs can extend the JSON. In contrast we used JSONModel. We don't expose it for out-of-the-box JSON deserialization, but extending the JSON definition with custom classes requires subclassing and following JSONModel conventions.

I'd be curious to hear more about your concerns about ORKESerialization and thoughts on how best to tackle JSON in a general way. How about I share what came up with and then post to the mailing list to discuss alternative approaches?

jwe-apple:
Your change looks great. I'll test it a bit later with the other changes.

Could you send an email to researchkit@apple.com? I need to get your contributor agreement sorted.

Absolutely post to researchkit-dev, let's have a discussion about this.

jwe-apple:
ok to test

ResearchKit/Common/ORKAnswerFormat.m
```diff
@@ -61,6 +61,13 @@ id ORKNullAnswerValue() {
 #undef SQT_CASE
 }
 
+NSNumberFormatterStyle ORKNumberFormattingStyleConvert(ORKNumberFormattingStyle style) {
+    NSDictionary *map =@{ @(ORKNumberFormattingStyleDefault)  : @(NSNumberFormatterDecimalStyle),
```
jwe-apple: Ah. Let's create this map once, and keep it as a static (dispatch_once pattern)

ResearchKit/Common/ORKAnswerFormat_Internal.h
```diff
@@ -1,5 +1,6 @@
 /*
  Copyright (c) 2015, Apple Inc. All rights reserved.
+ Copyright (c) 2015, Scott Guelich.
```
jwe-apple: What have you added to this header?

ResearchKit/Common/ORKAnswerFormat.h
```diff
@@ -84,6 +85,17 @@ typedef NS_ENUM(NSInteger, ORKChoiceAnswerStyle) {
...
+/// An enumeration of the format styles available for scale answers.
+typedef NS_ENUM(NSInteger, ORKNumberFormattingStyle) {
+    /// The default decimal style.
+    ORKNumberFormattingStyleDefault,
+    
+    /// Percent style.
+    ORKNumberFormattingStylePercent,
+    
+    /// Currency style.
+    ORKNumberFormattingStyleCurrency
```
jwe-apple: Hmm, the result would then need to report what currency was used. Otherwise you have no record of the units of measure.

ResearchKit/Common/ORKAnswerFormat_Internal.h
scott-ce: Clearly nothing. :smile: Xcode's class counterparts tracking sometimes gets confused in the latest version, and I probably edited this thinking it was ORKAnswerFormat.h.

ResearchKit/Common/ORKAnswerFormat.h
scott-ce: How about I add another property for overriding the locale used by the number formatter, along with a corresponding property on ORKScaleQuestionResult?

NSNumberFormatter's currencyCode by itself wouldn't account for other formatting issues like grouping/decimal separators, whether to show fractional units, etc. and this would allow folks to override the locale for decimal numbers too, if necessary.

jwe-apple: How about expose the number formatter as a read-only property? Users of the framework can then subclass and override the numberFormatter method to get specific desired behavior, including (for instance) adding a setter.

scott-ce: Sorry didn't see your response earlier. I just committed code that adds locale properties to both scale answer types and the result.

I like the idea of providing a full number formatter that can be subclassed. However, as you pointed out, we still need to return the currency used and NSLocale seems the best way to do that. At this point I could add the read-only number formatter to the step answer formats in addition to the locale, or replace this new locale and just keep it on the answer result. What's your preference?

scott-ce:
Or were you thinking of backing out the style enum as well, and simply having a read-only number formatter that can be overridden?

jwe-apple:
I don't think this works. For one thing, the locale identifier *might* work but it's not really getting at the important element, which is what currency the user saw.

I'd think on the result we'd want a nullable currency code field, holding the output of `[formatter currencyCode]` (a three-letter ISO currency code). I'm torn between having it there as a property and just defining a metadata key for it and including it as metadata on the result. Leaning toward a property.

As far as having a locale on the step, I'm not too happy with that either. Rather have a number formatter property suitable for override, but still expose the enum. If we do have the number formatter, how does the subclasser ensure that their display parameters make it to the result object? One option might be to include the actual formatted string on the result object as well as the numeric value - the "no code" option.

scott-ce:
I'm by no means an expert on NSLocale, but shouldn't `[locale objectForKey:NSLocaleCurrencyCode]` return the same result as `[formatter currencyCode]` here, so long as the currency code isn't explicitly overridden?

jwe-apple:
Yes, but let's say you serialize the result. It's better to have the currencyCode, which tells you the units of the numeric value, than it is to have a locale identifier. For that to be useful, you'd need to know at least the OS version and know the details of any change in the locale specification. For example, if a country were to leave or enter the euro, their currency would get updated at some point - how would you know if the change had happened on that device yet?

scott-ce:
Was curious whether your concern is specifically with NSLocale or JSON serialization. Because if we do provide an extensible NSNumberFormatter subclass, that can't easily be captured in the JSON either, as you mentioned.

Tangentially… right now in our JSON deserialization solution we pass in an image converter (UIImage*(^)(id)) rather than try to come up with a universal solution for image deserialization. I was planning to extend this to support passing in a single object that provided various default converters which could be selectively overridden. With this approach it's not as important to be rigid with the JSON (de)serialization, because developers can customize it as needed.

But ORKESerialization (and any other solution) should still do something reasonable by default, and I hear what you're saying about currency codes. This PR originally started because we needed a way to support percentages… What about dropping ORKNumberFormattingStyleCurrency, keeping .Default and .Percent, providing a readonly number formatter without serialization support, and leaving currency as something that requires subclassing?

jwe-apple:
> On 12 May 2015, at 14:56, Scott G <notifications@github.com> wrote:
> 
> Was curious whether your concern is specifically with NSLocale or JSON serialization. Because if we do provide an extensible NSNumberFormatter subclass, that can't easily be captured in the JSON either, as you mentioned
> 
My idea is to expose the NSNumberFormatter as a read-only property, so that subclasses of the answer format, developed external to the framework, can easily change the behavior. This would not need to be serialized, because code within our framework can always reproduce a number formatter from its settings. From your last paragraph, sounds like you're on the same page.

> Tangentially… right now in our JSON deserialization solution we pass in an image converter (UIImage*(^)(id)) rather than try to come up with a universal solution for image deserialization. I was planning to extend this to support passing in a single object that provided various default converters which could be selectively overridden. With this approach it's not as important to be rigid with the JSON (de)serialization, because developers can customize it as needed.
> 
Ok

> But ORKESerialization (and any other solution) should still do something reasonable by default, and I hear what you're saying about currency codes. This PR originally started because we needed a way to support percentages… What about dropping ORKNumberFormattingStyleCurrency, keeping .Default and .Percent, providing a readonly number formatter without serialization support, and leaving currency as something that requires subclassing?
> 
Perfect, totally on board.

>

scott-ce:
Updated to add a percent style for continuous scales (doesn't work with discrete scales, because of the integer restriction) and a readonly numberFormatter property.

Because we tried out a few approaches and GitHub was reporting merge conflicts, I collapsed this branch down to a single clean commit. Not sure what GitHub is still seeing a conflict, it merges cleanly on my system.

jwe-apple:
I'm happy, just the couple bits of documentation that need improvement. I'll test later on.

jwe-apple:
Tested, looks good, I'm going to merge this one.

Issue #135, opened 2015-05-01
Scale question step loses its default value on second visit


rsanchezsaez:
The scale question step loses its default value on second visit.

Steps to reproduce:
1. Go to a scale question step with a default value: the scale has a default value. Do not input any value.
2. Go back.
2. Go forward again: the scale doesn't have a default value.

brucehappy:
Definitely saw this during my testing, but wrongly assumed it was the intended behavior.

jwe-apple:
Yep, sounds like a bug. note that if you skip the step and then come back, the default value should *not* appear.

rsanchezsaez:
Also, not sure what the expected behaviour is, but when even when a default value shows, the *Next* button is disabled until you interact with the slider.

rsanchezsaez:
The problem seems to be that as soon as you view a question step, the answer is set to `NSNull`. So going back is currently indistinguishable from skipping the question.

To provide the desired behaviour, we could define an empty `ORKSkippedAnswer : NSNull` class which is only set when a question is skipped. That way we can show the default value when the answer is `NSNull`, but avoid showing it when it is `ORKSkippedAnswer`.

This would also keep the current behaviour where: you go to a question, set a value, go back, go forward again, and the step preserves the value that you previously entered.

Does this sound reasonable?

rsanchezsaez:
Fixed via [pull request #157](https://github.com/ResearchKit/ResearchKit/pull/157).

PR #157, opened 2015-05-07
Fix ORKQuestionStepViewController losing its default value when going back

rsanchezsaez:
Fixes [issue #135](https://github.com/ResearchKit/ResearchKit/issues/135)

 When you skip the step, the default value is hidden. Default value doesn't reappear when you go back twice, and go forward again.

I have also fixed a small visual glitch (occurring in stable as well): when you skip a step, the current value was hidden before the step transition. I guess we don't need to update the view just before an outgoing transition: the cell will be fine and empty if the user ever comes back.

### DIFF
--- a/ResearchKit/Common/ORKQuestionStepViewController.m
+++ b/ResearchKit/Common/ORKQuestionStepViewController.m
@@ -127,7 +127,6 @@ typedef NS_ENUM(NSInteger, ORKQuestionSection) {
 }
 
 - (instancetype)initWithStep:(ORKStep *)step {
-    
     self = [super initWithStep:step];
     if (self) {
         _defaultSource = [ORKAnswerDefaultSource sourceWithHealthStore:[HKHealthStore new]];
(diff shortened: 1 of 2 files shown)


jwe-apple:
Hm, now I see this change, I think I may have made a mistake when I commented on the original issue. Really sorry for this.

On the first part of the change, I think it's actually the desired behavior. When you enter an answer, say, "2", and go forward, then go back twice, then forward again, you don't want to see the default. You want to see "2", which was the answer you most recently answered for this question. When you skip, you're saying, my answer for this question is "null". That's actually a first class answer (the "null" answer). If you go back twice and then forward again, you still want to see "null" so you don't need to "enter" it again.

On the second part of the change, we want to see the value being hidden just before the step transition. This "glitch" is to communicate to the user that the value has been cleared by taking the "Skip" action. Otherwise it's not immediately obvious that "Skip" clears the answer.

rsanchezsaez:
> On the first part of the change, I think it's actually the desired behavior. When you enter an answer, say, "2", and go forward, then go back twice, then forward again, you don't want to see the default. You want to see "2", which was the answer you most recently answered for this question. When you skip, you're saying, my answer for this question is "null". That's actually a first class answer (the "null" answer). If you go back twice and then forward again, you still want to see "null" so you don't need to "enter" it again.

Yeah, that's the behavior after my commit. Before the commit, if go back (without entering any value), and go forward again, the default value was hidden. So I think we're on the same page?

> On the second part of the change, we want to see the value being hidden just before the step transition. This "glitch" is to communicate to the user that the value has been cleared by taking the "Skip" action. Otherwise it's not immediately obvious that "Skip" clears the answer.

I beg to disagree: the "*glitch*" doesn't look nice. ;) It's not something I'd expect from an iOS app. What I'd do is add a quick fade animation for clearing the value, and launch the transition animation after the fade completes. But I guess this would not be a quick thing to add, as there are many different types of question cells involved.

I can revert this second part if you want (but I'll make sure the thumb is cleared at the same time as the answer in the scale questions).

jwe-apple:
> On 7 May 2015, at 16:01, Ricardo Sánchez-Sáez <notifications@github.com> wrote:
> 
> On the first part of the change, I think it's actually the desired behavior. When you enter an answer, say, "2", and go forward, then go back twice, then forward again, you don't want to see the default. You want to see "2", which was the answer you most recently answered for this question. When you skip, you're saying, my answer for this question is "null". That's actually a first class answer (the "null" answer). If you go back twice and then forward again, you still want to see "null" so you don't need to "enter" it again.
> 
> Yeah, that's the behavior after my commit. Before the commit, if go back (without entering any value), and go forward again, the default value was hidden. So I think we're on the same page?
> 
No, I don't think so. We want the default value to be hidden the second time through, if the question had previously been skipped. Which is the original behavior, right?

> On the second part of the change, we want to see the value being hidden just before the step transition. This "glitch" is to communicate to the user that the value has been cleared by taking the "Skip" action. Otherwise it's not immediately obvious that "Skip" clears the answer.
> 
> I beg to disagree: the "glitch" doesn't look nice. ;) It's not something I'd expect from an iOS app. What I'd do is add a quick fade animation for clearing the value, and launch the transition animation after the fade completes. But I guess this would not be a quick thing to add, as there are many different types of question cells involved.
> 
> I can revert this second part if you want (but I'll make sure the thumb is cleared at the same time as the answer in the scale questions).
> 
A fade might be nicer in principle, but we don't want to do it here because it will make the transition too slow overall.

Yes, let's make sure the thumb is cleared at the same time as the answer.

rsanchezsaez:
> > >  On the first part of the change, I think it's actually the desired behavior. When you enter an answer, say, "2", and go forward, then go back twice, then forward again, you don't want to see the default. You want to see "2", which was the answer you most recently answered for this question. When you skip, you're saying, my answer for this question is "null". That's actually a first class answer (the "null" answer). If you go back twice and then forward again, you still want to see "null" so you don't need to "enter" it again.

> >  Yeah, that's the behavior after my commit. Before the commit, if go back (without entering any value), and go forward again, the default value was hidden. So I think we're on the same page?

> No, I don't think so. We want the default value to be hidden the second time through, if the question had previously been skipped. Which is the original behavior, right?

Sorry about the confusion. If the question has been previously skipped, the default value is hidden both with and without the commit. If the question has been previously answered, the user entered value is shown both with and without the commit.

What the commit fixes is this:
1. You view a question with a default value.
2. You do not enter any value, nor press skip.
3. You go back, and go forward again.
4. The default value is not there.

After the commit:
1..3. Same
4. The default value is there

>>> On the second part of the change, we want to see the value being hidden just before the step transition. This "glitch" is to communicate to the user that the value has been cleared by taking the "Skip" action. Otherwise it's not immediately obvious that "Skip" clears the answer.

>>  I beg to disagree: the "glitch" doesn't look nice. ;) It's not something I'd expect from an iOS app. What I'd do is add a quick fade animation for clearing the value, and launch the transition animation after the fade completes. But I guess this would not be a quick thing to add, as there are many different types of question cells involved.

 >> I can revert this second part if you want (but I'll make sure the thumb is cleared at the same time as the answer in the scale questions).

> A fade might be nicer in principle, but we don't want to do it here because it will make the transition too slow overall.

Well, if the fade is really really quick (1/3 or 1/4 of the transition time), I maintain that it's better than the label just disappearing. Or you could launch the fade animation at the same time as the transition, and if its duration is shorter than the transition, the user would notice it disappearing as well. That'd be my suggestion. ;)

> Yes, let's make sure the thumb is cleared at the same time as the answer.

I'l do this then.

rsanchezsaez:
I have replaced the second commit by one updating the scale layout when thumb visibility changes.

See my previous comment regarding the first commit, I still think it's a valid fix.

jwe-apple:
Ok, that's clear.

jwe-apple:
Tested, looks good.

Issue #119, opened 2015-04-28
Can't sign consent in landscape on iPhone 6


getaaron:
If my iPhone 6 is in landscape when the `ORKConsentReviewStep` appears, the view is too short vertically, which causes vertical scrolling behavior. This makes it impossible to sign.

If I rotate to portrait, then back to landscape, the issue persists.

If I rotate to portrait, sign the form, then back to landscape, the issue no longer persists.

rsanchezsaez:
Confirmed. It happens on *stable* as well.

jwe-apple:
Yep, this is known. In the shipping apps landscape is disabled. If you have a need to actually have rotation enabled, happy to accept a PR.

jwe-apple:
<rdar://problem/20245432>

PR #160, opened 2015-05-08
Fix unable to draw signature when scrolling is enabled

rsanchezsaez:
Fixes [issue #119](https://github.com/ResearchKit/ResearchKit/issues/119) by prioritizing the signature recognizer over the scrollview one. You can still scroll by dragging outside the signature area.


jwe-apple:
Yep, looks good.

Issue #148, opened 2015-05-06
[iPad] Two finger tapping task vertical spacing improvement


jwe-apple:
The vertical spacing of the two finger tapping active task UI is too compressed. iPad specific metrics should be introduced! Example:

![img_0007](https://cloud.githubusercontent.com/assets/11575859/7504340/4f3a0ef8-f3fe-11e4-8aa5-ab9696b3973d.PNG)

edelabar:
What is the desired location?  Midway between the current location and total taps counter?

jwe-apple:
Make a first pass; it'll get reviewed next week. I would possibly see if it can be done just by adjusting existing metrics first, rather than making a larger change that might break the iPhone specs.

jwe-apple:
As a guess, probably somewhere above center

edelabar:
Fixed when #164 was merged.

jwe-apple:
ta

PR #164, opened 2015-05-08
Fix for Issue #148 - [iPad] Two finger tapping task vertical spacing improvement 

edelabar:
Implementation of enhancement #148.
* On iPad, added more space between title, progress bar, and tap count.  Refactored code should allow more space to be easily added in-between if desired.
* Added `LAYOUT_DEBUG` colors to aid in layout debugging (per observed convention)

**iPad Screenshot (Portrait)**
ios simulator screen shot may 8 2015 4 26 41 pm

**iPad Screenshot (Landscape)**
ios simulator screen shot may 8 2015 4 26 48 pm

### DIFF
--- a/ResearchKit/ActiveTasks/ORKTappingContentView.m
+++ b/ResearchKit/ActiveTasks/ORKTappingContentView.m
@@ -38,6 +38,9 @@
 #import "ORKHelpers.h"
 
 
+// #define LAYOUT_DEBUG 1
+
+
 @interface ORKTappingContentView ()
 
 @property (nonatomic, strong) ORKSubheadlineLabel *tapCaptionLabel;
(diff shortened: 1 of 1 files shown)


jwe-apple:
Ah, I didn't actually mean to move the bottom buttons, though that might be a reasonable idea (though perhaps not quite as far as this).

When you start tapping the progress bar gets inserted between the tap count and the title. I think the title <-> progress bar <-> tap count spacing also needs to be increased on iPad.

After your next iteration, please also include a screenshot for the rotated (landscape) case.

jwe-apple:
As to the method you used - looks about right!

jwe-apple:
Also please email researchkit@apple.com so we can get your contributor agreement sorted.

edelabar:
Honestly, that makes much more sense.  I'll take another pass.

jwe-apple:
Haha, I try...

edelabar:
I think that actually simplified the fix.  How's that look?

jwe-apple:
Looks good to me. Going to let the build machine check it and then merge.

ok to test

PR #173, opened 2015-05-10
ActiveStep fixes

rsanchezsaez:
- An `ORKCountdownLabel` fix which caused the `ORKActiveStepTimerView` to always show 0 as the starting value.
- A small `ORKActiveStep` subclass code homogenization.
- Fix some erroneous init methods missing the `[super init]` success check.


jwe-apple:
r=me

PR #175, opened 2015-05-11
Extract ORKDefineStringKey() (was 'DEF_KEY()') to helpers

rsanchezsaez:
Mini pull request: cherry-picked this from the `navigationrules` branch to avoid future conflicts. I plan to define new keys in the constraint refactor.

---

By the way the `_UppercaseConstraintKey` static string names in `ORKStepHeaderView.m` and `ORKVerticalContainerView.m` look a little weird. The `_` prefix feels like it should only be used for *ivar* names. Should we come up with a better naming convention for those?

How about this:
`_IllustrationToCaptionBaselineKey` -> `ORKStepHeaderViewIllustrationToCaptionBaselineKey`,
`_TopToIllustrationConstraintKey` -> `ORKVerticalContainerViewTopToIllustrationConstraintKey `,
etc.

### DIFF
--- a/ResearchKit/Common/ORKHelpers.h
+++ b/ResearchKit/Common/ORKHelpers.h
@@ -199,3 +199,5 @@ ORKCGFloatNearlyEqualToFloat(CGFloat f1, CGFloat f2) {
     return (ABS(f1 - f2) <= ORKCGFloatEpsilon);
 }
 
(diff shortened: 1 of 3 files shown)


jwe-apple:
They're static, right? As a general rule, don't want to use a full prefix for statics since they're file local, unless we expect to make them extern in future. Happy to drop the underscore though.

PR #97, opened 2015-04-23
[WIP] Validate ORKFormStep items and ORKOrderedTask for identifier uniqueness

garnett:
Fixes https://github.com/ResearchKit/ResearchKit/issues/85

Issue #85, opened 2015-04-22
Validation checks in ORKOrderedTask and ORKFormStep for identifier's uniqueness


jwe-apple:
<rdar://problem/20430177>

In a simple ordered task, or in a form step, the identifier is used to identify the result of a question or form item. Validation code already exists for steps and tasks; that validation should be extended in the case of `ORKOrderedTask` and `ORKFormStep` to verify that the identifiers of their items are unique.

A unit test to verify the validation behavior would be nice too!

jwe-apple:
Completed

PR #179, opened 2015-05-12
Complete unique identifier validation for task steps and form step items

rsanchezsaez:
Takes over [PR #97](https://github.com/ResearchKit/ResearchKit/pull/97) completing [Issue #85](https://github.com/ResearchKit/ResearchKit/issues/85).

I tried to make the new `ORKTaskTests` unit test file as coherent to the one on `rsanchezsaez-navigationrules` as possible to avoid future merge conflicts.


ResearchKit/Common/ORKTask.h
```diff
@@ -167,6 +167,13 @@ ORK_AVAILABLE_DECL
 - (ORKTaskProgress)progressOfCurrentStep:(ORKStep *)step withResult:(ORKTaskResult *)result;
 
 /**
+ Validates the task parameters.
+ 
+ This method should be used to check that all the task parameters are correct. E.g., `ORKOrderedTask` makes sure that all its step identifiers are unique. It should throw an exception if validation fails.
```
jwe-apple: Let's wrap this text more in line with other files in here.
Also, this is a protocol method, so this documentation needs to describe what an implementer should do, in addition to what the caller should do.

In this case, throw an exception on error, because an invalid task is considered unrecoverable.

That's actually an assumption we should consider changing. A validation method like this should probably return a boolean and an error, rather than throw an exception.

jwe-apple: Of course the same consideration would apply to `ORKStep`, so we could do that on a follow-on issue.

jwe-apple: Also - no abbreviations. Instead of 'e.g', use 'For example' or similar.

ResearchKitTests/ORKTaskTests.m
```diff
@@ -0,0 +1,210 @@
...
+ORKDefineStringKey(ORKTBlankStepIdentifier);
+ORKDefineStringKey(ORKTSevereHeadacheStepIdentifier);
+ORKDefineStringKey(ORKTLightHeadacheStepIdentifier);
+ORKDefineStringKey(ORKTOtherSymptomStepIdentifier);
+ORKDefineStringKey(ORKTEndStepIdentifier);
+
+ORKDefineStringKey(ORKTOrderedTaskIdentifier);
+ORKDefineStringKey(ORKTNavigableOrderedTaskIdentifier);
+
+- (void)getTaskSteps:(out NSArray **)outSteps stepIdentifiers:(out NSArray **)outStepIdentifiers {
```
jwe-apple: call it -generateTaskSteps perhaps?

ResearchKitTests/ORKTaskTests.m
```diff
@@ -0,0 +1,210 @@
...
+@end
+
+
+@implementation ORKTaskTests {
+    NSArray *_orderedTaskStepIdentifiers;
+    NSArray *_orderedTaskSteps;
+    ORKOrderedTask *_orderedTask;
+}
+
+ORKDefineStringKey(ORKTHeadacheChoiceValue);
```
jwe-apple: These should juts be static (I hope!), so they don't need an "ORKT" prefix.

jwe-apple:
Overall, great work, comments are largely cosmetic. Thanks for taking this on!

ResearchKitTests/ORKTaskTests.m
rsanchezsaez: Are you sure? I kind of like the ORKT prefix, I think it makes the static variable usage clearer. I feel weird about declaring statics without any prefix.

ResearchKit/Common/ORKTask.h
rsanchezsaez: Thanks for the documentation comment feedback, I just rewrote it.

However, I feel confortable about throwing an exception from this method instead of returning a boolean/error. What would you do with the returned error other than throwing an exception? Also, `ORKAnswerFormat`'s `validationParameters` does the same (in addition to `ORKStep`'s).

If you still want to do that nevertheless, let me know and I can create a follow-on issue.

ResearchKitTests/ORKTaskTests.m
rsanchezsaez: Done. :)

rsanchezsaez:
Added some more commits and comments. Let me know if you strongly want me to take care of the unaddressed issues. :)

ResearchKitTests/ORKTaskTests.m
jwe-apple: I'm pretty sure. It's helpful to be able to readily identify a symbol with local scope, and the prefix is just noise anyway. In Swift, for instance, all these nasty prefixes go away.

But, it's test code so I don't care that much.

jwe-apple:
No further action required. I'll open an issue to convert to convert the validation exceptions in future.

ResearchKitTests/ORKTaskTests.m
rsanchezsaez: You convinced me: done.

jwe-apple:
Thanks for completing the merge.

PR #180, opened 2015-05-12
Issue #1 - Image Capture

brucehappy:
Here is my first attempt at finalizing the image capture code I used for my mock ups.  As I'm sure you are aware, you will need to test on a real device in order to use the camera.

The step supports different orientations and updates the preview and resultant image orientation accordingly.  In landscape orientation, I have made the continue button view translucent to give the preview more room.  The step supports specifying the template image to show over the preview, as well as insets to use for that image from the preview content frame.  These insets can be points or percentage, as specified with a boolean property.  Navigating back and forward through an image capture task which has already captured an image should work as expected.  The file saved to disk uses complete unless open protection.

I do think there needs to be a little more work on handling the situation when a user denies access to the camera.  I had some code in there to specifically request permission, but took it out for simplicity.  I think it needs to go back in to help handle that situation better.

In my mock ups, I had hacked InstructionStep in order to get the images in the steps leading to the image capture tinted, as well as, changing the default text of the continue button on the first two steps.  I didn't want to include those hacks in the PR, but I do think it looks much better with the tinting enabled, and presently there is no way to enable that within ORKTest.  So, I think we should look to add something to support this.  As for the continue button text in the steps, I do prefer my mocked up version too.  Let me know if you think we should pursue either of these.

I do see some oddness on the ORKStepHeaderView's layout of the instruction label, but since that's only used when there was an error in the image capture, I'm putting it on the back burner. 

Happy to hear any and all suggestions to improve this.


jwe-apple:
You shouldn't need to hack in order to tint the images. I'll take a look tomorrow morning. Do take a well-earned break!

jwe-apple:
BTW the access request for camera should probably go through the same route that other permissions requests go, so they happen in tasks automatically, immediately after the last initial instruction step in the task.

brucehappy:
Given your comments on the tinting and the permissions, I have a feeling you were expecting an OrderedTask, which I didn't do.  I'm also wondering if you were expecting the image capture to be a Recorder subclass, and couple that with an ActiveTask subclass for the capturing, which I also didn't do.  All of this is possible, of course.

I think it makes sense to create an OrderedTask static that creates an activestep subclass (that will just ask for permission for the camera and present the instructions on taking the picture, including a tinted image), and then creates the capture step, and then optionally creates a conclusion step.  My mock up flow, though, has two captures, with an additional instruction step before each capture, and one instruction step at the beginning of the whole flow, and one conclusion step at the end.  If we used this newly proposed OrderedTask static to create the mock up flow in ORKTest, then I'd have to call it twice, one for each hand capture, and then pull the steps out of each created OrderedTask, and then place them all, along with the other instruction and conclusion step, into a new OrderedTask containing the whole flow.  I'd still end up without tinted images in the instruction steps I created directly in ORKTest, though.  How does this sound?

With regard to making image capture follow the Recorder concept, that would likely be a bigger effort of refactoring.  But, there is certainly an argument to be made to do image capture under the Recorder umbrella, as audio is already done there.

jwe-apple:
No, don't think using a recorder should be necessary or desirable. Instead I'd just refactor, moving the `requestedPermissions` property from `ORKActiveStep` to `ORKStep` on the basis that image capture's not really an active step, but does require a permission. The logic for permissions management is in `ORKTaskViewController`, and only assumes that the task implementation will correctly summarize the permissions from its steps.

Also don't think you need to create an ordered task - we want image capture to be mixable with other steps.

brucehappy:
Ok, I can do the refactor of `requestedPermissions` and add a new permission type for camera.

Understood regarding the ordered task.  Again, the only nit with creating the instruction steps outside RK (like they would be in `ORKOrderedTask`) is there is no public API for tinting the images.  So the instruction steps created by ORKTest would not be tinted, and the developer would have to pre-tint them.

brucehappy:
Ok, done with the move of `requestedPermissions`.  I have to say, however, that I really dislike the approach taken when a permission is not granted.  In ORKTest, you cannot advance to the next step, and there is no indication of the reason.  It has the appearance of a hang.

My step can handle not being able to get access to the camera, and puts up an error message, and allows the step to be skipped by pressing Next (after signaling that he step has failed to remove he result).  But, with this new approach of using the `ORKTaskViewController` for permission handling, my step never gets the chance to alert the user to the problem.  How is this handled in the real apps that use ORK?

jwe-apple:
The error is passed back to the task view controller's delegate, and it's the developer's responsibility to decide what to do. They could show an alert, and force forward navigation if they want. We should change ORKTest to present an alert to the user on error.

ResearchKit/Localized/en.lproj/Localizable.strings
```diff
@@ -129,6 +129,14 @@
 "TASK_COMPLETE_TITLE" = "Activity Complete";
 "TASK_COMPLETE_TEXT" = "Your data will be analyzed and you will be notified when your results are ready.";
 
+/* Image catpure step. */
+"CAPTURE_BUTTON_CAPTURE_IMAGE" = "Capture Image";
+"CAPTURE_BUTTON_RECAPTURE_IMAGE" = "Recapture Image";
+"CAPTURE_ERROR_INPUT_NOT_FOUND" = "This device does not contain a camera.  This step cannot be completed.";
+"CAPTURE_ERROR_OUTPUT_NOT_FOUND" = "This device cannot capture an image.  In order to complete this step, allow this app access to the camera in the settings.";
```
jwe-apple: I don't think we want the first sentence here. It's not that the device can't capture, is it? Isn't it only the second part that's valid?

ResearchKit/Localized/en.lproj/Localizable.strings
```diff
@@ -129,6 +129,14 @@
 "TASK_COMPLETE_TITLE" = "Activity Complete";
 "TASK_COMPLETE_TEXT" = "Your data will be analyzed and you will be notified when your results are ready.";
 
+/* Image catpure step. */
+"CAPTURE_BUTTON_CAPTURE_IMAGE" = "Capture Image";
+"CAPTURE_BUTTON_RECAPTURE_IMAGE" = "Recapture Image";
+"CAPTURE_ERROR_INPUT_NOT_FOUND" = "This device does not contain a camera.  This step cannot be completed.";
```
jwe-apple: s/This device does not contain a camera/No camera found/

Testing/ORKTest/ORKTest/MainViewController.m
```diff
@@ -1569,6 +1580,73 @@ - (IBAction)showImageChoices:(id)sender {
...
 }
 
+# pragma mark - Image Capture
+- (id<ORKTask>)makeImageCaptureTask {
+    NSMutableArray *steps = [NSMutableArray new];
+    
+    {
+        ORKInstructionStep *step = [[ORKInstructionStep alloc] initWithIdentifier:@"begin"];
+        step.title = @"Hands";
+        step.image = [UIImage imageNamed:@"hands_solid"];
```
jwe-apple: These images need to be converted to template style images. See the examples in ORKTask.m

ResearchKit/Common/ORKImageCaptureStep.h
```diff
@@ -0,0 +1,42 @@
...
+ OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+
+#import <ResearchKit/ResearchKit.h>
+
+
+ORK_CLASS_AVAILABLE
+@interface ORKImageCaptureStep : ORKStep
```
jwe-apple: These properties and the class as a whole need headerdoc. I can't see a use for non-percentage based insets, so we can drop that property.

ResearchKit/Common/ORKImageCaptureStepViewController.h
```diff
@@ -0,0 +1,38 @@
...
+ OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+
+#import <ResearchKit/ResearchKit_Private.h>
+
+
+ORK_CLASS_AVAILABLE
+@interface ORKImageCaptureStepViewController : ORKStepViewController
```
jwe-apple: Headerdoc required.

ResearchKit/Common/ORKImageCaptureView.h
```diff
@@ -0,0 +1,55 @@
...
+ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+
+#import <UIKit/UIKit.h>
+#import <AVFoundation/AVFoundation.h>
+#import "ORKImageCaptureStep.h"
+
+
+@protocol ORKImageCaptureViewDelegate <NSObject>
```
jwe-apple: Is this header intended to be ASSUME_NONNULL?

ResearchKit/Common/ORKImageCaptureCameraPreviewView.h
```diff
@@ -0,0 +1,46 @@
...
+ OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+
+#import <UIKit/UIKit.h>
+#import <AVFoundation/AVFoundation.h>
+
+
+@interface ORKImageCaptureCameraPreviewView : UIView
```
jwe-apple: If you're marking for nullability, wrap header in NS_ASSUME_...

jwe-apple:
Couple more comments:
1) Please confirm you have rights to the hand images, i.e., did you create them yourself?
2) It would be good to add a Swift example to ORKCatalog too. This would only need to include one instruction step and an image capture step, so the user can see the results in the result tab. We might want to add image viewing to the result part of ORKCatalog.

For 2), would be happy for it to come as a separate issue / PR, in which case, just open the issue.

ResearchKit/Localized/en.lproj/Localizable.strings
brucehappy: I only use CAPTURE_ERROR_OUTPUT_NOT_FOUND when this if statement returns false:
```
    AVCaptureStillImageOutput *stillImageOutput = [[AVCaptureStillImageOutput alloc] init];
    if ([_captureSession canAddOutput:stillImageOutput])
```
but the only reason that canAddOutput would be returning false is indeed likely to be permissions being denied.  I'll remove the first sentence.

ResearchKit/Localized/en.lproj/Localizable.strings
brucehappy: Ok.

Testing/ORKTest/ORKTest/MainViewController.m
brucehappy: You must mean a different source file as that one doesnt exist.

ResearchKit/Common/ORKImageCaptureStep.h
brucehappy: Well, I had thought in future we may want an option to lock the orientation for the step, giving them better control over how the template image is presented, and since the template image is scaled to fit, as is the camera preview, someone might want to just say 8pt all around the template image.  It's perhaps a non-typical case.  I'll remove, and if point based insets are needed in future, we can certainly re-introduce the property.

brucehappy:
1. Yes, I created them myself using the tapping hand images in ORK as a base.
2. Sure, I can add a swift example to ORKCatalog.  I'll also take a look at adding image viewing to the result portion of the app.

Testing/ORKTest/ORKTest/MainViewController.m
brucehappy: If you mean do something like this:
```
        step.image = [[UIImage imageNamed:@"hands_solid" inBundle:nil compatibleWithTraitCollection:nil] imageWithRenderingMode:UIImageRenderingModeAlwaysTemplate];
```
If so, this will unfortunately not make the hands show up as blue in the instruction step since the `shouldTintImages` in the `ORKInstructionStep` will be false by default, and so the `ORKTintedImageView` will recreate the image with `UIImageRenderingModeAutomatic` before showing it.  This is the issue I was referring to earlier regarding hacking `ORKInstructionStep` so I could get at this property.

brucehappy:
All comments should now be addressed with the exception of the non-tinted images on the instruction steps created in `ORKTest` and `ORKCatalog` for image capture tasks.  See my comments on this topic to your line note.

`ORKCatalog` was updated to include an image capture step and to show the image in the result tab.  Please let me know if you'd rather I removed the hand images used for the instruction step and image capture step in this example.

Testing/ORKTest/ORKTest/MainViewController.m
rsanchezsaez: Shouldn't we just make `ORKStep`'s `shouldTintImages` property *public* instead of *private*? That way we could easily tint those (and any other) instruction step images from outside the framework.

brucehappy: Well, thats what I had done in the version I didn't check in.  I presumed there was a good reason why it wasn't made public.  Would like to hear what @jwe-apple has to say on the topic.

jwe-apple: shouldTintImages seemed like such an implementation detail of the framework that it seemed undesirable to expose it on ORKStep for general use, especially since in most apps you will already know what your tint color is.

My feeling is that ORKTintedImageView's shouldApplyTint property should not *override* the rendering mode of the input image to turn off tinting if it is already in templated rendering mode, which can only happen if someone has explicitly gone to the trouble of setting it. There are some funny details in ORKTintedImageView e.g. self.image = image to re-apply the tint, which would need to be resolved when changing this behavior.

jwe-apple:
@brucehappy I'm about to add a block of comments now from another reviewer here. Apologies I didn't get these to you sooner. Both comments and nitpicks, so feel free to push back if you feel strongly.

ResearchKit/Common/ORKImageCaptureCameraPreviewView.m
```diff
@@ -0,0 +1,187 @@
...
+    _templateImageView.image = templateImage;
+}
+
+- (void)setHideTemplateImage:(BOOL)hideTemplateImage {
+    _templateImageView.alpha = hideTemplateImage ? 0 : 1;
+}
+
+- (BOOL)hideTemplateImage {
+    return _templateImageView.alpha==0;
+}
```
jwe-apple: For the above pair of methods:

Can the setter be renamed to setTemplateImageHidden:(BOOL)hidden
The getter would be better if renamed to isTemplateImageHidden

ResearchKit/Common/ORKImageCaptureStepViewController.m
```diff
@@ -0,0 +1,276 @@
...
+}
+
+- (void)capturePressed:(void (^ __nullable)(BOOL))handler {
+    // Capture the image via the output
+    dispatch_async(self.sessionQueue, ^{
+    	[[self stillImageOutput] captureStillImageAsynchronouslyFromConnection:[[self stillImageOutput] connectionWithMediaType:AVMediaTypeVideo] completionHandler:^(CMSampleBufferRef imageDataSampleBuffer, NSError *error) {
+            [self queue_captureImageFromData:imageDataSampleBuffer withHandler:handler];
+    	}];
+    });
+}
```
jwe-apple: Will the behavior be safe if the capture button or recapture button is pressed twice in quick succession? Looks like startRunning and stopRunning might get called multiple times in different orders. If the session is not running and captureStillImageAsynchronouslyFromConnection: gets called, will it be okay?

ResearchKit/Common/ORKImageCaptureView.m
```diff
@@ -0,0 +1,257 @@
...
+    [[NSNotificationCenter defaultCenter] removeObserver:self];
+}
+
+-(void)queue_sessionRunning {
+    dispatch_async(dispatch_get_main_queue(), ^{
+        _previewView.hideTemplateImage = NO;
+    });
+}
+
+- (void)orientationChange {
```
jwe-apple: How about orientationChanged or orientationDidChange?

ResearchKit/Common/ORKImageCaptureView.m
```diff
@@ -0,0 +1,257 @@
...
+        [self.mconstraints addObjectsFromArray:[NSLayoutConstraint constraintsWithVisualFormat:@"H:|[_continueSkipContainer]|" options:NSLayoutFormatDirectionLeadingToTrailing metrics:nil views:dictionary]];
+        [self.mconstraints addObjectsFromArray:[NSLayoutConstraint constraintsWithVisualFormat:@"V:|[_headerView]-[_continueSkipContainer]|" options:NSLayoutFormatDirectionLeadingToTrailing metrics:nil views:dictionary]];
+    } else {
+        // If we do not have an error to show, layout the previewView and continueSkipContainer
+        [self.mconstraints addObjectsFromArray:[NSLayoutConstraint constraintsWithVisualFormat:@"H:|[_previewView]|" options:NSLayoutFormatDirectionLeadingToTrailing metrics:nil views:dictionary]];
+        [self.mconstraints addObjectsFromArray:[NSLayoutConstraint constraintsWithVisualFormat:@"H:|[_continueSkipContainer]|" options:NSLayoutFormatDirectionLeadingToTrailing metrics:nil views:dictionary]];
+    
+        // Float the contine view over the previewView if in landscape to give more room for the preview
```
jwe-apple: Typo: 'continue' view is missing the u in the comment.

ResearchKit/Common/ORKImageCaptureView.m
```diff
@@ -0,0 +1,257 @@
...
+            _continueSkipContainer.backgroundColor = [_continueSkipContainer.backgroundColor colorWithAlphaComponent:CONTINUE_ALPHA_OPAQUE];
+        }
+    }
+    
+    [self addConstraints:self.mconstraints];
+    
+    [super updateConstraints];
+}
+
+- (BOOL)floatContinueView {
```
jwe-apple: Can this be renamed to isApplicationInLandscapeOrientation or something? Or have it inline, since it’s just one line?

ResearchKit/Localized/en.lproj/Localizable.strings
```diff
@@ -129,6 +129,14 @@
 "TASK_COMPLETE_TITLE" = "Activity Complete";
 "TASK_COMPLETE_TEXT" = "Your data will be analyzed and you will be notified when your results are ready.";
 
+/* Image catpure step. */
```
jwe-apple: s/catpure/capture/

ResearchKit/Localized/en.lproj/Localizable.strings
```diff
@@ -129,6 +129,14 @@
...
+"CAPTURE_BUTTON_CAPTURE_IMAGE" = "Capture Image";
+"CAPTURE_BUTTON_RECAPTURE_IMAGE" = "Recapture Image";
+"CAPTURE_ERROR_CAMERA_NOT_FOUND" = "No camera found.  This step cannot be completed.";
+"CAPTURE_ERROR_NO_PERMISSIONS" = "In order to complete this step, allow this app access to the camera in the settings.";
```
jwe-apple: s/the settings/Settings/

ResearchKit/Common/ORKImageCaptureCameraPreviewView.m
brucehappy: Done

ResearchKit/Common/ORKImageCaptureView.m
brucehappy: Done

ResearchKit/Common/ORKImageCaptureView.m
brucehappy: Done

ResearchKit/Common/ORKImageCaptureView.m
brucehappy: Inlined.

ResearchKit/Localized/en.lproj/Localizable.strings
brucehappy: Done

ResearchKit/Localized/en.lproj/Localizable.strings
brucehappy: Done

jwe-apple:
ok to test

ResearchKit/Common/ORKImageCaptureStepViewController.m
brucehappy: Excellent point.  Regardless of whether it's okay if this is called multiple times in quick succession, it was not what I was intending to be possible, and so I've modified the code to prevent it from occurring.  I made the change in the `ORKImageCaptureView`, however, as that is where the button actions are targeted.  I took the same approach as used in `ORKNavigationContainerView` where a BOOL is used in a similar situation.

brucehappy:
I believe I have now addressed all the comments.

brucehappy:
Oops!  Forgot the tinting!

jwe-apple:
Build failures were a setup issue - all is well

Testing/ORKTest/ORKTest/MainViewController.m
brucehappy: So then, you are saying that if an image was given to the `ORKTintedImageView` that had a `renderingMode` of `UIImageRenderingModeAutomatic`, then whether it is shown as a template image or not should be governed by the `shouldApplyTint` property of `ORKTintedImageView`.  If the `renderingMode` of the image given to the `ORKTintedImageView` is either `UIImageRenderingModeAlwaysTemplate` or `UIImageRenderingModeAlwaysOriginal`, then you do not want `ORKTintedImageView` to change it, no matter what the `shouldApplyTint` property is set to.  Is that right?

jwe-apple: Yes. I hadn't thought about `UIImageRenderingModeAlwaysOriginal`, but this way of dealing with it seems consistent.

brucehappy:
Ok, I've addressed the tint as well.  Please let me know whether this approach is appropriate.

jwe-apple:
Approach looks good!

brucehappy:
Spacing fixed.

rsanchezsaez:
I think that `ORKTest`'s `ORKJSONSerializationTests` tests fail after this merge.

brucehappy:
@rsanchezsaez It looks like the merge had the right changes to ORKTest in it:
https://github.com/ResearchKit/ResearchKit/commit/17679ff56e5e0aa338e2399fd47e700153c1e0bb

jwe-apple:
I've actually got a fix for these test failures in another branch, coming today.

brucehappy:
@rsanchezsaez Oh, you meant in `testORKSecureCoding`, yes, correct, I missed that.  And it looks like @jwe-apple has addressed it in #214. Thanks @jwe-apple!

PR #182, opened 2015-05-13
Visual consent view controller cleanup and optimization

### LINKED ISSUE #177: [iPad] Poor performance during consent transitions
The consent animated transitions are a bit slow (observed frame drops on an iPad Air). Need to profile and find out why. If we need to disable the tinting of these animations on iPad, then we should do so and document it.

rsanchezsaez:
Im my investigation on [Issue #177](https://github.com/ResearchKit/ResearchKit/issues/177), I have cleaned `ORKVisualConsentStepViewController` and `ORKVisualConsentTransitionAnimation` up.

The visual consent step view controller now features a much simpler animator logic (it should be easier to read and maintain).

I performed extensive testing and all seems to work fine. I made sure there are no retain cycles (the animator and view controller are being deallocated when appropriate). I also made sure all transition-related UI calls are being performed on the main thread.

This PR does not fix #177 yet. Rather, it's a prequel for an eventual fix.


jwe-apple:
Generally an improvement.

Been a while since I looked at this code, so a couple things I'm not sure about:

1- Why do you think the dealloc method on the animator is useless? You might be right but I don't follow. Does the animator retain itself?

2- One thing the semaphore based code did do is timeout if the animator didn't complete, and I don't see that carried through. It is possible for other parts of the system to misbehave so that the animator never completes - we saw this in testing.

rsanchezsaez:
1) The animator owns the *display link* and the *display link* retains the animator (its target). So the animator can only deallocated after the display link is invalidated, which happens in the `finish` method. The `dealloc` method was this:

    - (void)dealloc {		
        [self finish];		
        [_displayLink invalidate];		
    }

so if you are inside the deallocator, you know fore sure that `[self finish]` has already been called (and `_displayLink invalidate` is redundant because `finish` already does that).

2) I'll look into that.

jwe-apple:
1) Yep

rsanchezsaez:
Regarding [Issue #177](https://github.com/ResearchKit/ResearchKit/issues/177), I've been trying hard to optimize this to no avail.

The problem seems to be that calling `[pageViewController setViewControllers:@[viewController] direction:direction animated:animated completion:...]` on the main thread interferes with the video playback.

The thing is that once you play the (jerky) transition animation once, going back to the previous step and going forward again results in a smooth animation. Because of this, my initial guess was that there should be a way to preload the video so it's smooth, but after much tests I now have doubts.

I tried reusing the `ORKVisualConsentTransitionAnimator` (and most of its owned objects) and calling `[moviePlayer seekToTime:toleranceBefore:toleranceAfter:` well in advance of pressing next, but that didn't help.

On the other hand, replacing the the `ORKEAGLMoviePlayerView` by a simple un-tinted `AVPlayer`-backed view did indeed improved performance. So maybe this is the way to go after all. We could have pre-tinted videos for iPad. But it's a shame that we cannot universally use the clever dynamic video tinting approach.

 Thoughts? Anything else I should try?

jwe-apple:
Ok. Could you run the GPU profiler and see if anything obvious pops up? If there's nothing obvious to you, we'll merge your cleanup after you've addressed item 2 (thanks), and then you can leave this one with me.

rsanchezsaez:
I dealt with 2) by re-adding the semaphore. Now the semaphore is contained within the `doAnimateFromViewController:toController:direction:url:animateBeforeTransition:transitionBeforeAnimate:completion:` method, so the logic should be easier to follow. 

---

I did some minor OpenGL cleanup and optimizations in `ORKEAGLMoviePlayerView`. These are not too relevant as OpenGL was not the bottleneck, but a few less redundant OpenGL calls are performed now.

---

Finally, I think I fixed [Issue #177](https://github.com/ResearchKit/ResearchKit/issues/177). By profiling I realized that the CPU bottleneck was due to decoding the huge consent image of the next view controller (`[_CSIRenditionBlockData expandCSIBitmapData:fromSlice:makeReadOnly:]` method) at the same time as the video was playing. So now I'm using `ORKImageByTintingImage()` instead of `imageWithRenderingMode:UIImageRenderingModeAlwaysTemplate` to render the visual consent images. This function seems much more performant, and the transitions look completely smooth on an *iPad 4th gen*.

---

I also merged the latest *master* head and fixed some leftover merge conflicts from the *tone audiometry* pull request.

jwe-apple:
Overall looks good, just questions about the timeout.

jwe-apple:
r=me

PR #184, opened 2015-05-14
Allow users to add a dot to the signature view

scelis:
Add a small circle to the current signature path in touchesBegan to
properly support dots in the signature view. This provides a nicer
experience to users, allowing them to dot their i's.


jwe-apple:
Nice. r=me

PR #185, opened 2015-05-14
Fix various multitouch issues in ORKSignatureView

scelis:
This commit fixes a couple of issues with extra touches in the signature
view:

* Adding a second touch while signing would cause the path to jump to
  the second touch, which is probably not what the user wanted.
* Once two touches were applied to the signature view, if both touches
  ended simultaneously, gestureTouchesEnded: would return before calling
  signatureViewDidEditImage: on the delegate. This could get the UI into
  an odd state where there is a signature in the view but no clear
  button below the view.

By ignoring the extra touches completely, we can also clean up the code
a little and remove the extra touches checks in the various touch event
handlers.


jwe-apple:
ok to test

Looks good but I ought to test this one.

jwe-apple:
r=me

PR #189, opened 2015-05-15
ORKTest: Fix remaining scale merge conflict

rsanchezsaez:
I missed one additional merge conflict which caused `ORKTest` not to compiler. Sorry.


jwe-apple:
thanks

PR #188, opened 2015-05-15
Fix double bottom separator in boolean and text choice questions

rsanchezsaez:
Fixes [Issue #183](https://github.com/ResearchKit/ResearchKit/issues/183).

Needs [pull request #187](https://github.com/ResearchKit/ResearchKit/pull/187) as a pre-requisite.

PR #191, opened 2015-05-16
Avoid using magic number in ORKQuestionStepViewController

rsanchezsaez:
Makes [pull request #188](https://github.com/ResearchKit/ResearchKit/pull/188)'s fix slightly prettier.

### DIFF
--- a/ResearchKit/Common/ORKQuestionStepViewController.m
+++ b/ResearchKit/Common/ORKQuestionStepViewController.m
@@ -592,7 +592,8 @@ typedef NS_ENUM(NSInteger, ORKQuestionSection) {
 - (void)tableView:(UITableView *)tableView willDisplayCell:(UITableViewCell *)cell forRowAtIndexPath:(NSIndexPath *)indexPath {
     cell.layoutMargins = UIEdgeInsetsZero;
     if (indexPath.section == ORKQuestionSectionSpace2) {
-        cell.separatorInset = (UIEdgeInsets){.left = 10000.0};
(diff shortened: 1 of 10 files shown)


ResearchKit/Common/ORKQuestionStepViewController.m
```diff
@@ -592,7 +592,8 @@ - (UITableViewCell *)tableView:(UITableView *)tableView cellForRowAtIndexPath:(N
 - (void)tableView:(UITableView *)tableView willDisplayCell:(UITableViewCell *)cell forRowAtIndexPath:(NSIndexPath *)indexPath {
     cell.layoutMargins = UIEdgeInsetsZero;
     if (indexPath.section == ORKQuestionSectionSpace2) {
-        cell.separatorInset = (UIEdgeInsets){.left = 10000.0};
+        // Hide double bottom separator (the last answer cell already has one)
+        cell.separatorInset = (UIEdgeInsets){.left = cell.bounds.size.width};
```
jwe-apple: I haven't checked, but is this definitely called on all cells after a rotation?

rsanchezsaez:
Oops, you are right, `tableView:(UITableView *)tableView willDisplayCell:forRowAtIndexPath:` is not called after rotation.

I went back to the magic number, but extracted it as the `ORKScreenMetricMaxDimension` constant in `ORKSkin` (and reused it everywhere else).

jwe-apple:
Yep, thanks.

PR #197, opened 2015-05-18
_signaturePageContent not copied (fix for issue 196)

stownsend:
Changes to ConsentDocument.m to fix bug outlined in issue 196, _signaturePageContent not being copied when the consent document is copied.


jwe-apple:
Yep, r=me. Thanks

PR #198, opened 2015-05-18
Tests for ORKConsentDocument

alexbasson:
This PR adds test coverage for `ORKConsentDocument`. To do so required a multi-stage process, as reflected in the commits:

1) Add an initializer to `ORKConsentDocument` to allow for dependency injection, as it depends on `ORKPDFHTMLWriter`. `-init` now calls through to this initializer, injecting default objects for the dependencies.
2) Write tests to cover `-makePDFWithCompletionHandler`. In the process of writing these tests, it became apparent that it would be beneficial to use formatter objects for `ORKConsentSection` and `ORKConsentSignature`, so those formatter classes were extracted.
3) Write tests to cover the formatters mentioned above.

As this process added three tests classes to the cover consent behavior, a Consent group was added under the ResearchKitTests group to hold consent-related test classes.

### DIFF
--- a/ResearchKit/Consent/ORKConsentSignatureFormatter.h
+++ b/ResearchKit/Consent/ORKConsentSignatureFormatter.h
@@ -1,0 +1,8 @@
+#import <ResearchKit/ResearchKit.h>
+#import <ResearchKit/ORKConsentSignature.h>
+
+@interface ORKConsentSignatureFormatter : NSObject
+
+- (NSString *)HTMLForSignature:(ORKConsentSignature *)signature;
+
+@end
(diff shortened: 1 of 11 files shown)


jwe-apple:
ok to test

ResearchKit/Consent/ORKConsentSignatureFormatter.h
```diff
@@ -0,0 +1,8 @@
+#import <ResearchKit/ResearchKit.h>
+#import <ResearchKit/ORKConsentSignature.h>
+
+@interface ORKConsentSignatureFormatter : NSObject
+
+- (NSString *)HTMLForSignature:(ORKConsentSignature *)signature;
```
jwe-apple: A little unfortunate to have to allocate an object just to call this stateless method. That said, it does make the test code considerably cleaner than the alternatives I've thought of.

jwe-apple:
Tests pass, looks good.

ResearchKit/Consent/ORKConsentSignatureFormatter.h
rsanchezsaez: Why not make the method a *class method* or the class a *singleton class*?

alexbasson: I made them instance methods for two reasons:

1) Habit. At Pivotal, we use [Cedar](https://github.com/pivotal/cedar) for testing Objective-C code, and while Cedar has very rich support for mocks out of the box, it doesn't support mocking class methods. Of course here, I'm hand-rolling the mocks anyway, so this wouldn't be an issue.

2) It more closely follows the pattern established by Foundation formatter classes such as `NSDateFormatter` and `NSNumberFormatter`. Those classes feature state while the formatters I've introduced do not, but I thought that in terms of maintaining a consistent API, it was worth making them instance methods.

rsanchezsaez: *Number 2* makes a lot of sense. Thanks for the explanation.

jwe-apple: To add to that - in general, using separate, composable objects for this makes things more testable since you can swap or nil out parts of the implementation easily at runtime rather than having to subclass the class under test multiple times. If you used a class method you'd still need a way to override what class is used, which would require subclassing to override (like, for example, the `layerClass` class method on UIView).

rsanchezsaez: Thanks.

PR #201, opened 2015-05-19
Fix identifier uniqueness issues

jwe-apple:
1) Form items can have no identifier if they are just placeholders
for section headers in the table view. We should ignore the nil
identifiers for the purpose of the uniqueness check.

2) In ORKTest, the scale questions task had a non-unique step
identifier in it.


rsanchezsaez:
Code reviewed and tested. Form item identifier validation correctly ignores `nil` values.

jwe-apple:
Thanks!

Issue #174, opened 2015-05-11
Create coding style document


rsanchezsaez:
I took the liberty to draft a [ResearchKit Coding Style](https://github.com/rsanchezsaez/ResearchKit/wiki/ResearchKit-Coding-Style) document.

It mainly deals with visual formatting (whitespace, newlines, alignment, etc.). I think it can be useful to maintain code homogenization.

What do you think?

brucehappy:
Yes, please.  I end up looking in other files, and find inconsistencies, so having a master document is wonderful.

jwe-apple:
Seems useful. Do get a PR together, and please send an email to researchkit-dev as well to invite comment.

I'll want to tone down things like requiring alphabetical order and exact numbers of lines of whitespace between classes etc. The threshold for this is whether you'd feel it was too nitpicky when commenting on a PR for a new feature. But otherwise this seems like a great contribution - thanks for the effort!

rsanchezsaez:
I'm afraid GitHub doesn't support PR of wiki pages. Either you'll have to copy and paste it into this project's wiki, or I can submit it as a regular markdown file inside the repo.
---
I've removed the alphabetical order requirement, that was silly.

However, I think it's useful to enforce a consistent number of empty newlines among the fules, the rules are not that hard to follow I'd say. Otherwise you end with arbitrary spacing in different files.
---
I'll send a mail to the list just now.

rsanchezsaez:
@jwe-apple: By the way, I noticed that I missed a few files in my previous whitespace standardization PR (sorry about that). They're only the recorder subclasses. Could I possibly do these as well? Thanks.

p2:
Good to have indeed! I can live with the spaces-for-indentation, one question about the _opening bracket_:

Agreed that it should go on the same line, but maybe **except** for top level method declarations? I'm used to seeing those after a newline (in other 3rd party libraries as well as Apple code), improves readability quite a bit. So the example would be:

```
// DO
- (void)doSomethingWithString:(NSString *)string
{
    if (condition) {
        ...
    } else {
        ...
    }
}
```

rsanchezsaez:
@p2: Thanks for your comment.

Regarding braces, I could go with either style.

The problem is that all of the *ResearchKit core* codebase was already in the *opening brace on the same line* format, even for method definitions. So, adopting this would require an extensive reformatting (simple to do in an automated way, but potentially introducing *git-blame* noise).

[...]

PR #204, opened 2015-05-19
Add coding-sytle-guide.md document

rsanchezsaez:
I moved the *Coding Style Guide* from my fork's wiki to a standalone document and linked it from *CONTRIBUTING.md* ([Issue #174](https://github.com/ResearchKit/ResearchKit/issues/174)).

Creating a pull request to try and spur more feedback.


jwe-apple:
Thanks, I'd forgotten about this.

jwe-apple:
@YuanZhu-apple I think we want this content, but probably on the wiki?

rsanchezsaez:
I though so too at first. The problem is that *GitHub wikis* [don't support Pull Requests](http://stackoverflow.com/a/11481887/269753), so third party collaboration to those documents gets cumbersome.

jwe-apple:
We expect this to be pretty much write-once (and thank you for writing it).

docs-standalone/coding-style-guide.md
```diff
@@ -0,0 +1,276 @@
...
+
+Declare one variable per line even if they have the same type. In general it's a good idea to initialize variables with a reasonable value.
+
+    // DO
+    int variable1 = -1;
+    int variable2 = -1;
+    int *variablePointer = NULL;
+    id object = nil;
+    NSString *string = nil;
+
```
YuanZhu-apple: Using ARC, strong, weak, and autoreleasing stack variables are now implicitly initialized with nil.

https://developer.apple.com/library/ios/releasenotes/ObjectiveC/RN-TransitioningToARC/Introduction/Introduction.html#//apple_ref/doc/uid/TP40011226-CH1-SW5

Can we drop this nil assignment requirement? Since it is not required.

rsanchezsaez: What's the case with variables of primitive types? Should we mention those instead?

YuanZhu-apple: Correct me if I'm wrong,  primitive types need to be set to NULL, because they are not managed by reference counter and ARC.

rsanchezsaez: I think you are right (although not all of them should be initialized to `NULL`: `BOOL`s, `int`s, `float`s, `double`s, etc. could benefit from being initialized to a reasonable value of the appropriate type). 

Anyway, are we sure we want to change this then? I personally like initializing all stack variables, including ARC-managed object pointers, for visual homogeneity.

YuanZhu-apple: Since in those cases initializing with nil is not required, I imagine some contributors would argue about it when we enforce this guideline. For something is hard to enforce, I tend to remove.

rsanchezsaez: Ok, I added a note about this with the link to the documentation. I now suggest optionally initializing them for visual homogeneity.

If you feel strongly about removing the suggested explicit initialization, tell me and I'll delete the note.

YuanZhu-apple: I think this is fine.

docs-standalone/coding-style-guide.md
```diff
@@ -0,0 +1,285 @@
...
+Do not use two or more empty lines in any other cases.
+
+Use exactly one empty line to separate:
+- The last *forward declaration* and the *`@interface` declaration*.
+- The *`@interface` declaration* and the first *method* or *property declaration*.
+- The last *method* or *property declaration* and the *`@end` keyword*.
+- The *`@implementation` line* and the first *method definition*.
+- The closing bracket of the last *method definition* and the *`@end` keyword*.
+- The *`@param` section* and the *`@return` line* in a *[HeaderDoc](https://developer.apple.com/library/mac/documentation/DeveloperTools/Conceptual/HeaderDoc/intro/intro.html) comment*.
+
```
YuanZhu-apple: Worth to mention that ResearchKit is using AppleDoc instead of HeaderDoc to generate header document. Due to AppleDoc has better support on modern Objective-C and better html styling. 

Also HeaderDoc's comment style is slightly different from AppleDoc's.

rsanchezsaez: Got it. I'll add that over the weekend.

rsanchezsaez:
I've added an *appledoc* section to the guide, and have replaced all mentions of *HeaderDoc* by *appledoc*.

docs-standalone/coding-style-guide.md
```diff
@@ -0,0 +1,299 @@
...
+You can optionally use one (and only one) blank like to separate:
+- Groups of related `#import` statements.
+- Groups of related statements in a single method implementation.
+
+
+#### 1.6. Header File Example
+
+    /*
+     Copyright (c) 2015, Apple Inc. All rights reserved.
+     ...
```
YuanZhu-apple: How about replacing `Apple Inc.` with `AUTHOR'S NAME` to remind developer to put his/her name in the copyright line when creating a new file.

rsanchezsaez: Good idea. Instead, I replaced it with `John Appleseed` which should also be self explanatory. ;)

PR #214, opened 2015-05-21
UI review changeset

jwe-apple:
- Add comments to remind developers to be careful what images they use for image capture steps.
- Update wording for both active tasks
- Add an extra step before countdown for the audiometry active task and be more explicit about headphones
- Adjust color, size, and alignment of supplementary scale labels
- Turn off audio tone generation when inactive


jwe-apple:
This has had internal review.

PR #216, opened 2015-05-21
ORKEAGLMoviePlayerView: fix incorrect aspect ratio after rotation

rsanchezsaez:
Fixes consent animation having an incorrect aspect ratio after a rotation.

Also:
- Add guards so GL functions aren't called if OpenGL is not setup.
- Delete program and buffers on dealloc.


jwe-apple:
Looks good to me.

PR #217, opened 2015-05-22
Add optional tinted image cache (only enabled for consent step view controller)

rsanchezsaez:
Deals with [Issue #177](https://github.com/ResearchKit/ResearchKit/issues/177).

When cache is enabled, all tinted images are created using `ORKImageByTintingImage()`, which forces tinted image decoding. It clears the cache on consent step view controller dealloc.


jwe-apple:
Did you see a measurable performance difference from doing this?

ResearchKit/Common/ORKTintedImageView.m
```diff
@@ -77,17 +158,25 @@ - (UIImage *)imageByTintingImage:(UIImage *)image {
...
+            } else {
+                _tintedImage = [image imageWithRenderingMode:UIImageRenderingModeAlwaysTemplate];
+            }
         } else {
             // Manually apply the tint for animated images (template rendering mode doesn't work: <rdar://problem/19792197>)
             NSMutableArray *images = [NSMutableArray array];
             for (UIImage *image in image.images) {
-                [images addObject:ORKImageByTintingImage(image, tintColor, screenScale)];
+                if (_enableTintedImageCaching) {
+                    [[ORKTintedImageCache sharedCache] tintedImageForImage:image tintColor:tintColor scale:screenScale];
```
brucehappy: You are not adding the image you are pulling from the cache into the images array.

rsanchezsaez: Thanks for the comment. Actually, `tintedImageForImage:tintColor:scale:` does that automatically. :)

rsanchezsaez:
Yeah, I'm pretty sure it makes consent transitions noticeably smoother on an *iPad 4th gen*.

ResearchKit/Common/ORKTintedImageView.m
brucehappy: It automatically adds to the local variable called `images` declared on line 173?

rsanchezsaez: No, it adds each tinted frame image from the animated image. If you build the tinted animated image a second time, it should retrieve the individual frames from the cache.

But yeah, the fact that loop variable has the same name as the argument it's confusing, I'm going to rename it.

rsanchezsaez: Ah, damn, you are right. I was blinded by the code. Thanks for spotting the mistake!

ResearchKit/Common/ORKTintedImageView.m
```diff
@@ -57,6 +65,80 @@ static inline BOOL ORKIsImageAnimated(UIImage *image) {
...
+    if (self) {
+        _image = image;
+        _tintColor = tintColor;
+        _scale = scale;
+    }
+    return self;
+}
+
+- (BOOL)isEqual:(id)object {
+    if ([self class] != [object class]) {
```
jwe-apple: This prevents subclassing, but ok.

rsanchezsaez: Ah. I modeled it after `ORKStep`'s `isEqual:`.

ResearchKit/Common/ORKTintedImageView_Internal.h
```diff
@@ -0,0 +1,39 @@
...
+ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+
+#import <UIKit/UIKit.h>
+
+
+@interface ORKTintedImageCache : NSCache
+
++ (instancetype)sharedCache;
```
jwe-apple: Since this is living in someone else's app, if we're going to have a long-lived singleton, we need to ensure it handles memory warnings.

rsanchezsaez: Doesn't `NSCache` empty the cache on memory warnings automatically? Or do you mean something else?

ResearchKit/Consent/ORKVisualConsentStepViewController.m
```diff
@@ -506,10 +512,24 @@ - (void)observedScrollViewDidScroll:(UIScrollView *)scrollView {
...
+}
+
+- (void)showViewController:(ORKConsentSceneViewController *)viewController forward:(BOOL)forward animated:(BOOL)animated preloadNextViewController:(BOOL)preloadNextViewController {
     [self showViewController:viewController
                      forward:forward
                     animated:animated
-                  completion:nil];
+                  completion:^(BOOL finished) {
+                      if (preloadNextViewController) {
+                          ORKConsentSceneViewController *nextViewController = [self viewControllerForIndex:[self currentIndex]+1];
```
jwe-apple: This code is pretty ugly. We should not have to create an entire view controller in order to add an image to the tint cache, especially an image that just comes from a model object that we already have at this level.

ResearchKit/Common/ORKTintedImageView.m
```diff
@@ -77,21 +159,33 @@ - (UIImage *)imageByTintingImage:(UIImage *)image {
...
     UIColor *tintColor = self.tintColor;
     CGFloat screenScale = self.window.screen.scale; // Use screen.scale; self.contentScaleFactor remains 1.0 until later
-    if ((![_appliedTintColor isEqual:tintColor] || !ORKCGFloatNearlyEqualToFloat(_appliedScaleFactor, screenScale))) {
+    if (screenScale > 0 && (![_appliedTintColor isEqual:tintColor] || !ORKCGFloatNearlyEqualToFloat(_appliedScaleFactor, screenScale))) {
         _appliedTintColor = tintColor;
         _appliedScaleFactor = screenScale;
         
         if (!ORKIsImageAnimated(image)) {
-            _tintedImage = [image imageWithRenderingMode:UIImageRenderingModeAlwaysTemplate];
+            if (_enableTintedImageCaching) {
```
jwe-apple: Feels like this conditional behavior belongs in a method..

ResearchKit/Consent/ORKVisualConsentStepViewController.m
rsanchezsaez: I did this thinking on a compact way of handling both the default active consent images, and any user-supplied custom ones. *GlucoSuccess*, for example, uses custom images together with its own custom animations.

ResearchKit/Common/ORKTintedImageView_Internal.h
jwe-apple: Oops! Yes, this is fine.

ResearchKit/Common/ORKTintedImageView.m
jwe-apple: Never mind, on second reading it's fine. Not having a good morning!

ResearchKit/Consent/ORKVisualConsentStepViewController.m
jwe-apple: ORKConsentSceneViewController has this code when setting the section:

```
    UIImage *image = nil;
    if (consentSection.type == ORKConsentSectionTypeCustom) {
        image = consentSection.customImage;
    } else {
        image = ORKImageForConsentSectionType(consentSection.type);
    }
```

Perhaps we could put this in a private method on ORKConsentSection.

rsanchezsaez: Ok, I'll do this tomorrow.

rsanchezsaez: Done. Sorry for the delay.

rsanchezsaez:
I'm now preloading the next consent section image only.

This is how the CPU profiling looks when running on iPad 4th gen.:

*Before implementing the cache*

screen shot 2015-05-25 at 19 30 22

*With `ORKTintedImageCache`*

screen shot 2015-05-25 at 19 31 23

The last three CPU *mountains* are one visual transition each. With the cache, much of the spike is moved after the transition finishes.

jwe-apple:
Looks good, r=me

PR #218, opened 2015-05-22
ORKAnswerFormat: add scale convenience methods

rsanchezsaez:
Adds some scale answer format convenience methods.


jwe-apple:
Rather not continue increasing the number of methods here. Why don't we just add the extra parameter to the old "maxi" factory methods?

rsanchezsaez:
The factory methods already have the extra parameters, I just added factory convenience methods (not `init...` ones) for the "optional" arguments (`maximumValueDescription`; `minimumValueDescription`; and `vertical`, which I guess it can default to `NO`).

I felt that it'd make scale usage less verbose, but feel free to close unmerged if you don't want these.

jwe-apple:
The factory methods are not great in Swift, so don't want to extend them more than strictly necessary. Thanks for the idea though.

Issue #202, opened 2015-05-19
Auto layout warnings


jwe-apple:
I've not seen any actually failed layout, but I'm getting auto layout spew on the console in forms in ORKTest.

Definitely a regression.

brucehappy:
I think the spew may be due to ImageCapture.  It wasn't setting `self.translatesAutoresizingMaskIntoConstraints` to NO in its two views nor was it doing it on the subviews during `init` (before `updateContraints`).  I've updated #180 with a commit for this.  Can you check if that was the cause?

jwe-apple:
I'm seeing it mostly in forms, actually

jwe-apple:
@rsanchezsaez Assigning to you as you've been making the auto layout changes. Sorry if it's not you!

jwe-apple:
Still happening.

rsanchezsaez:
Could you provide steps to reproduce? I tried `ORKTest`'s *miniform* on iPhone 6 and iPad 4th gen. and was not able to reproduce.

jwe-apple:
ORKTest, iPhone 6, iOS 8.3, Mini forms step 3. Scroll.

If you can't reproduce, I'll take a look tomorrow or at the weekend.

brucehappy:
I have only been able to get a layout error message (and failed layout) from the vertical scale slider, and only on orientation change.  I have checked, and my original branch https://github.com/brucehappy/ResearchKit/tree/issue_4 does not show the issue, so this is a regression.  Here's the message:
```
2015-05-22 04:03:43.402 ORKTest[28041:588120] Unable to simultaneously satisfy constraints.
	Probably at least one of the constraints in the following list is one you don't want. Try this: (1) look at each constraint and try to figure out which you don't expect; (2) find the code that added the unwanted constraint or constraints and fix it. (Note: If you're seeing NSAutoresizingMaskLayoutConstraints that you don't understand, refer to the documentation for the UIView property translatesAutoresizingMaskIntoConstraints) 
(
    "<NSLayoutConstraint:0x7fdbcceb3970 H:|-(51)-[ORKScaleSlider:0x7fdbcce852a0]   (Names: '|':ORKScaleSliderView:0x7fdbcce85140 )>",
    "<NSLayoutConstraint:0x7fdbcceb3ab0 H:[ORKScaleSlider:0x7fdbcce852a0]-(51)-|   (Names: '|':ORKScaleSliderView:0x7fdbcce85140 )>",
[...]

PR #225, opened 2015-05-25
Fix autolayout warnings after rotation on form with vertical scales

rsanchezsaez:
Fixes [Issue #202](https://github.com/ResearchKit/ResearchKit/issues/202).

Also simplifies vertical scale autolayout code: I have replaced the `ORKScreenMetricVerticalScaleHorizontalMargin` metric (which was an indirect way of calculating the vertical scale height to avoid scrolling on *iPhone 5s* and up) with the simpler `ORKScreenMetricVerticalScaleHeight`. This avoids the use of the two layered delegate to obtain the scale height.


jwe-apple:
r=me, but need to test.

jwe-apple:
Looks good.

PR #229, opened 2015-05-26
Fix overviews' example code warnings by using fenced code blocks

rsanchezsaez:
I have submitted an [*appledoc* pull request](https://github.com/tomaz/appledoc/pull/535) which allows us to fix the spurious cross reference warnings in the sample code blocks within the overview markdown documents.

This *ResearchKit* pull request makes the sample code in the markdown overviews *fenced code blocks* (these don't interfere with the current appledoc version either), and removes the *echo warnings* in the docs building script.

If you update to *appledoc v2.2.1 (build 1334)* (find it in the above PR) and merge this, the warnings when building the documentation will be gone.


jwe-apple:
Looks good. r=me

PR #232, opened 2015-05-29
Device motion reaction time code improvements

rsanchezsaez:
This spun off while I was refactoring the constraints code. I'm submitting it as a separate PR because it fixes an erroneous `ORKDeviceMotionReactionTimeStep`'s *copy* method. 

Also:
- Refactors and simplifies constraint code.
- Conforms to whitespace conventions.
- Replaces `true`/`false` usage by `YES`/`NO`.
- Uses `SystemSoundID` instead of `UInt32`.


jwe-apple:
Looks good. r=me

jwe-apple:
Yep, ok. None of these changes are too big, so merging for 1.1.

PR #238, opened 2015-05-31
Only attempt to create outputDirectory if it is non-nil

### LINKED ISSUE #236: Setting taskViewController's outputDirectory to nil throws an exception
During state restoration, sometimes the outputDirectory can fail to load. This should not result in an exception - it should be accepted gracefully (developer can then set the outputDirectory to fix it up).

jwe-apple:
Fixes #236. If something attempts to set the outputDirectory to nil, before presenting the task view controller, no exception should be thrown.


rsanchezsaez:
Reviewed: fix looks good.

Issue #234, opened 2015-05-30
Camera feed in image capture task does not rotate correctly


jwe-apple:
In the image capture task, under screen rotation, the rotation does not correctly rotate the camera feed to match.

brucehappy:
This was working in my original feature branch.  I'll check master in a little while to see whats up.

jwe-apple:
cool, thanks.

brucehappy:
I just tested this on my iPhone 5 iOS 8.3 using ORKTest, and it rotates along with the app (which excludes upside down portrait).  Which device and os is this happening on?

jwe-apple:
mm, yes. I'm also having trouble reproducing, so I'm returning this to the original reporter. Thanks for looking into it.

jwe-apple:
I've reproduced on iPad. Fix in #239.

jwe-apple:
Fixed in #239

PR #239, opened 2015-05-31
Issue #234 - Fix rotation behavior on iPad.

jwe-apple:
Existing listened for device orientation notifications, rather than status bar orientation.
Status bar orientation is the right one to listen for.


rsanchezsaez:
I am not able to reproduce the issue running current *master* on an iPad 4th gen. (iOS 8.3), do you have reproducing steps? 

Anyway, I have reviewed the fix and looks good. On my device, it continues to work as expected after merging.

PR #240, opened 2015-05-31
Reaction time task: fix issues

rsanchezsaez:
This PR fixes three issues:

- *[Regression]* Failure animation was not showing if you shook the device before the blue dot appeared.
- Failure animation could be triggered a second time if you shook the device while the timeout failure animation was being shown.
- Reaction time result could cause a crash in `ORKCatalog` because the `timestamp` row was selectable.


jwe-apple:
r=me

PR #246, opened 2015-06-01
Improvements to image capture active step

rsanchezsaez:
Some improvements to the image capture active step:

- Redrew the hand capture example images in *ORKTest* (the previous images had some minor visual artifacts). Also added proper @3x images, scaled from a vectorial source.
- Made template image to inset from the actual video frame instead of the superview frame (the superview frame goes to the sides of the screen, but in some instances the video frame doesn't; run on iPad before and after to see the difference).
- Add quick fade in/out animation to template image.
- Two additional minor fixes.


ResearchKit/Common/ORKImageCaptureCameraPreviewView.m
```diff
@@ -58,31 +58,62 @@ - (instancetype)init {
...
         [self.layer addSublayer:_previewLayer];
+        
+        _capturedImageView = [[UIImageView alloc] init];
+        _capturedImageView.contentMode = UIViewContentModeScaleAspectFit;
+        [self addSubview:_capturedImageView];
     
         _templateImageView = [[ORKTintedImageView alloc] init];
         _templateImageView.contentMode = UIViewContentModeScaleAspectFit;
         _templateImageView.shouldApplyTint = YES;
+        _templateImageView.hidden = 0;
```
jwe-apple: This should be NO, not 0.

jwe-apple:
Actually just going to accept this, you can fix it in a future PR or I'll pick it somewhere.
r=me

brucehappy:
So, I've just now run master after this merge, and it appears that the layout changes in this PR have changed the landscape mode of the image capture task a great deal.  The first image below is my original branch, and next is master.  I had setup the layout to use all the available space and to float the bottom buttons on top of the preview with translucency to help facilitate that.  This concept appears to have been removed.

Unfortunately, I don't have an iPad to test on, so I can't see the issue that was occurring that these layout changes resolved.  But to me, this is a step backwards regarding landscape layout on the phone.

issue_1_landscape
master_landscape

rsanchezsaez:
This was a rotation issue: if you entered the image capture step while already being in landscape mode, layout was correct. Fixed on [PR #250](https://github.com/ResearchKit/ResearchKit/pull/265).

PR #247, opened 2015-06-02
Fix picker layout centering in certain conditions.

jwe-apple:
The constraints implementation change updates to correct behavior, but did
not affect the actual bug.


jwe-apple:
Reviewed.

PR #248, opened 2015-06-02
Fix unexpected text truncation

jwe-apple:
In certain scenarios, the interaction of auto layout and the selection of
the preferredMaxLayoutWidth could result in text being abnormally
truncated.

This fix lets the labels in the headerview be full width, by reducing the
content hugging priority.

This is slightly less efficient from a rendering perspective but makes
the autolayout calculation more stable.


rsanchezsaez:
Are there steps for reproducing text truncation? :)

jwe-apple:
r=Patrick

PR #249, opened 2015-06-02
Speech rate should use default speech rate

jwe-apple:
Using a speech rate relative to the maximum speech rate is error
prone.


jwe-apple:
r=Patrick

PR #250, opened 2015-06-02
Keyboard does not hide when tapping on margins on iPad

jwe-apple:
Steps to reproduce:
- Build and deploy ORKTest to iPad sim
- Miniform -> Weight
- Tap on weight, Keyboard appears
- Tap on side of screen

Actual result:
- Keyboard does not hide

Expected result:
- Keyboard should hide.

Fix: allow parent view to specify a different tap off scope for the tap off
gesture recognizer by exposing a tapOffView property on ORKTableContainerView.

This has no API impact, since ORKTableContainerView is internal to the framework.


jwe-apple:
r=Patrick

PR #251, opened 2015-06-02
Reaction time task does not time out in sim

jwe-apple:
The reaction time task was untestable in the simulator. Here, we add
support for both timing out the task in the simulator, and for using the
shake gesture in the simulator.

### DIFF
--- a/ResearchKit/ActiveTasks/ORKDeviceMotionReactionTimeViewController.m
+++ b/ResearchKit/ActiveTasks/ORKDeviceMotionReactionTimeViewController.m
@@ -88,10 +88,12 @@ static const NSTimeInterval OutcomeAnimationDuration = 0.3;
 {
     if(event.type == UIEventSubtypeMotionShake)
     {
-        ORKDeviceMotionReactionTimeResult *reactionTimeResult = [[ORKDeviceMotionReactionTimeResult alloc] initWithIdentifier:self.step.identifier];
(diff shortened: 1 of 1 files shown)


ResearchKit/ActiveTasks/ORKDeviceMotionReactionTimeViewController.m
```diff
@@ -80,8 +80,23 @@ - (void)viewWillDisappear:(BOOL)animated {
...
 }
 
+#if TARGET_IPHONE_SIMULATOR
+- (void)motionBegan:(UIEventSubtype)motion withEvent:(UIEvent *)event
+{
+    if(event.type == UIEventSubtypeMotionShake)
+    {
+        ORKDeviceMotionReactionTimeResult *reactionTimeResult = [[ORKDeviceMotionReactionTimeResult alloc] initWithIdentifier:self.step.identifier];
+        reactionTimeResult.timestamp = _stimulusTimestamp;
+        [_results addObject:reactionTimeResult];
```
rsanchezsaez: This needs an `if (_validResult)`check before adding the result to the results array, same as with `recorder:didCompleteWithResult:`.

Otherwise, triggering a shake gesture while the blue dot is not showing stores a valid result and consumes and attempt (even though the failure animation is correctly shown).

rsanchezsaez:
Reviewed, small issue reported. Otherwise all good.

PR #252, opened 2015-06-02
Exception when saving and restoring image capture step.

jwe-apple:
If the app is terminated while on an image capture step, and restarted
by build and run from Xcode, an exception would occur while restoring.

This is a symptom of serializing NSURL's directly, rather than serializing
either a relative path or bookmark data. This matters because the path can
change after a reinstall or restore from backup.

Here we make two changes. First, the output directory in task and step
view controller is serialized as bookmark data rather than directly as a URL.
Secondly, for results serialization, NSURL references are stored as paths
relative to the NSHomeDirectory (i.e. the root of the writable sandbox).

### DIFF
--- a/ResearchKit/Common/ORKHelpers.m
+++ b/ResearchKit/Common/ORKHelpers.m
@@ -431,11 +431,11 @@ NSData *ORKBookmarkDataFromURL(NSURL *url) {
 }
 
 NSString *ORKPathRelativeToURL(NSURL *url, NSURL *baseURL) {
-    NSURL *url1 = [url URLByStandardizingPath];
-    NSURL *baseURL1 = [baseURL URLByStandardizingPath];
+    NSURL *standardizedURL = [url URLByStandardizingPath];
+    NSURL *standardizedBaseURL = [baseURL URLByStandardizingPath];
     
-    NSString *path = [url1 absoluteString];
-    NSString *basePath = [baseURL1 absoluteString];
+    NSString *path = [standardizedURL absoluteString];
+    NSString *basePath = [standardizedBaseURL absoluteString];
     
     if ([path hasPrefix:basePath]) {
         NSString *relativePath = [path substringFromIndex:[basePath length]];
@@ -448,7 +448,7 @@ NSString *ORKPathRelativeToURL(NSURL *url, NSURL *baseURL) {
     }
 }
 
-static NSURL *homeDirectory() {
+static NSURL *ORKHomeDirectory() {
     static NSURL *homeDirectory = nil;
     static dispatch_once_t onceToken;
     dispatch_once(&onceToken, ^{
@@ -462,7 +462,18 @@ NSURL *ORKURLForRelativePath(NSString *relativePath) {
         return nil;
     }
     
-    return [NSURL fileURLWithPath:relativePath relativeToURL:homeDirectory()];
+    NSURL *homeDirectory = ORKHomeDirectory();
+    CFURLRef url = CFURLCreateWithFileSystemPathRelativeToBase(kCFAllocatorDefault, (CFStringRef)relativePath, kCFURLPOSIXPathStyle, false, (CFURLRef)homeDirectory);
+    if (url != NULL) {
+        NSNumber *isDirectory;
+        if ([(__bridge NSURL *)url getResourceValue:&isDirectory forKey:NSURLIsDirectoryKey error:nil] && [isDirectory boolValue]) {
+            CFRelease(url);
+            url = CFURLCreateWithFileSystemPathRelativeToBase(kCFAllocatorDefault, (CFStringRef)relativePath, kCFURLPOSIXPathStyle, true, (CFURLRef)homeDirectory);
+        }
+    }
+    
+    NSURL *nsurl = (__bridge_transfer NSURL *)url;
+    return nsurl;
 }
 
 NSString *ORKRelativePathForURL(NSURL *url) {
@@ -470,7 +481,7 @@ NSString *ORKRelativePathForURL(NSURL *url) {
         return nil;
     }
     
-    return ORKPathRelativeToURL(url, homeDirectory());
+    return ORKPathRelativeToURL(url, ORKHomeDirectory());
 }
 
 id ORKDynamicCast_(id x, Class objClass) {
--- a/ResearchKit/Common/ORKHelpers.m
+++ b/ResearchKit/Common/ORKHelpers.m
@@ -395,6 +395,84 @@ UIFont *ORKLightFontWithSize(CGFloat size) {
     return font;
 }
 
+NSURL *ORKURLFromBookmarkData(NSData *data) {
+    if (data == nil) {
+        return nil;
+    }
+    
+    BOOL bookmarkIsStale = NO;
+    NSError *bookmarkError = nil;
+    NSURL *bookmarkURL = [NSURL URLByResolvingBookmarkData:data
+                                                   options:NSURLBookmarkResolutionWithoutUI
+                                             relativeToURL:nil
+                                       bookmarkDataIsStale:&bookmarkIsStale
+                                                     error:&bookmarkError];
+    if (! bookmarkURL) {
+        ORK_Log_Debug(@"Error loading URL from bookmark: %@", bookmarkError);
+    }
+    
+    return bookmarkURL;
+}
+
+NSData *ORKBookmarkDataFromURL(NSURL *url) {
+    if (! url) {
+        return nil;
+    }
+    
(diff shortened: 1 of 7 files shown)


ResearchKit/Common/ORKHelpers.m
```diff
@@ -395,6 +395,84 @@ CGFloat ORKTableViewLeftMargin(UITableView *tableView) {
...
+    
+    BOOL bookmarkIsStale = NO;
+    NSError *bookmarkError = nil;
+    NSURL *bookmarkURL = [NSURL URLByResolvingBookmarkData:data
+                                                   options:NSURLBookmarkResolutionWithoutUI
+                                             relativeToURL:nil
+                                       bookmarkDataIsStale:&bookmarkIsStale
+                                                     error:&bookmarkError];
+    if (! bookmarkURL) {
+        ORK_Log_Debug(@"Error loading URL from bookmark: %@", bookmarkError);
```
rsanchezsaez: `ORK_Log_Debug()` seems a bit overused currently: it's used elsewhere to log non-problemating behaviour like `viewDidAppear`, `dealloc`, etc. There are some potentially serious errors, like this one, which may benefit from a stronger logging mechanism.

Should we maybe replace `ORK_Log_Debug()` and `ORK_Log_Oops()` by `ORK_Log_Debug_Info()`, `ORK_Log_Debug_Warn()`, `ORK_Log_Debug_Error()` (and provide a way of configuring log verbosity)?

There are also things like [CocoaLumberjack] (https://github.com/CocoaLumberjack/CocoaLumberjack), but those may be going overboard.

ResearchKit/Common/ORKHelpers.m
```diff
@@ -395,6 +395,84 @@ CGFloat ORKTableViewLeftMargin(UITableView *tableView) {
...
+                                              error:&error];
+    if (! bookmark) {
+        ORK_Log_Debug(@"Error converting URL to bookmark: %@", error);
+    }
+    return bookmark;
+}
+
+NSString *ORKPathRelativeToURL(NSURL *url, NSURL *baseURL) {
+    NSURL *url1 = [url URLByStandardizingPath];
+    NSURL *baseURL1 = [baseURL URLByStandardizingPath];
```
rsanchezsaez: I don't feel entitled to make these kind of comments about your code, but I can't help to say that I'd name these `standardizedURL` and `standardizedBaseURL`. ;)

ResearchKit/Common/ORKTaskViewController.m
```diff
@@ -1279,7 +1280,7 @@ - (void)decodeRestorableStateWithCoder:(NSCoder *)coder {
     _taskRunUUID = [coder decodeObjectOfClass:[NSUUID class] forKey:_ORKTaskRunUUIDRestoreKey];
     self.showsProgressInNavigationBar = [coder decodeBoolForKey:_ORKShowsProgressInNavigationBarRestoreKey];
     
-    _outputDirectory = [coder decodeObjectOfClass:[NSURL class] forKey:_ORKOutputDirectoryRestoreKey];
+    _outputDirectory = ORKURLFromBookmarkData([coder decodeObjectOfClass:[NSData class] forKey:_ORKOutputDirectoryRestoreKey]);
```
rsanchezsaez: Might be a good idea to use the property setter here to ensure output directory creation?

ResearchKit/Common/ORKHelpers.m
```diff
@@ -395,6 +395,84 @@ CGFloat ORKTableViewLeftMargin(UITableView *tableView) {
...
+        if ([relativePath hasPrefix:@"/"]) {
+            relativePath = [relativePath substringFromIndex:1];
+        }
+        return relativePath;
+    } else {
+        return path;
+    }
+}
+
+static NSURL *homeDirectory() {
```
rsanchezsaez: Despite being static, I'd rename to `ORKHomeDirectoryURL()`.

ResearchKit/Common/ORKHelpers.m
```diff
@@ -395,6 +395,84 @@ CGFloat ORKTableViewLeftMargin(UITableView *tableView) {
...
+    });
+    return homeDirectory;
+}
+
+NSURL *ORKURLForRelativePath(NSString *relativePath) {
+    if (! relativePath) {
+        return nil;
+    }
+    
+    return [NSURL fileURLWithPath:relativePath relativeToURL:homeDirectory()];
```
rsanchezsaez: This method is not part of `NSURL`, I'm afraid.

rsanchezsaez:
Reviewed, it doesn't compile due to a missing `NSURL` method. Added some additional comments. Otherwise code looks good.

---

I didn't realize the application container directory UUID changed between Xcode launches. :)

---

Out of curiosity, what's the advantage of using a bookmark for state restoration (and `ORKConsentSection`'s `customAnimationURL` key archiving) instead of using relative URLs everywhere?

ResearchKit/Common/ORKHelpers.m
jwe-apple: You are entitled. I'll change it.

ResearchKit/Common/ORKTaskViewController.m
jwe-apple: good idea.

ResearchKit/Common/ORKHelpers.m
jwe-apple: You're probably right. Though really we just need to give framework users an easy way to turn it off rather than hijacking DEBUG. I'm not sure there's a real need for more than two levels; but there needs to be a way to turn some or all of it off without affecting your own app's console logging.

I am a bit cautious about the use of Oops, which unconditionally logs. The issue is that anything going to the console is inherently public. If the log message were to leak information about your health, that's generally not a good thing.

In any event, not a change that's going to happen in this PR. Filed #256.

ResearchKit/Common/ORKTaskViewController.m
jwe-apple: Can't use the setter because it's undefined what _hasBeenPresented might be. Should be NO, but ... safer to just ensure directory existence and be explicit about it.

ResearchKit/Common/ORKHelpers.m
jwe-apple: I tend to give static functions clearly non-extern naming, but it does look neater. Changing.

jwe-apple:
The bookmark is the documented, supported way of doing this and works wherever on the platform the file might be located. However it didn't make a lot of sense for result serialization - results ought not to use opaque formats when serialized. Since results can only be created inside the writable part of the app sandbox, the root of that is known, and we can safely paths relative to the home directory. For content that might have shipped with the app, which is not guaranteed to be beneath that root, it's better to use the bookmark.

rsanchezsaez:
I see, thanks for the info!

rsanchezsaez:
I just noticed that this made the `ORKJSONSerializationTests` to fail, because a file relative URL is not equal to the absolute URL.

I was playing with making the `NSURL`s set in `applySomeValueToClassProperty:forObject:index:forEqualityCheck:` relative, and that made the file result tests to pass, but in turn makes `ORKConsentSection`'s `customAnimationURL` check to fail because the relative `NSURL` is unarchived as an absolute one when it's transformed back from the bookmark.

How should we fix it?

jwe-apple:
Probably by changing the test to standardize the URL. I'll have a look in a bit.

rsanchezsaez:
Ok, I'm submitting a related PR which doesn't affect the tests one way or the other.

rsanchezsaez:
By the way, for the bookmarks to be successfully stored the test files must exist. I was doing this in `applySomeValueToClassProperty:forObject:index:forEqualityCheck:`

    NSURL *homeDirectoryURL = [NSURL fileURLWithPath:NSHomeDirectory()];
    NSString *filePath0 = @"Documents/blah";
    NSURL *url0 = [NSURL fileURLWithFileSystemRepresentation:filePath0.fileSystemRepresentation isDirectory:NO relativeToURL:homeDirectoryURL];
    NSString *filePath1 = @"Documents/xxx";
    NSURL *url1 = [NSURL fileURLWithFileSystemRepresentation:filePath1.fileSystemRepresentation isDirectory:NO relativeToURL:homeDirectoryURL];
    
    dispatch_once(&onceToken, ^{
        BOOL success = NO;
        success = [[NSFileManager defaultManager] createFileAtPath:url0.path
                                                          contents:nil
                                                        attributes:nil];
        success = [[NSFileManager defaultManager] createFileAtPath:url1.path
                                                          contents:nil
                                                        attributes:nil];
    });

Issue #244, opened 2015-06-01
Whether user agreed to consent review should be explicit on result


jwe-apple:
The signature result from the consent review step should be explicit as to whether the user agreed to the consent. At the moment this is implicit. See http://stackoverflow.com/questions/30571615/researchkit-is-it-possible-to-get-choices-variable-from-orkconsentreviewstep

jwe-apple:
Fixed in #254

PR #254, opened 2015-06-02
Add explicit consent review indication

jwe-apple:
Adds a `consented` property to ORKConsentSignatureResult. This property
is true if the consent has been explicitly agreed by the user, and false
otherwise.

Added ORKESerializer support and verified unit tests pass.

### DIFF
--- a/Testing/ORKTest/ORKTest/ORKESerialization.m
+++ b/Testing/ORKTest/ORKTest/ORKESerialization.m
@@ -892,8 +892,8 @@ ret =
    ENTRY(ORKConsentSignatureResult,
          nil,
          (@{
-            PROPERTY(signature, ORKConsentSignature, NSObject, NO, nil, nil),
-            PROPERTY(consented, NSNumber, NSObject, NO, nil, nil),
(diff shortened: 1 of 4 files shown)


jwe-apple:
Addresses #244

Testing/ORKTest/ORKTest/ORKESerialization.m
```diff
@@ -892,7 +892,8 @@ @implementation ORKESerializer
    ENTRY(ORKConsentSignatureResult,
          nil,
          (@{
-            PROPERTY(signature, ORKConsentSignature, NSObject, NO, nil, nil)
+            PROPERTY(signature, ORKConsentSignature, NSObject, NO, nil, nil),
+            PROPERTY(consented, NSNumber, NSObject, NO, nil, nil),
```
rsanchezsaez: Shouldn't `writeAfterInit` be `YES`for both of this properties?

jwe-apple: good catch.

PR #255, opened 2015-06-02
Update image capture task images in ORKCatalog

rsanchezsaez:
I forgot to update `ORKCatalog` to the new images. Also fixes the small `BOOL` issue elsewhere.


jwe-apple:
Great.

PR #258, opened 2015-06-02
ORKURLForRelativePath(): use NSURL methods instead of CFURL's

rsanchezsaez:
Also: minor variable renames and whitespace conforming.


jwe-apple:
r=me

PR #260, opened 2015-06-03
Fix tests serializing NSURL

jwe-apple:
Fixes test failure introduced in #252, by comparing file URLs by
their absoluteString rather than a direct equality check.


jwe-apple:
Nope, not quite right.

PR #262, opened 2015-06-03
Fix tests and rename Device Motion Reaction Time to Reaction Time

### LINKED ISSUE #257: Rename 'ORKDeviceMotionReactionTime...' to 'ORKReactionTime...'
The `ORKDeviceMotionReactionTime...` class names are a tad too long. I don't see any shortcomings to renaming them  to 'ORKReactionTime...'. It's also more in line to other device-motion using tasks like `ORKFitnessTask`.

jwe-apple:
Makes all tests pass in both ORKTest and ResearchKit.
Fixes test failures introduced in #252.

Also fixes #257, renaming to ORKReactionTime...


jwe-apple:
ok to test

ResearchKit/Common/ORKHelpers.h
```diff
@@ -175,6 +175,11 @@ ORKEqualObjects(id o1, id o2) {
     return (o1 == o2) || (o1 && o2 && [o1 isEqual:o2]);
 }
 
+ORK_INLINE BOOL
+ORKEqualURLs(NSURL *url1, NSURL *url2) {
+    return ORKEqualObjects(url1, url2) || ([url1 isFileURL] && [url2 isFileURL] && [[url1 absoluteString] isEqualToString:[url2 absoluteString]]);
+}
```
YuanZhu-apple: How about rename this to ORKEqualFileURLs?

ResearchKit/Common/ORKOrderedTask.m
```diff
@@ -283,7 +283,7 @@ - (instancetype)initWithCoder:(NSCoder *)aDecoder {
 static NSString * const ORKSpatialSpanMemoryStepIdentifier = @"cognitive.memory.spatialspan";
 static NSString * const ORKToneAudiometryPracticeStepIdentifier = @"tone.audiometry.practice";
 static NSString * const ORKToneAudiometryStepIdentifier = @"tone.audiometry";
-static NSString * const ORKDeviceMotionReactionTimeStepIdentifier = @"reactionTime.deviceMotion";
+static NSString * const ORKReactionTimeStepIdentifier = @"reactionTime.deviceMotion";
```
YuanZhu-apple: Rename the identifier string to "reactionTime" to align with the const value name?

Testing/ORKTest/ORKTest/ORKESerialization.m
```diff
@@ -833,6 +833,12 @@ @implementation ORKESerializer
            PROPERTY(outputVolume, NSNumber, NSObject, NO, nil, nil),
            PROPERTY(samples, ORKToneAudiometrySample, NSArray, NO, nil, nil),
            })),
+   ENTRY(ORKReactionTimeResult,
+         nil,
+         (@{
+            PROPERTY(timestamp, NSNumber, NSObject, NO, nil, nil),
+            PROPERTY(fileResult, ORKResult, NSObject, NO, nil, nil)
```
rsanchezsaez: Should `writeAfterInit` be `YES`? If that's the case, the issue affects most of the result subclasses.

Testing/ORKTest/ORKTestTests/ORKJSONSerializationTests.m
```diff
@@ -569,13 +577,30 @@ - (void)testORKSecureCoding {
...
         
         NSData *data2 = [NSKeyedArchiver archivedDataWithRootObject:newInstance];
+        
+        NSKeyedUnarchiver *unarchiver2 = [[NSKeyedUnarchiver alloc] initForReadingWithData:data2];
+        unarchiver2.requiresSecureCoding = YES;
+        unarchiver2.delegate = self;
+        id newInstance2 = [unarchiver2 decodeObjectOfClasses:[NSSet setWithArray:classesWithSecureCoding] forKey:NSKeyedArchiveRootObjectKey];
+        NSData *data3 = [NSKeyedArchiver archivedDataWithRootObject:newInstance2];
+        
         if (![data isEqualToData:data2]) { // allow breakpointing
```
rsanchezsaez: This if check to allow breakpointing seems unneeded. One can use the "Test Failure" breakpoint instead.

Testing/ORKTest/ORKTest/ORKESerialization.m
jwe-apple: writeAfterInit only needs to be YES if an initializer block is specified. If there's no init block, all parameters are *assumed* to be writeAfterInit since they can't be written in the initializer.

rsanchezsaez: Ah, I see. Still, wouldn't it be clearer to be explicit about it?

jwe-apple: Perhaps, but also more error prone.

rsanchezsaez: Why error prone? The opposite relies on a not so clear assumption, no? :)

Issue #242, opened 2015-06-01
State restoration makes the image capture step to fail


rsanchezsaez:
Restoring the app to an image capture step makes the view controller unstable.

*Steps to reproduce*
1. Launch `ORKTest` from Xcode.
2. Open the *Image Capture* task.
3. Go to step 4 (image capture step).
4. Send the app to background and hit stop on Xcode so the state is stored.
5. Launch `ORKTest` again, it should restore to the image capture step.
6. Try capturing an image.

If you try to capture an image after state restoration, a zoomed portion of the image is shown on screen and a *"The captured image could not be saved"* error is shown.

If you then tap "*Recapture image*", the layout of the hand template breaks and the error message is not dismissed. If you press "Capture", it goes to the next step instead of capturing a new image. Capture also fails when you reach *step 7* (additional image capture step).

Ending the task and restarting it fixes it.

jwe-apple:
Is the output directory correctly set on the image capture step VC during restoration?

jwe-apple:
Reproduced.

brucehappy:
I'd love to reproduce and fix, but I'm tied up today.  The soonest I could get to it is tomorrow afternoon.

rsanchezsaez:
> Is the output directory correctly set on the image capture step VC during restoration?

The problem seems to be related to the output directory indeed. The output directory *ivar* is correctly being set during restoration. And checking with iExplorer, I can see the output directory existing on the device.

After restoration, however, `[_capturedImageData writeToURL:url options:NSDataWritingAtomic|NSDataWritingFileProtectionCompleteUnlessOpen error:&writeError]` fails with this error:

    2015-06-02 00:39:47.983 ORKTest[3750:765665] [ORK][OOPS] -[ORKImageCaptureStepViewController writeCapturedDataWithError:] 249 Error Domain=NSCocoaErrorDomain Code=4 "The operation couldn’t be completed. (Cocoa error 4.)" UserInfo=0x170266500 {NSFilePath=/var/mobile/Containers/Data/Application/AFB848AA-F633-460F-871E-F5BD600AF3C6/Documents/350FE580-3955-42AC-9F50-0BB2E0ED1A8E/right3.jpg, NSUnderlyingError=0x170249de0 "The operation couldn’t be completed. No such file or directory"}


I tried creating the output directory again just before saving the captured image, but I get this error:

[...]

PR #263, opened 2015-06-03
Image Capture Step handling of file write error (Fix for #242)

brucehappy:
Experiencing an error will now completely shut down the image capture step, including removing the continue and skip buttons, and resulting in only error text shown in the view.

#242

### DIFF
--- a/ResearchKit/Common/ORKImageCaptureView.m
+++ b/ResearchKit/Common/ORKImageCaptureView.m
@@ -140,8 +140,8 @@
     _error = error;
     _headerView.alpha = error==nil ? 0 : 1;
     _headerView.instructionLabel.text = [error.userInfo valueForKey:NSLocalizedDescriptionKey];
-    if (error && self.continueTitle) {
-        _continueSkipContainer.continueButtonItem.title = self.continueTitle;
+    if (error) {
+        _continueSkipContainer.continueButtonItem = nil;
         _continueSkipContainer.skipButtonItem = nil;
     }
     
@@ -203,6 +203,11 @@ const CGFloat CONTINUE_ALPHA_OPAQUE = 0;
 }
 
 - (void)setContinueButtonItem:(UIBarButtonItem *)continueButtonItem {
+    // If we are in an error state, then do not configure the continue button
+    if (self.error) {
+        return;
+    }
+
     // Intercept the continue button press.  This can be called multiple
     // times with the same UIBarButtonItem, or with different UIBarButtonItems,
     // so capture it whenever the button does not point to our selector
@@ -214,9 +219,9 @@ const CGFloat CONTINUE_ALPHA_OPAQUE = 0;
         continueButtonItem.target = self;
     }
     
-    // If we do not have an error to show, and we haven't already gotten a captured
-    // image, then change the title of the button to be appropriate for the capture action
-    if (!self.capturedImage && !self.error) {
+    // If we haven't already gotten a captured image, then change the title
+    // of the button to be appropriate for the capture action
+    if (!self.capturedImage) {
         continueButtonItem.title = ORKLocalizedString(@"CAPTURE_BUTTON_CAPTURE_IMAGE", nil);
     } else if(self.capturedImage) {
         _continueSkipContainer.skipButtonItem = _skipButtonItem;
@@ -241,9 +246,13 @@ const CGFloat CONTINUE_ALPHA_OPAQUE = 0;
                 // Hide the template image after capturing
                 _previewView.templateImageHidden = YES;
         
-                // Reset the continue button title and configure the skip button as a recapture button
-                _continueSkipContainer.continueButtonItem.title = self.continueTitle;
-                _continueSkipContainer.skipButtonItem = _skipButtonItem;
+                // If we experienced an error during the capture (likely writing the file to disk)
+                // then do not reconfigure the buttons
+                if (!self.error) {
+                    // Reset the continue button title and configure the skip button as a recapture button
+                    _continueSkipContainer.continueButtonItem.title = self.continueTitle;
+                    _continueSkipContainer.skipButtonItem = _skipButtonItem;
+                }
             }
             // Stop ignoring presses
             _capturePressesIgnored = NO;
--- a/ResearchKit/Common/ORKImageCaptureStepViewController.m
+++ b/ResearchKit/Common/ORKImageCaptureStepViewController.m
@@ -207,12 +207,23 @@
 }
 
 - (void)handleError:(NSError *)error {
-    // Tell the task view controller that we have failed so that it removes our result
-    STRONGTYPE(self.delegate) strongDelegate = self.delegate;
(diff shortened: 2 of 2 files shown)


rsanchezsaez:
I think that the *Continue* button should be present on error. If an user has a problem with the camera, the researcher might want for her to be able to continue the task anyway, don't you think? And maybe the *Recapture Image" button too, in case of a transient camera error.

rsanchezsaez:
Also, thinking a bit more about it, recalculating the constraints on rotation looks fine, but doing so after an error seems overboard. Why not have the error label and the preview view properly layout, and just hide/show the appropriate ones on error?

ResearchKit/Common/ORKImageCaptureView.m
```diff
@@ -140,8 +140,8 @@ - (void)setError:(NSError * __nullable)error {
     _error = error;
     _headerView.alpha = error==nil ? 0 : 1;
     _headerView.instructionLabel.text = [error.userInfo valueForKey:NSLocalizedDescriptionKey];
-    if (error && self.continueTitle) {
-        _continueSkipContainer.continueButtonItem.title = self.continueTitle;
+    if (error) {
+        _continueSkipContainer.continueButtonItem = nil;
         _continueSkipContainer.skipButtonItem = nil;
     }
```
jwe-apple: Shouldn't the other case should be handled here, i.e. restoring the correct continue button item if error -> nil?

ResearchKit/Common/ORKImageCaptureView.m
```diff
@@ -203,6 +203,11 @@ - (UIBarButtonItem *)continueButtonItem {
 }
 
 - (void)setContinueButtonItem:(UIBarButtonItem *)continueButtonItem {
+    // If we are in an error state, then do not configure the continue button
+    if (self.error) {
+        return;
```
jwe-apple: And in this case, shouldn't the continueButtonItem that's been set still be saved, just not applied to the continueSkipContainer? So that if the error is cleared, you can restore the correct item.

ResearchKit/Common/ORKImageCaptureView.m
```diff
@@ -214,9 +219,9 @@ - (void)setContinueButtonItem:(UIBarButtonItem *)continueButtonItem {
...
     }
     
-    // If we do not have an error to show, and we haven't already gotten a captured
-    // image, then change the title of the button to be appropriate for the capture action
-    if (!self.capturedImage && !self.error) {
+    // If we haven't already gotten a captured image, then change the title
+    // of the button to be appropriate for the capture action
+    if (!self.capturedImage) {
         continueButtonItem.title = ORKLocalizedString(@"CAPTURE_BUTTON_CAPTURE_IMAGE", nil);
     } else if(self.capturedImage) {
```
jwe-apple: Not this patch, but this needs a space before (

ResearchKit/Common/ORKImageCaptureView.m
```diff
@@ -241,9 +246,13 @@ - (void)capturePressed {
...
         
-                // Reset the continue button title and configure the skip button as a recapture button
-                _continueSkipContainer.continueButtonItem.title = self.continueTitle;
-                _continueSkipContainer.skipButtonItem = _skipButtonItem;
+                // If we experienced an error during the capture (likely writing the file to disk)
+                // then do not reconfigure the buttons
+                if (!self.error) {
+                    // Reset the continue button title and configure the skip button as a recapture button
+                    _continueSkipContainer.continueButtonItem.title = self.continueTitle;
+                    _continueSkipContainer.skipButtonItem = _skipButtonItem;
```
jwe-apple: Again, want to handle both cases properly. Just factor it out into a method that checks self.error and does the right thing.

ResearchKit/Common/ORKImageCaptureView.m
rsanchezsaez: Shouldn't we allow skipping the question on error?

jwe-apple: Subject to the step's "optional" property, yes.

jwe-apple:
When you update this patch, could you submit it against the convergence branch rather than master, please?

brucehappy:
I have been away, and should be able to look at this today (2015-06-05).  I see @YuanZhu-apple has submitted a modified version of this PR (PR #269) but it does not appear to resolve what I see as your central issue, namely being able to get back into a fully working state by setting the error to nil after it was set to non-nil.

There are currently three handled errors, and they generally fall into either, no permissions, no camera, or file write error.  When @rsanchezsaez made his comments in #242 about how the step responded to the file write error, I used this PR to turn off everything once an error has occurred.  A no permissions error will manifest itself as not even being able to get to this step in the first place.  A no camera error can never be "resolved" as you cant add a camera to the device and continue.  And the file write error is likely due to something wrong with the outputDirectory, which would require the app embedding ORK to handle outside of the VC.  All of these scenarios suggested to me that forcing them to either go back or Cancel was ok (getting rid of the continue buttons).  In general, I'm not sure how the error would be "undone".  There is no public method/property exposed that would call handleError with nil.  But if thats whats needed, then I need to modify this PR so that there is a way to do that.  On setting a nil error, I would also need to re-setup the capture session and still image output, and start the capture process again, just as if they hit back and then Next and started a new instance of the VC.

I completely missed the concept of "optional" in the image capture step.  So, it certainly sounds like in the case of an error on an optional image capture step, you should get a skip button below the error message.  This actually points out another issue, related to non-error state. When you are in the initial state of the step, there is only a Capture Image button, and no Skip this step button.  Once the Image is captured, the Capture Image button changes to Next, and the skip button is repurposed to say Recapture Image.  So, if we are to support optional properly, I would think that when the Capture Image button is visible, a Skip this step button should also be available.  Would it be ok, if the rest of the process remained the same?  If they have captured an image, then there would be a Next button, and a Recapture button.  If they decided to skip at this phase, they would need to click Recapture, and then the proposed Skip this step button would appear that they could click on that.

@jwe-apple thoughts on these issues?

brucehappy:
Shortly, I will be closing this PR and opening a new one with this feedback taken into account, and made against the convergence branch.  I'd change this PR, but
https://github.com/isaacs/github/issues/18

YuanZhu-apple:
@brucehappy I agree with you that errors in imageCaptureStep not likely recoverable with the step. So we don't need to spend too much time on that. But we do need to give user a way out from the error state: either go back or skip to next step. 

And your are right about honoring the step's `optional` attribute. If a step is optional, make skip button available under `capture` button when no image was taken; and when an image was taken, user need to tap `recapture` to go to a capturing state to skip.

Really appreciate your PR, and we are under pressure to release 1.1 ASAP with this issue fixed. 
After your new PR is out, I am going to close mine.

YuanZhu-apple:
@brucehappy When do you think your new PR will be ready?

brucehappy:
10 mins

YuanZhu-apple:
That's awesome!

brucehappy:
Opened PR #272

PR #265, opened 2015-06-03
Fix Image Capture View constraints not being updated after a rotation

rsanchezsaez:
After the iPad rotation fix in ~~PR 250~~ [PR #239](https://github.com/ResearchKit/ResearchKit/pull/239) constraints were not being updated (sorry I missed that in my review). Recalculation is needed because in landscape the bottom view overlaps the video preview.


jwe-apple:
#250 wasn't a rotation fix. Which PR do you mean?

rsanchezsaez:
Sorry, I meant #239.

ResearchKit/Common/ORKImageCaptureView.m
```diff
@@ -156,35 +149,57 @@ - (void)setError:(NSError * __nullable)error {
 const CGFloat CONTINUE_ALPHA_OPAQUE = 0;
 
 - (void)updateConstraints {
+    if (_variableConstraints) {
```
jwe-apple: My understanding of our reading of the documentation is that removal of old constraints should happen before setNeedsUpdateConstraints, rather than in the updateConstraints method. Is that not your understanding?

ResearchKit/Common/ORKImageCaptureView.m
```diff
@@ -119,6 +115,7 @@ - (void)orientationDidChange {
     }
     [_previewView setVideoOrientation:orientation];
     [self.delegate videoOrientationDidChange:orientation];
+    [self setNeedsUpdateConstraints];
```
jwe-apple: So, out of this whole change, this seems to be the *only* functional change. Is that also your understanding?

jwe-apple: The rest of the changes are nice enough but not needed for release. I'd like a new PR direct to the convergence branch which adds this line only.

ResearchKit/Common/ORKImageCaptureView.m
rsanchezsaez: From `updateConstraints` documentation:

    You must not invalidate any constraints as part of your constraint update phase. You also must not invoke a layout or drawing phase as part of constraint updating.

I think that *invalidating constraints* means doing anything that would potentially call `setNeedsUpdateConstraints`. I'm assuming that *invalidate* != *deactivate*, so I figured we can remove or deactivate constraints in `updateConstraints`. In addition, *ResearchKit* is already removing constraints this way in several other places and it doesn't seem to be a problem.

---

But I agree that the docs aren't very clear on this. Could you possibly get an authoritative answer?

ResearchKit/Common/ORKImageCaptureView.m
rsanchezsaez: The removal of constraints also needs to be back to avoid duplicate ones. I will submit the convergence PR tonight.

ResearchKit/Common/ORKImageCaptureView.m
jwe-apple: Yes, I'm aware we're doing it elsewhere, but I recalled discussing that exact line a few weeks back. But yes, that's is a good interpretation - the documentation is just saying you should avoid creating an infinite recursion. Ignore my comment.

jwe-apple:
@rsanchezsaez Please just update this patch now I've merged convergence back to master.

YuanZhu-apple:
@rsanchezsaez Can you update you patch so I can get this merged?

rsanchezsaez:
Updated.

Issue #266, opened 2015-06-03
"Learn More" modal views show "Cancel", would "Done" be preferable?


p2:
When going through consenting, there's always a "Cancel" button top right to abort consenting. When tapping "Learn More" on the individual steps, the learn more view is presented modal and also has a "Cancel" button top right. I feel it would be nicer if it was a "Done" button, both semantically and also to distinguish it from canceling consenting.

jwe-apple:
Agree.

YuanZhu-apple:
Fix in #268

PR #268, opened 2015-06-05
"Learn More" modal views: replace "Cancel" with "Done"  (Fix for #266)

YuanZhu-apple:
Quick fix for #266


jwe-apple:
Seems like we really ought to change the method name too. Shouldn't be any side effects, should there?

jwe-apple:
r=me

PR #271, opened 2015-06-05
In FormItemCell, lower a heightConstraint's priority 

YuanZhu-apple:
To avoid conflicts with system supplied UIView-Encapsulated-Layout-Height constraint.


jwe-apple:
r=me

PR #269, opened 2015-06-05
Improve Image Capture Step error handling (Fix for #242) based on PR #263

YuanZhu-apple:
Thanks @brucehappy, this PR is mostly based his PR #263 

With a few changes to address review comments in PR #263:
1. Recover button states when the error is cleared.
2. Improved error handling in (setContinueButtonItem:)
3. Fixed a bug in layout code

If current step is not optional, next button will be shown but disabled. Please a screenshot below.
img_0258

PR #272, opened 2015-06-06
Image Capture Step optional support

brucehappy:
This is a continuation of PR #263 taking into account @YuanZhu-apple 's PR #269.

Added support for optional to Image Capture Step.
Simplified the saving and using of continue and skip buttons.
Made the ORKCatalog use of Image Capture Step non-optional.

From PR #263:
Experiencing an error will now completely shut down the image capture step, including removing the continue and skip buttons, and resulting in only error text shown in the view.

### DIFF
--- a/ResearchKit/Common/ORKImageCaptureView.m
+++ b/ResearchKit/Common/ORKImageCaptureView.m
@@ -232,6 +232,15 @@ const CGFloat CONTINUE_ALPHA_OPAQUE = 0;
     }
 }
 
+- (void)setContinueButtonItem:(UIBarButtonItem *)continueButtonItem {
+    _continueButtonItem = continueButtonItem;
+    
+    // If the capture button is not currently being used as the continue button, then use this new button
+    if (_continueSkipContainer.continueButtonItem != _captureButtonItem) {
+        _continueSkipContainer.continueButtonItem = continueButtonItem;
+    }
+}
+
 - (void)capturePressed {
     // If we are still waiting for the delegate to complete, ignore futher presses
     if (_capturePressesIgnored)
--- a/ResearchKit/Common/ORKImageCaptureView.m
+++ b/ResearchKit/Common/ORKImageCaptureView.m
@@ -76,7 +76,6 @@
         _continueSkipContainer.bottomMargin = 15;
         _continueSkipContainer.optional = YES;
         _continueSkipContainer.backgroundColor = ORKColor(ORKBackgroundColorKey);
-        _continueSkipContainer.continueButtonItem = _captureButtonItem;
         [self addSubview:_continueSkipContainer];
         
         NSDictionary *dictionary = NSDictionaryOfVariableBindings(self, _previewView, _continueSkipContainer, _headerView);
@@ -84,6 +83,8 @@
         
         [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(orientationDidChange) name:UIApplicationDidChangeStatusBarOrientationNotification object:nil];
         [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(queue_sessionRunning) name:AVCaptureSessionDidStartRunningNotification object:nil];
+        
+        [self updateAppearance];
     }
     return self;
 }
@@ -130,10 +131,15 @@
     _showSkipButtonItem = imageCaptureStep.optional;
 }
 
-- (void)setCapturedImage:(UIImage * __nullable)capturedImage {
-    _previewView.capturedImage = capturedImage;
-    
-    if (capturedImage) {
+- (void)updateAppearance {
+    if (self.error) {
+        // Hide the template image if there is an error
+        _previewView.templateImageHidden = YES;
+        
+        // Show skip, if available, and hide the template and continue/capture button
+        _continueSkipContainer.continueButtonItem = nil;
+        _continueSkipContainer.skipButtonItem = _skipButtonItem;
+    } else if (self.capturedImage) {
         // Hide the template image after capturing
         _previewView.templateImageHidden = YES;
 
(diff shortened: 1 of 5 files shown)


ResearchKit/Common/ORKImageCaptureView.m
```diff
@@ -197,31 +221,15 @@ - (void)setSession:(AVCaptureSession * __nullable)session {
...
-    } else if(self.capturedImage) {
-        _continueSkipContainer.skipButtonItem = _skipButtonItem;
+        // If the recapture button is not currently being used as the skip button, then use this new button
+        if (_continueSkipContainer.skipButtonItem != _recaptureButtonItem) {
+            _continueSkipContainer.skipButtonItem = skipButtonItem;
+        }
     }
-    
-    _continueSkipContainer.continueButtonItem = continueButtonItem;
 }
```
YuanZhu-apple: Is `_showSkipButtonItem` necessary here? I think it should leave ORKStepViewController to handle `optional`. (StepVC will set nil or leave it unset)

And even if a step is not optional, a skip button can still be set by application, and we need to make it possible.

I think this would just work:
```
- (void)setSkipButtonItem:(UIBarButtonItem *)skipButtonItem {
        _skipButtonItem = skipButtonItem;
        // If the recapture button is not currently being used as the skip button, then use this new button
        if (_continueSkipContainer.skipButtonItem != _recaptureButtonItem) {
            _continueSkipContainer.skipButtonItem = skipButtonItem;
        }
}
```

Similar to skipButtonItem, it is also nice to have a setter for continueButtonItem:
```
- (void) setContinueButtonItem:(UIBarButtonItem *) continueButtonItem {
        _continueButtonItem = continueButtonItem;
        // If the recapture button is not currently being used as the skip button, then use this new button
        if (_continueSkipContainer.continueButtonItem != _captureButtonItem) {
            _continueSkipContainer.continueButtonItem = continueButtonItem;
        }
}
```

brucehappy: The ``ORKStepViewController.setupButtons`` function sets a non-nil skipButtonItem on ``ORKImageCapturetepViewController`` no matter what ``optional`` is set to in the ``ORKImageCaptureStep``.  Therefore, this check is to ensure that we only show the VC provided skip button item in the ``ORKNavigationContainerView`` inside the ``ORKImageCaptureView`` if optional is YES.

I can add the continue button item setter if you like.

YuanZhu-apple:
I tested on a device and found a bug:

Restoring the app to an image capture step "Done" button was shown instead of "Next".

Steps to reproduce:
img_0259

1. Launch ORKTest from Xcode.
2. Open the Image Capture task.
3. Go to step 4 (image capture step).
4. Take a picture, "Next" button is correct.
5. Send the app to background and hit stop on Xcode so the state is stored.
6. Launch ORKTest again, it should restore to the image capture step.
7. "Done" button was shown instead of "Next"

brucehappy:
Yeah, the same thing happens if you capture an image, hit next, and then hit back.  I assumed this was "normal" since none of my code changes the continue button text.  I'll debug and see how its getting changed to Done.

ResearchKit/Common/ORKImageCaptureView.m
YuanZhu-apple: Oops, you are right, we need `_showSkipButtonItem `. 
Just add the continue button item setter.

brucehappy:
Added the setter, which was the cause of the "Done" being shown.  The Done button was being set initially, and then once the delegate was being set on the VC, it called setupButtons again with "Next" as the continue, but without the setter it wasn't being updated on the ``ORKNavigationContainerView``.  Great catch (and a lesson for me not to assume something is "normal").

brucehappy:
I'm done "refining".  Looking forward to your review.

YuanZhu-apple:
Looks really good now. 
@jwe-apple  Shall we merge it?

ResearchKit/Common/ORKImageCaptureView.m
```diff
@@ -140,10 +157,18 @@ - (UIImage *)capturedImage {
...
+    _headerView.instructionLabel.text = error==nil ? nil : [error.userInfo valueForKey:NSLocalizedDescriptionKey];
+    
+    if (error) {
+        // Show skip, if available, and hide the template and continue/capture button
+        _previewView.templateImageHidden = YES;
+        _continueSkipContainer.continueButtonItem = nil;
+        _continueSkipContainer.skipButtonItem = _skipButtonItem;
+    } else {
+        // Reset to running state
+        _previewView.templateImageHidden = NO;
```
jwe-apple: Can it ever be true that we set the error to nil, and that there is a captured image present?

I think this code and the code in setCapturedImage: should be factored out so there's just one set of conditionals applied, i.e., so the logic depends only on state and not on transient factors.

jwe-apple:
Great if it's all working, but, from code review perspective only, I'm concerned about the code for setting continue/skip/recapture buttons. I'd be much more confident if there were just one method which was responsible, given the current state, for bringing the view configuration into line. Right now it's distributed across multiple setters, and so it's hard to review for correctness.

ResearchKit/Common/ORKImageCaptureView.m
brucehappy: Once the error is set, the layout changes to only show the error, and no longer show the ``previewView`` containing the captured image.  So, you'd like to see the code thats in ``setCapturedImage`` and ``setError`` that deals with what to change the buttons to moved into its own function.  And so that function would have the three cases: error, captured image non-nil, and captured image nil.  Is that right?

jwe-apple: Yep. For safety I imagine you could also call it from the other setters, for example the continue and skip item setters.

brucehappy:
Agreed that this is safer, less code, and easier to read.  win-win-win.  Thanks for the feedback.

jwe-apple:
r=me on the diff, - Yuan, please verify one more time ;-)

YuanZhu-apple:
I approve this patch:)

PR #273, opened 2015-06-06
Allow caption and instruction labels to grow or shrink in step header view

brucehappy:
This PR came out of my testing for PR #272.  Specifically, if you test Image Capture Step on the simulator, you will get an error related to there being no camera.  An ``ORKStepHeaderView`` is used to display the error, and specifically, only its ``instructionLabel`` is set to have text.  The result is that the error text is not centered on screen.  If you then rotate the device to landscape and back to portrait, it will center.  My debugging suggests that this is due to vertical content hugging priority for ``captionLabel`` and ``instructionLabel``.

This PR sets the default vertical content hugging priority on both labels to Low as a default.  Then during the update of the constraint constants, if only one of the labels contains text, then that label will have its vertical hugging priority set lower than the other label.  This will allow the label with text to grow to fill, and the label without text to shrink to nothing.

As this touched a very commonly used component, and therefore has a high test burden, I submitted this against master for consideration after 1.1.


YuanZhu-apple:
Not sure if you see this in simulator, "CAPTURE_ERROR_CAMERA_NOT_FOUND" label is not displayed and the continue button is misaligned. 
Is you PR try to get this fixed?
ios simulator screen shot jun 10 2015 3 11 56 pm

brucehappy:
@YuanZhu-apple I just merged master into this branch and gave it a try, and it is working for me, please try again.  If you were testing this PR against master as it was before convergence was merged into it, then the bug in the ``ORKImageCaptureView`` where the constraints were activated inside an else was still present, thus was probably responsible for what you saw.  It should be resolved now.

ResearchKit/Common/ORKStepHeaderView.m
```diff
@@ -114,10 +114,15 @@ - (instancetype)initWithFrame:(CGRect)frame {
...
+        [_captionLabel setContentHuggingPriority:UILayoutPriorityDefaultLow forAxis:UILayoutConstraintAxisVertical];
+        [_instructionLabel setContentHuggingPriority:UILayoutPriorityDefaultLow forAxis:UILayoutConstraintAxisVertical];
         [_learnMoreButton setContentHuggingPriority:UILayoutPriorityRequired forAxis:UILayoutConstraintAxisVertical];
         [_learnMoreButton setContentCompressionResistancePriority:UILayoutPriorityRequired forAxis:UILayoutConstraintAxisVertical];
         
 #ifdef LAYOUT_DEBUG
+        _captionLabel.backgroundColor = [UIColor yellowColor];
+        _learnMoreButton.backgroundColor = [UIColor blueColor];
+        _instructionLabel.backgroundColor = [UIColor greenColor];
         self.backgroundColor = [[UIColor purpleColor] colorWithAlphaComponent:0.2];
```
YuanZhu-apple: Can we set these debug background color with alpha 0.2 or 0.3 to be consistent with other LAYOUT_DEBUG code?

ResearchKit/Common/ORKStepHeaderView.m
```diff
@@ -114,10 +114,15 @@ - (instancetype)initWithFrame:(CGRect)frame {
             [self addSubview:_instructionLabel];
         }
         
+        [_captionLabel setContentHuggingPriority:UILayoutPriorityDefaultLow forAxis:UILayoutConstraintAxisVertical];
+        [_instructionLabel setContentHuggingPriority:UILayoutPriorityDefaultLow forAxis:UILayoutConstraintAxisVertical];
```
YuanZhu-apple: Are these necessary at here? I assume `updateConstraintConstants` will always be called.

brucehappy: No, they are not strictly needed.  I was doing this to be consistent with the ``learnMoreButton`` getting its default hugging priority set here.

brucehappy:
Comments addressed.

ResearchKit/Common/ORKStepHeaderView.m
```diff
@@ -118,6 +118,9 @@ - (instancetype)initWithFrame:(CGRect)frame {
         [_learnMoreButton setContentCompressionResistancePriority:UILayoutPriorityRequired forAxis:UILayoutConstraintAxisVertical];
         
 #ifdef LAYOUT_DEBUG
+        _captionLabel.backgroundColor = [[UIColor yellowColor] colorWithAlphaComponent:0.2];
+        _learnMoreButton.backgroundColor = [[UIColor blueColor] colorWithAlphaComponent:0.2];
+        _instructionLabel.backgroundColor = [[UIColor greenColor] colorWithAlphaComponent:0.2];
         self.backgroundColor = [[UIColor purpleColor] colorWithAlphaComponent:0.2];
```
YuanZhu-apple: Second thought on this, I just found that these labels had been assigned with debug colors in method `updateConstraints`
```
        for (UIView *view in views) {
#ifdef LAYOUT_DEBUG
            view.backgroundColor = [[UIColor greenColor] colorWithAlphaComponent:0.3];
            view.layer.borderColor = [UIColor redColor].CGColor;
            view.layer.borderWidth = 1.0;
#endif
```
The newly added LAYOUT_DEBUG  color is being override, see screenshot below.

screen shot 2015-06-12 at 10 31 11 am

I like the idea to assign them with different colors, can you remove the old debug code in `updateConstraints` and move red color borders assignment code to here?

brucehappy:
Ok, removed the section in the ``updateConstraints`` function and added borders to the components in the initializer.  I changed the border color to match the background color so they can be distinguished.

Issue #259, opened 2015-06-02
Minor animation glitch when showing keyboard on a form


rsanchezsaez:
There's a minor animation glitch when showing the keyboard on a form with text field questions (e.g., `ORKHealthKitQuantityTypeAnswerFormat`).

When you tap any of the text field while the keyboard is hidden, the placeholder label performs a glitchy animation. Try `ORKTest`'s *Mini Form* to reproduce.

100% reproducible on *iPhone 6* running *iOS 8.3*. Happens both on *master* and *stable*.

PR #274, opened 2015-06-06
Minor animation glitch when showing keyboard on a form #259

niakrise:
Added saving for previous _suffixLabel.frame.


YuanZhu-apple:
This is great, the glitch in first case has been fixed. 


Can you take a look the "jumping" in the second case?

niakrise:
Ok, I fixed second case too.

YuanZhu-apple:
Great work!

PR #275, opened 2015-06-07
Update links

kant:


jwe-apple:
Sorry, these links are correct. They get converted to real links when we run appledoc, and the generated documentation appears here: http://researchkit.org/docs/docs/Overview/GuideOverview.html

PR #276, opened 2015-06-07
Minor changes to Spanish translation

kant:


jwe-apple:
Thanks for the feedback. Tracking with rdar://problem/21277286

jwe-apple:
Thanks for the suggestions. We believe the current version is ok for this release but may revisit some aspects in future.

PR #277, opened 2015-06-07
Update link

kant:


jwe-apple:
Same comment as on previous link update patch.

Issue #193, opened 2015-05-17
Entering invalid value in a form allows going to next step


rsanchezsaez:
If the participant types an invalid value in a form field (e.g., she types letters in a numeric filed) and presses next, the view controller shows an *Invalid Value* alert, but the view controller goes to next step. If you go back, the field where you entered the invalid value shows `NaN`.

The view controller should prevent from going to the next step if the participant has typed an invalid value.

PR #278, opened 2015-06-08
prevent ORKFormStepViewController proceeding to the next step if the …

coxy1989:
…participant has typed an invalid value (issue #193)

### DIFF
--- a/ResearchKit/Common/ORKFormStepViewController.m
+++ b/ResearchKit/Common/ORKFormStepViewController.m
@@ -882,7 +882,7 @@
 }
 
 - (void)formItemCell:(ORKFormItemCell *)cell answerDidChangeTo:(id)answer {
-    if (answer && cell.formItem.identifier) {
+    if (answer && cell.formItem.identifier && [cell isAnswerValid]) {
         [self setAnswer:answer forIdentifier:cell.formItem.identifier];
     } else if (answer == nil && cell.formItem.identifier) {
         [self removeAnswerForIdentifier:cell.formItem.identifier];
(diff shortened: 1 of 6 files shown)


YuanZhu-apple:
This patch fails to prevent from going to the next step if the participant has typed an invalid value.

1. Go to Mini Form's 2nd step in ORKTest App
2. Type a number in first field
3. Paste some text in second field
4. Tap next button at the bottom, alert shows but fails to prevent going to the next step.

ios simulator screen shot jun 8 2015 5 24 54 pm

coxy1989:
Ah yes, apologies. This approach covers that case.

ResearchKit/Common/ORKFormStepViewController.m
```diff
@@ -521,8 +521,19 @@ - (BOOL)allAnswered {
...
+        if (isNonNull && ![item.impliedAnswerFormat isAnswerValid:answer]) {
+            return NO;
+        }
+    }
+    return YES;
+}
+
 - (BOOL)continueButtonEnabled {
-    return ([self allAnswered] || (self.step.optional && ([self numAnswered] > 0 || ! self.skipButtonItem)));
+    return ([self allAnswersValid] && ([self allAnswered] || (self.step.optional && ([self numAnswered] > 0 || ! self.skipButtonItem))));
```
YuanZhu-apple: A little hard to read, how about splitting this one line to something like:

```
BOOL optionalButNotEmpty = self.step.optional && ([self numAnswered] > 0 || ! self.skipButtonItem);
return [self allAnswersValid] && ([self allAnswered] || optionalButNotEmpty);
```

And put some documentation around that?

PR #282, opened 2015-06-08
Vietnamese is a family name first language

jwe-apple:
Add Vietnamese to the list of such languages for consent review.


jwe-apple:
r=Yuan

PR #283, opened 2015-06-08
Revert "Vietnamese is a family name first language"

jwe-apple:
Reverts ResearchKit/ResearchKit#282


jwe-apple:
This needed to go into convergence first.

PR #285, opened 2015-06-08
Fix speech rate on iOS 8.3 and earlier

jwe-apple:
Patch earlier this week adjusted the speech rate to use the default speech
rate. This is too fast a rate on iOS 8.3 and earlier.


jwe-apple:
r=Justin

Issue #292, opened 2015-06-13
Value picker in a non optional form allows going to next step


julientherier:
If I add a value picker in a non optional form step, I can go to next step if:
• I select a value in the picker,
• Then go back to "Select a Value".

It should disabled to next button in this case.

YuanZhu-apple:
Thanks for catching this!
I confirm it is a bug.

PR #298, opened 2015-06-21
Fix issue #292 Value picker in a non optional form allows going to next step

YuanZhu-apple:
#292

Add a shared method `ORKIsAnswerEmpty(id answer)` to judge if an answer is empty.
Also add a non optional form step in ORKTest to verify this fix.


YuanZhu-apple:
@julientherier Can you review this one?

julientherier:
@YuanZhu-apple I'll be able to review it on Friday, sorry for the delay.

julientherier:
I can confirm the code behaves as expected.
Thanks for this fix, I'd like to use it in my next active task, it will help me.

PR #304, opened 2015-07-01
Reverse family/given name order for zh, ko and ja

weijentu:
When making a pdf file through ORKConsentDocument, family/given also have to be switched its order when the language configuration with a prefix@"zh", @"ko" or @"ja". Besides nameStr "&nbsp;" is not necessary in these languages.

### DIFF
--- a/ResearchKit/Common/ORKHelpers.h
+++ b/ResearchKit/Common/ORKHelpers.h
@@ -165,6 +165,8 @@ NSCalendar *ORKTimeOfDayReferenceCalendar();
 NSDateComponents *ORKTimeOfDayComponentsFromDate(NSDate *date);
 NSDate *ORKTimeOfDayDateFromComponents(NSDateComponents *dateComponents);
 
+BOOL ORKCurrentLocalePresentsFamilyNameFirst();
+
 UIFont *ORKTimeFontForSize(CGFloat size);
 UIFontDescriptor *ORKFontDescriptorForLightStylisticAlternative(UIFontDescriptor *descriptor);
 
@@ -178,17 +180,6 @@ ORKEqualObjects(id o1, id o2) {
 ORK_INLINE BOOL
 ORKEqualFileURLs(NSURL *url1, NSURL *url2) {
     return ORKEqualObjects(url1, url2) || ([url1 isFileURL] && [url2 isFileURL] && [[url1 absoluteString] isEqualToString:[url2 absoluteString]]);
-}
-
-ORK_INLINE BOOL
-ORKCurrentLocalePresentsFamilyNameFirst() {
-    NSString * language = [[[NSLocale preferredLanguages] firstObject] substringToIndex:2];
-    static dispatch_once_t onceToken;
-    static NSArray *familyNameFirstLangs = nil;
-    dispatch_once(&onceToken, ^{
-        familyNameFirstLangs = @[@"zh",@"ko",@"ja"];
-    });
-    return (language != nil) && [familyNameFirstLangs containsObject:language];
 }
 
 ORK_INLINE NSArray *
--- a/ResearchKit/Common/ORKHelpers.h
+++ b/ResearchKit/Common/ORKHelpers.h
@@ -180,6 +180,17 @@ ORKEqualFileURLs(NSURL *url1, NSURL *url2) {
     return ORKEqualObjects(url1, url2) || ([url1 isFileURL] && [url2 isFileURL] && [[url1 absoluteString] isEqualToString:[url2 absoluteString]]);
 }
 
+ORK_INLINE BOOL
+ORKCurrentLocalePresentsFamilyNameFirst() {
+    NSString * language = [[[NSLocale preferredLanguages] firstObject] substringToIndex:2];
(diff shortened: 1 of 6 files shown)


ResearchKit/Consent/ORKConsentSignature.h
```diff
@@ -90,6 +90,13 @@ ORK_CLASS_AVAILABLE
                                     dateFormatString:(nullable NSString *)dateFormatString
                                           identifier:(NSString *)identifier;
 
+/**
+ Returns a boolean to indicate family name order.
+ 
+ Use this factory method to indicate family name first in the languages with a prefix @"zh", @"ko" and @"ja".
+ */
++ (BOOL)currentLocalePresentsFamilyNameFirst;
+
```
YuanZhu-apple: I understand this method should be accessible from both viewController and PDF generator. But I feels this is not API, it should be placed in `ORKHelpers.h` with other internal utility methods.

ResearchKit/Consent/ORKConsentSignatureFormatter.m
```diff
@@ -58,7 +58,11 @@ - (NSString *)HTMLForSignature:(ORKConsentSignature *)signature {
             if (signature.familyName) {
                 [names addObject:signature.familyName];
             }
-            nameStr = [names componentsJoinedByString:@"&nbsp;"];
+            if ([ORKConsentSignature currentLocalePresentsFamilyNameFirst]) {
+                nameStr = @"";
```
YuanZhu-apple: Why assign an empty string here?

ResearchKit/Consent/ORKConsentSignatureFormatter.m
```diff
@@ -58,7 +58,11 @@ - (NSString *)HTMLForSignature:(ORKConsentSignature *)signature {
...
+                names = [[[names reverseObjectEnumerator] allObjects] mutableCopy];
+            }
+            nameStr = [names componentsJoinedByString:nameStr];
```
YuanZhu-apple: Keep `@"&nbsp;"` as the separator.

YuanZhu-apple:
Hey @weijentu, YES, the name order in PDF should follow `currentLocalePresentsFamilyNameFirst `.

ResearchKit/Consent/ORKConsentSignature.h
weijentu: Yeah, I agree with you. At first, I just thought it's all related to ORKConsentSignature. To put the method in ORKHelper is much better.

ResearchKit/Consent/ORKConsentSignatureFormatter.m
weijentu: A family name + given name without a space is very common in the Asian country because it's not mandatory. However, maybe you are right, I think it could be more formal with a space in an official document.

ResearchKit/Consent/ORKConsentSignatureFormatter.m
```diff
@@ -58,7 +59,10 @@ - (NSString *)HTMLForSignature:(ORKConsentSignature *)signature {
             if (signature.familyName) {
                 [names addObject:signature.familyName];
             }
-            nameStr = [names componentsJoinedByString:@"&nbsp;"];
+            if (ORKCurrentLocalePresentsFamilyNameFirst()) {
+                names = [[[names reverseObjectEnumerator] allObjects] mutableCopy];
+            }
+            nameStr = [names componentsJoinedByString:nameStr];
```
YuanZhu-apple: It would be nice to make the separator explicit, just use:
`nameStr = [names componentsJoinedByString:@"&nbsp;"];`

weijentu: Yes, I just updated it :)

ResearchKit/Common/ORKHelpers.h
```diff
@@ -180,6 +180,17 @@ ORKEqualFileURLs(NSURL *url1, NSURL *url2) {
...
+ORKCurrentLocalePresentsFamilyNameFirst() {
+    NSString * language = [[[NSLocale preferredLanguages] firstObject] substringToIndex:2];
+    static dispatch_once_t onceToken;
+    static NSArray *familyNameFirstLangs = nil;
+    dispatch_once(&onceToken, ^{
+        familyNameFirstLangs = @[@"zh",@"ko",@"ja"];
+    });
+    return (language != nil) && [familyNameFirstLangs containsObject:language];
+}
+
```
YuanZhu-apple: As far as I understand, `inline` keyword would actually copy the code to the places where it get referenced. It causes `static dispatch_once_t onceToken` and `static NSArray *familyNameFirstLangs` to be duplicated. Since we want `familyNameFirstLangs` to be only constructed once, do not use `inline` with this function. Do something similar to `NSDateComponents *ORKTimeOfDayComponentsFromDate(NSDate *date);`.

weijentu: Thank you for reminding me! It's a great highlight.

YuanZhu-apple:
Thanks @weijentu, looks good, merged.

Issue #228, opened 2015-05-26
Accessibility hint for capture process should be exposed as property on image capture step


jwe-apple:
The image capture step should probably include an accessibility hint. Since the description of what exactly to do is specific to the particular use case, we should expose this as a property on the image capture step.

brucehappy:
The ``Localizable.strings`` has two strings that have existed since the epoch of the project on github, and that seem to be remnants of a never included Media step. 
```
/* Media step */
"MEDIA_BUTTON_HAVE_IMAGE" = "Tap to capture a different image";
"MEDIA_BUTTON" = "Tap to capture an image";
```
This raises the question, do we need two hints?   One for the state where an image is not captured, and one for when it is already captured?  If so, presumably we need allow the developer to specify both hints via properties.  And, would you like to repurpose the above strings, with different names, for the default hints of the image capture step when the developer does not specify any?

brucehappy:
@jwe-apple Also would like to know if you believe the hint should be on the Capture/Recapture buttons, or if it should be on the preview.  Currently, I've got the hint on the preview (switching it between the two hints depending on the current mode), and I've also made the magic tap on the preview click either the capture or recapture button depending on which is active.

YuanZhu-apple:
@brucehappy 
"MEDIA_BUTTON_HAVE_IMAGE" and "MEDIA_BUTTON" are not being used, we need to clean it up.   

Add a hint for the image template on the top of preview view, so voice over users can understand the aim of the image capture. 

Currently,  it only says "Capture preview, image". Ideally, it would say something like: "Capture preview, right hand, image".

umerkhan-apple:
https://github.com/ResearchKit/ResearchKit/pull/305

brucehappy:
I had worked on this issue, as indicated in my previous comments.  Apologies for not submitting a PR after the response from @YuanZhu-apple .  I do not wish to step on anyone's toes, but I can submit the PR now, if desired.

umerkhan-apple:
@brucehappy I figured you were busy, so I went ahead and started working on the issue. I have updated the PR following your suggestions. Please compare the PR with your version and let me know of any differences or suggestions. Let me know if you feel your version is significantly different and would like me to close my PR?

brucehappy:
Ok, just ping me by replying to the issue in future, if you don't mind.  It could save us from duplicating work.

[...]

PR #305, opened 2015-07-01
Accessibility hint for Image Capture Step

umerkhan-apple:
Added the requirement of an accessibility hint for image capture step. This is only required if the user provides a template image for an overlay.
#228

### DIFF
--- a/ResearchKit/Common/ORKImageCaptureCameraPreviewView.h
+++ b/ResearchKit/Common/ORKImageCaptureCameraPreviewView.h
@@ -41,6 +41,7 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic) AVCaptureVideoOrientation videoOrientation;
 @property (nonatomic, weak, nullable) UIImage *templateImage;
 @property (nonatomic) UIEdgeInsets templateImageInsets;
+@property (nonatomic, weak, nullable) NSString *templateImageAccessibilityHint;
 @property (nonatomic, getter=isTemplateImageHidden) BOOL templateImageHidden;
 @property (nonatomic, weak, nullable) UIImage *capturedImage;
 
--- a/ResearchKit/Common/ORKImageCaptureCameraPreviewView.h
+++ b/ResearchKit/Common/ORKImageCaptureCameraPreviewView.h
@@ -41,7 +41,7 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic) AVCaptureVideoOrientation videoOrientation;
 @property (nonatomic, weak, nullable) UIImage *templateImage;
 @property (nonatomic) UIEdgeInsets templateImageInsets;
-@property (nonatomic, weak, nullable) NSString *templateImageAccessibilityHint;
+@property (nonatomic, copy, nullable) NSString *templateImageAccessibilityHint;
 @property (nonatomic, getter=isTemplateImageHidden) BOOL templateImageHidden;
 @property (nonatomic, weak, nullable) UIImage *capturedImage;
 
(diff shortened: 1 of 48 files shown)


ResearchKit/Common/ORKImageCaptureCameraPreviewView.h
```diff
@@ -41,6 +41,7 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic) AVCaptureVideoOrientation videoOrientation;
 @property (nonatomic, weak, nullable) UIImage *templateImage;
 @property (nonatomic) UIEdgeInsets templateImageInsets;
+@property (nonatomic, weak, nullable) NSString *templateImageAccessibilityHint;
```
YuanZhu-apple: Shouldn't use `copy` instead of `weak`?

samples/ORKCatalog/ORKCatalog.xcodeproj/project.pbxproj
```diff
@@ -208,6 +208,7 @@
 				TargetAttributes = {
 					869230BD1AAA890A00BFE11B = {
 						CreatedOnToolsVersion = 6.2;
+						DevelopmentTeam = 7JYUG8QGJ3;
```
brucehappy: You should undo this commit to ``project.pbxproj``

ResearchKit/Common/ORKImageCaptureStep.m
```diff
@@ -88,4 +88,12 @@ - (ORKPermissionMask)requestedPermissions {
     return mask | ORKPermissionCamera;
 }
 
+- (void)validateParameters{
+    [super validateParameters];
+    
+    if ( self.templateImage != nil && self.templateImageAccessibilityHint == nil) {
```
YuanZhu-apple: Nice to check if (string.length == 0)

ResearchKit/Common/ORKImageCaptureStep.h
```diff
@@ -65,4 +65,12 @@ ORK_CLASS_AVAILABLE
  */
 @property (nonatomic) UIEdgeInsets templateImageInsets;
 
+/**
+ Accessibility hint to be used in Voice Over for `templateImage`.
+ 
+ The accessibility hint provides voice over instructions for the user. When setting the
+ accessibility hint, be sure to clearly state what the 'templateImage' is displaying.
+ */
+@property (nonatomic, strong) NSString *templateImageAccessibilityHint;
```
YuanZhu-apple: Use `copy` instead of `strong` to prevent anybody pass in a mutable NSString.

brucehappy: A template image is NOT required in order to use the ``ORKImageCaptureStep``.  Since the existing accessibility hint is applied to the preview view, I do not believe this should be called ``templateImageAccessibilityHint``.  If you did not supply a template image, you may wish to provide a hint like "Take a picture of your hand" without mentioning alignment.

ResearchKit/Common/ORKImageCaptureStep.m
brucehappy: I would argue against this entire function.   A template image is NOT required in order to use the ``ORKImageCaptureStep`` and the existing accessibility hint is always applied to the preview view, not only when it has a valid template image.

ResearchKit/Common/ORKImageCaptureView.m
```diff
@@ -121,6 +121,7 @@ - (void)setImageCaptureStep:(ORKImageCaptureStep *)imageCaptureStep {
     
     _previewView.templateImage = imageCaptureStep.templateImage;
     _previewView.templateImageInsets = imageCaptureStep.templateImageInsets;
+    _previewView.templateImageAccessibilityHint = imageCaptureStep.templateImageAccessibilityHint;
```
YuanZhu-apple: Is there a reason not to set `_previewView.accessibilityHint` directly?

brucehappy:
The likely reason the tests are failing is because you have not updated ``ORKSerialization.m``

PR #306, opened 2015-07-01
Implementation of issue #228

brucehappy:
This implementation of issue #228 has a separate hint for each of the two different modes of the preview view.  @umerkhan-apple has already submitted #305 for #228.  Feel free to use this PR instead of that PR, or use the code in this PR to help get that PR finished and then close this PR.

### DIFF
--- a/ResearchKit/Common/ORKImageCaptureStep.h
+++ b/ResearchKit/Common/ORKImageCaptureStep.h
@@ -66,13 +66,23 @@ ORK_CLASS_AVAILABLE
 @property (nonatomic) UIEdgeInsets templateImageInsets;
 
 /**
- Accessibility hint of the preview.
+ Accessibility hint of the capture preview.
+ 
+ This property can be used to specify accessible instructions for capturing the image.  The
+ use of this property can assist in situations where the `templateImage` may not be visible
+ to the user.
+ 
+ For example, if you wish to capture an image of the user's right hand, you may use a template
+ image that displays the outline of the right hand.  You may also wish to set this property
+ to a string such as @"Extend your right hand, palm side down, one foot from your device.
+ Tap the Capture Image button, or two finger tap on the preview, to capture an image of your
+ extended right hand."
  */
-@property (nonatomic, strong) NSString *accessibilityInstructions;
+@property (nonatomic, copy) NSString *accessibilityInstructions;
 
 /**
  Accessibility hint of the capture button.
  */
-@property (nonatomic, strong) NSString *accessibilityHint;
+@property (nonatomic, copy) NSString *accessibilityHint;
 
 @end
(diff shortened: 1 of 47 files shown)


YuanZhu-apple:
Don't worry @brucehappy, our test environment caused unit test failure.

umerkhan-apple:
@brucehappy I'm happy with this implementation. I will close my PR and we can follow up on this one. Just going back to @YuanZhu-apple comment from before, do we need two accessibility hints? The captured image may not necessarily always be what the user was instructed to capture. Also, should the "accessibility hint"/instructions be required if a template image is being used? Or should it be required regardless? Or not required at all?

brucehappy:
I'll reply to your question @umerkhan-apple in #228 so people who are only following the issue (and not this PR) can see this discussion on the requirements.

ResearchKit/Common/ORKImageCaptureStep.h
```diff
@@ -65,4 +65,14 @@ ORK_CLASS_AVAILABLE
  */
 @property (nonatomic) UIEdgeInsets templateImageInsets;
 
+/**
+ Accessibility hint of the preview.
+ */
+@property (nonatomic, strong) NSString *accessibilityInstructions;
+
```
umerkhan-apple: I would suggest a more descriptive documentation of `accessibilityInstructions` since it is not part of the Accessibility Programming Guide for iOS. Include an example in the documentation to explicitly indicate how it should be used.

ResearchKit/Common/ORKImageCaptureStep.h
```diff
@@ -65,4 +65,14 @@ ORK_CLASS_AVAILABLE
...
+/**
+ Accessibility hint of the capture button.
+ */
+@property (nonatomic, strong) NSString *accessibilityHint;
+
```
umerkhan-apple: Use `copy` instead of `strong` to prevent anyone from passing in a NSMutableString.

ResearchKit/Common/ORKImageCaptureView.m
```diff
@@ -121,6 +121,10 @@ - (void)setImageCaptureStep:(ORKImageCaptureStep *)imageCaptureStep {
     
     _previewView.templateImage = imageCaptureStep.templateImage;
     _previewView.templateImageInsets = imageCaptureStep.templateImageInsets;
+    _previewView.accessibilityHint = imageCaptureStep.accessibilityInstructions;
+
```
umerkhan-apple: This sets the `accessibilityInstructions` on the previewView even after the image has been captured. Maybe we want to modify `updateAppearances` to not have the accessibility instructions when there is a `capturedImage'?

umerkhan-apple:
r=me.

YuanZhu-apple:
r=me. Thanks @brucehappy !

PR #307, opened 2015-07-02
ORKHTMLPDFWriter generates an extra empty page when useMetric is true

weijentu:
When useMetric is true, ORKHTMLPDFWriter is going to generate an extra
empty page before the signature page. Gaining little more pixels can
omit this extra empty page.

I also upload two output pdf files for quick references. The files are generated via ORKCatalog. Please note the page 3 of 4 is empty in A4 size while useMetric is true.

Example Consent.pdf in letter size
https://db.tt/EAHt9dMT

Example Consent.pdf in A4 size
https://db.tt/WV7LuDmS

### DIFF
--- a/ResearchKit/Common/ORKHTMLPDFWriter.m
+++ b/ResearchKit/Common/ORKHTMLPDFWriter.m
@@ -37,8 +37,8 @@
 #define PPI 72
 #define ORKSizeMakeWithPPI(width, height) CGSizeMake(width * PPI, height * PPI)
 
-static const CGFloat A4Width = 8.26666667;
-static const CGFloat A4Height = 11.6916667;
+static const CGFloat A4Width = 8.27;
+static const CGFloat A4Height = 11.69;
 static const CGFloat LetterWidth = 8.5f;
 static const CGFloat LetterHeight = 11.0f;
 


YuanZhu-apple:
Hey @weijentu can you specify your testing environment? Like XCode version and simulator version or device version. 
I don't see an extra page with my XCode 6.4 in A4 format.

weijentu:
Hi @YuanZhu-apple , I'm also using Xcode 6.4 with iOS 8.4 simulator (12H141) on Yosemite. I just created an A4Test branch in the following link under my forked repository. You can see what changes I've made for this test.
https://github.com/weijentu/ResearchKit/tree/A4Test

Just build and run the ORKCatalog.xcodeproj and then click Consent table view cell to finish all consent steps. After that, the pdf file output path will be printed out through ORKTaskViewControllerDelegate. Just access that path and open the pdf file. Then you probably will be able to see an empty page 3 right away.

YuanZhu-apple:
I tried your branch, still see 3 pages :|
screen shot 2015-07-02 at 3 41 38 pm
@umerkhan-apple Can you give a try when you are not busy?

weijentu:
@YuanZhu-apple I guess I know what's the difference. Did you try simulator for iPhone 4S or iPhone 5? I just tried iPhone 5S & iPhone 6. On 64-bit arch simulator, the result is truly correct with 3 pages. I think it makes sense because the precision of float is different. That also explains why gain resolution little bit can omit the empty page.

YuanZhu-apple:
@weijentu Interesting finding!
I can see the extra page on iPhone5 simulator.

umerkhan-apple:
@YuanZhu-apple I can see the extra page on iPhone5 and iPhone4S simulator.

PR #311, opened 2015-07-06
Fix: Initializing from a metatype must reference 'init' explicitly

danoli3:
Fix for Swift Error in Example.


YuanZhu-apple:
Hey @danoli3, from what I see you just delete a file...

danoli3:
Yeah not sure what happened there. Re-doing this PR.

Issue #310, opened 2015-07-05
UI briefly shows default button titles, although taskViewController: stepViewControllerWillAppear: overrides them


oliverschaefer:
Implementing `taskViewController: stepViewControllerWillAppear:` is the current way of modifying the default behavior of any `ORKStepViewController` (e.g. button titles). This approach has one major downside: When the UI flips to the next step, default button titles (e.g. cancel) remain visible for a quarter of a second.

oliverschaefer:
I'd suggest a new `ORKTaskViewControllerDelegate` method, such as `taskViewController: configureStepViewController:`, which should be called from `ORKTaskViewController`'s `viewControllerForStep:` method

umerkhan-apple:
Currently investigating this issue. Stay tuned, will let you know what I find.

YuanZhu-apple:
@umerkhan-apple Can you comment here how this issue is addressed in the PR?

umerkhan-apple:
Having ``[self setupButtons]`` inside ``viewDidLoad`` loads the default button titles in the navigation bar because the custom buttons have not been set at this point. The custom buttons are set inside ``viewWillAppear:`` when the delegate responds to the ``stepViewControllerWillAppear:``.

YuanZhu-apple:
Fixed in #315

PR #315, opened 2015-07-08
Implementation of issue # 310 - Custom title glitch for bar button items.

umerkhan-apple:
https://github.com/ResearchKit/ResearchKit/issues/310

### DIFF
--- a/ResearchKit/Common/ORKStepViewController.m
+++ b/ResearchKit/Common/ORKStepViewController.m
@@ -140,6 +140,10 @@
     [super viewWillAppear:animated];
     ORK_Log_Debug(@"%@", self);
     
+    // Required here (instead of viewDidLoad) because any custom buttons are set once the delegate responds to the stepViewControllerWillAppear,
+    // otherwise there is a minor visual glitch, where the original buttons are displayed on the UI for a short period. This is not placed after
+    // the delegate responds to the stepViewControllerWillAppear, so that the target from the button's item can be used, if the intention is to
+    // only modify the title of the button.
     [self setupButtons];
     
     if ([self.delegate respondsToSelector:@selector(stepViewControllerWillAppear:)]) {
--- a/ResearchKit/Common/ORKStepViewController.m
+++ b/ResearchKit/Common/ORKStepViewController.m
@@ -99,7 +99,6 @@
 
     self.view.backgroundColor = ORKColor(ORKBackgroundColorKey);
     
-    [self setupButtons];
 }
 
 - (void)setupButtons {
@@ -140,6 +139,7 @@
 - (void)viewWillAppear:(BOOL)animated {
     [super viewWillAppear:animated];
     ORK_Log_Debug(@"%@", self);
+    
     [self setupButtons];
     
     if ([self.delegate respondsToSelector:@selector(stepViewControllerWillAppear:)]) {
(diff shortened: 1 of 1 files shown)


YuanZhu-apple:
@oliverschaefer Can you review this one? Explanation is here: https://github.com/ResearchKit/ResearchKit/issues/310#issuecomment-120160523

oliverschaefer:
I think you need to call setupButtons after the delegate callback in viewWillAppear, not before.

I've noticed a second follow-up issue while reviewing: the learn more button title won't ever change, although the corresponding property is set correctly. Can you please check this?

oliverschaefer:
The custom learnMoreButtonTitle gets considered once you move the stepDidChange call from viewDidLoad to viewWillAppear. Might this have any unwanted side effects?

umerkhan-apple:
@oliverschaefer Moving the `[self setupButtons]` after the delegate callback in `viewWillAppear:` causes the back button to not attach the action specified by the custom button. Though, adding `[self stepDidChange]` after the delegate callback resolves the issue of `learnMoreButtonTitle` not updating. I have not encountered any unwanted side effects as of yet. Im pushing the changes, please review.

oliverschaefer:
@umerkhan-apple As long as `[self setupButtons]` is called before the delegate callback, I still encounter the glitch while navigating?!

umerkhan-apple:
@oliverschaefer I am not able to recreate the glitch after this fix. Could you elaborate as to exactly what you are encountering? 

As for the `learnMoreButtonTitle`, the button had to be reset after the the title property is set. I have updated the PR to reflect these changes. As a result, we no longer need `[self stepDidChange]' after the delegate callback.

oliverschaefer:
@umerkhan-apple Okay, I finally figured out what was going on. My delegate implementation of `stepViewControllerWillAppear` tried to change the title of the `cancelButtonItem` via
`stepViewController.cancelButtonItem?.title = "new title"`. That led to a noticeable glitch when navigating to another stepViewController, even with your fix in place. After re-reading the documentation of the `cancelButtonItem` property, I now use `stepViewController.cancelButtonItem = UIBarButtonItem(title: "new title", style: .Plain, target: stepViewController.cancelButtonItem?.target, action: stepViewController.cancelButtonItem?.action ?? "")`. That way the glitch is gone, so it should be save to merge the PR now.

@umerkhan-apple @YuanZhu-apple @jwe-apple 
However, it is a little confusing to modify titles by changing dedicated properties on the one hand (such as `learnMoreButtonTitle`) and by setting custom button items on the other hand (such as `cancelButtonItem`), but that leads to longer discussion (is it really advisable to provide your own task cancelation logic via custom targets/actions?)

umerkhan-apple:
@oliverschaefer It is really not advisable to provide your own task cancelation logic. The reason for setting the cancelButtonItem with a custom button is to allow the user to pass a nil value. ``When the value of the property is `nil`, the Cancel button is not displayed;`` On the other hand, `learnMoreButton` has a delegate callback to allow for customization.

oliverschaefer:
Okay, I see.

ResearchKit/Common/ORKStepViewController.m
```diff
@@ -140,12 +139,13 @@ - (void)stepDidChange {
 - (void)viewWillAppear:(BOOL)animated {
     [super viewWillAppear:animated];
     ORK_Log_Debug(@"%@", self);
+    
     [self setupButtons];
```
YuanZhu-apple: Can you add some documentation for why this is required at here? In case we forget next time:)

umerkhan-apple: @YuanZhu-apple Good call. PR coming up!

YuanZhu-apple:
r=me.

PR #317, opened 2015-07-09
Fix: Initializing from a metatype must reference 'init' explicitly

danoli3:
Okay this one is finally a correct PR ;3

Fixes:
swift


YuanZhu-apple:
Thanks for this fix!

Issue #323, opened 2015-07-14
Change to logName in ORKRecorder.m removes step identifier from ORKFileResult fileURL 


chrisinsfo:
Re: Commit 49e79ab9f4df264b0085515e12658c8f61a144aa in Stable branch.

Removing the step.identifier from the logName string has removed information from the ORKFileResult fileURL. This has resulted in filenames that cannot be discriminated. For example, instead of deviceMotion_fitness.walk and deviceMotion_fitness.rest as url path components in Apple Cardiology, we now have deviceMotion_deviceMotion. 

As an example, consider ORKOrderedTask / fitnessCheckTaskWithIdentifier(). This method initializes an ORKFitnessStep and sets recorderConfigurations with an identifier such as "pedometer". When ORKRecorder returns its logName which is used to form the url, the first component is [self recorderType] which is "pedometer". The second component of the logName is now self.identifier ("pedometer"). As we can see in this example, we have less information by returning self.identifier than self.step.identifier.

@YuanZhu-apple @rsanchezsaez . Could you please comment on the intent of the change?
@DwayneJengSage @peculiar @jwarmkessel Could you please consider this issue from your perspective?

chrisinsfo:
Further research has yielded this earlier issue: https://github.com/ResearchKit/ResearchKit/issues/2

umerkhan-apple:
@chrisinsfo The intent of the change is to allow unique identifiers for recorders. Using the old implementation with `deviceMotion_fitness.walk`, there would be no way to identify multiple instances of a recorder of that configuration. On the other hand, with the current implementation we get `deviceMotion_uniqueRecorderIdentifer`.

![screen shot 2015-07-14 at 2 10 41 pm](https://cloud.githubusercontent.com/assets/13040996/8685229/27e0358e-2a32-11e5-915c-3b4ee3850a13.png)

I will go ahead and update `logName` to also include the step identifier, so that it would result in `deviceMotion_stepIdentifier_recorderIdentifier`. I believe this should resolve the issue you're encountering with the fileURL.

chrisinsfo:
@umerkhan-apple . Sounds good, thanks.

umerkhan-apple:
After internal discussion, it has been decided that it is not secure to include any identifiers in `logName` since it is used to generate filenames. Instead, the `recorderType` and the `recorderUUID` will be used to generate `logName`.

PR #324, opened 2015-07-14
IOI # 323 - Refactored ORKRecorder's logName

umerkhan-apple:
Added step identifier to ORKRecorder's `logName` to make it more descriptive.
#323

### DIFF
--- a/ResearchKit/ActiveTasks/ORKRecorder.m
+++ b/ResearchKit/ActiveTasks/ORKRecorder.m
@@ -178,7 +178,7 @@
 }
 
 - (NSString *)logName {
-    return [NSString stringWithFormat:@"%@_%@_%@", [self recorderType], self.identifier, self.step.identifier];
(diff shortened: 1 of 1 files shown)


ResearchKit/ActiveTasks/ORKRecorder.m
```diff
@@ -178,7 +178,7 @@ - (NSString *)recorderType {
 }
 
 - (NSString *)logName {
-    return [NSString stringWithFormat:@"%@_%@", [self recorderType],self.identifier];
+    return [NSString stringWithFormat:@"%@_%@_%@", [self recorderType], self.step.identifier, self.identifier];
```
peculiar: Nit: does the order of identifier matter?

chrisinsfo: r=me. We can just pull it out of the URL as a filename and translate it to what Sage requires, or use as-is.

YuanZhu-apple:
r=me

Issue #294, opened 2015-06-15
Step result ignored on non-optional question step


brenbob:
I'm initializing a Choice Question step with a previously-saved step result. This works fine, and the question renders with the appropriate choice selected. 

But when I set the question step as non-optional, the step result is ignored. 

My apologies for lack of code example. I'm working to set up a generic example that repro's the problem. Also can check if this occurs for other question types.

YuanZhu-apple:
Can you make a fork and modify ORKText to reproduce this issue?

brenbob:
Not sure about doing a fork, but I can repro the issue with this basic survey implementation. If I set questionStep.optional to 'true', the correct choice is selected on render. If I set it to 'false', no choice is selected.

// main viewController

class SurveyVC: UIViewController, ORKTaskViewControllerDelegate {
    
    /**
    When a task is completed, the `SurveyViewController` calls this closure
    with the created task.
    */
    var taskResultFinishedCompletionHandler: (ORKResult -> Void)?
    
    override func viewDidAppear(animated: Bool) {
        super.viewDidAppear(animated)
        self.startSurvey()
    }
    
    /**
    Start the survey as a RK task
    */
    func startSurvey() {
        
        let task = surveyTask
        
        let taskViewController = SurveyTaskVC(task: task, taskRunUUID: nil)
        
        // Make sure we receive events from `taskViewController`.
        taskViewController.delegate = self
        
        taskViewController.defaultResultSource = taskViewController

        // Assign a directory to store `taskViewController` output.
        taskViewController.outputDirectory = NSURL(fileURLWithPath: NSSearchPathForDirectoriesInDomains(.DocumentDirectory, .UserDomainMask, true)[0] as! String, isDirectory: true)

       // let vc = self.view.window!.rootViewController
 
       self.presentViewController(taskViewController, animated: true, completion: nil)
        
    }
    
 
    var surveyTask: ORKOrderedTask {
        
        var steps = [AnyObject]()

        // Add a question step.
        var choices = [ORKTextChoice]()
        choices.append(ORKTextChoice(text: "YES", value: "1"))
        choices.append(ORKTextChoice(text: "NO", value: "0"))
        var format = ORKTextChoiceAnswerFormat(style: ORKChoiceAnswerStyle.SingleChoice, textChoices: choices)

        let questionStepTitle = "Are you happy?"
        let questionStep = ORKQuestionStep(identifier: "Q1", title: questionStepTitle, answer: format)
[...]

PR #327, opened 2015-07-20
Fix issue 294, adaptive UI for non-optional single choice step

YuanZhu-apple:
- Show choice cells with nextButton on step has an answer
 - Show arrowed choice cells without next button on step has no answer
#294


YuanZhu-apple:
@jwe-apple Can you review this change?

jwe-apple:
r=me

PR #330, opened 2015-07-23
Enhance argument validation for Predicate Step Navigation Rule

rsanchezsaez:
A user was having trouble using `ORKPredicateStepNavigationRule` because she was passing an array of *steps* (instead of *step identifiers*) to the `destinationStepIdentifiers` argument. I added argument checking to make sure the objects in the arrays are of the correct class. Throws an exception if they aren't.

This PR also slightly rewrites `ORKTest`'s `ORKNavigableOrderedTask` example so it's easier to read.


YuanZhu-apple:
r=me. Thank you @rsanchezsaez !

umerkhan-apple:
r-me.

PR #337, opened 2015-07-29
Fixed bug in ORKNumericAnswerFormat so the value shown is the full decimal

umerkhan-apple:
**Fixed this bug for Decimal Number:**
1. Input: "20.000123"
2. Tap next, then go back
3. Then saved answer is 20 instead of 20.000123


YuanZhu-apple:
r=me.

Issue #290, opened 2015-06-10
Implement size class support


rsanchezsaez:
Replace fixed device metrics by device size class support. Specially relevant in order to properly support iPad multitasking.

jwe-apple:
Yes indeed. On the ios9 branch, there have been specific issues identified which this should aim to fix:
- Image capture task rotating with a split screen can crash <rdar://problem/21191705>
- Image choice step can crash under rotation with split screen <rdar://problem/21169048>
- Scale answer formats lay out particularly badly in split screen <rdar://problem/21168750>

Note that you can't completely replace the fixed device metrics, since even for regular vertical size class, ResearchKit has multiple metric configurations. For the most part, only the metrics that depend on or affect horizontal spacing should really move to use size class.

brucehappy:
@jwe-apple, any chance we might be able to bridge the internal radar tracker with the github issue tracker for ORK?  I'm sure its a tall order.  But if its not possible, then encouraging internal apple folks to use the github tracker for new ORK issues?

jwe-apple:
No chance for a bridge I'm afraid. We generally do this manually, where the issue is suitable for public consumption. Clearly, before yesterday we couldn't have posted these ;-)

brucehappy:
Ok, I certainly understand the timing, and that it'll be a manual process.  Now I guess I just need an iPad Air 2 to test with :smiley:

rsanchezsaez:
Probably this is no longer needed due to the solution implemented by Yuan in [PR #370](https://github.com/ResearchKit/ResearchKit/pull/370).

kelseydedoshka-apple:
@rsanchezsaez closing this issue due to the merged solution you mentioned above!

Issue #110, opened 2015-04-26
Refactor layout code


rsanchezsaez:
Layout code for most of the views is currently a bit over the place. Consider doing some refactoring to make it more homogenous and easier to maintain.

I don't have any specific suggestions right now, but it certainly feels like there is room for simplification.

jwe-apple:
This is a bit of an umbrella issue. Specific tasks should at the least be separate PR's, happy to have multiple PR's reference this.

Things we've been thinking about:
- Further refactor form and step cells; there's still some duplication going on here.
- Simplify the header view. It's common across almost all templates, and its layout code seems more complicated than it needs to be.
- Look at the vertical container view. Consider splitting its implementation. Currently it creates a significant view structure within it and manages that entire structure. Consider giving some control to one or more of its subviews.

rsanchezsaez:
Another item:
 - Try to reuse more code between `ORKVerticalContainerView` and `ORKTableViewContainer` (specifically, on-screen keyboard handling).

umerkhan-apple:
@rsanchezsaez #343 has been merged. Is it okay to close this issue?

Issue #99, opened 2015-04-23
Revise all 'updateConstraints' implementations


rsanchezsaez:
Some `updateConstraints` implementations may be accumulating duplicate constraints. They also need to call `[super updateConstraints]` last, but some of them call it early. Revise this.

See comments [here](https://github.com/ResearchKit/ResearchKit/pull/98) for more details.

jwe-apple:
Yep.

jwe-apple:
While revising some of these for other issues, I remembered why so many of them have been set up dynamic. We thought developers might want to present our view controllers in frames of their own devising, for instance on iPad. So in that case the margins should depend on the presenting frame, rather than the screen size.

rsanchezsaez:
I think that presenting the view controllers on custom sized frames would continue to work with static constraints (as long as the containing frame doesn't change in the middle of a task). So I feel we can try to make the constraints static as much as we can (as long as the constraints correctly support a device orientation change).

The only thing is that all *ork metrics* are now generally gotten based on the *view's window*, which prevents it getting proper metrics for a custom frame. For metric extraction, we should be able to replace the *view's window* by the *task controller root view* without much trouble. Right?

jwe-apple:
@rsanchezsaez That's the point. The size of the task controller's root view is not necessarily known at the time of view instantiation - as we won't be in the view hierarchy at that point. Consider for example instantiation during UI state restoration.

rsanchezsaez:
Ah, I see.

rsanchezsaez:
In progress: I'll homogenize all `updateConstraints` methods and make sure `super` is called last. I'll also try to move constraints to `init` when they don't need to dynamically recreated (as in `ORKActiveStepQuantityView`, [PR #125](https://github.com/ResearchKit/ResearchKit/pull/125)).

I think we can make this work properly when using RK view controllers in custom frames even if we have some static constraints. But let's cross that bridge when we come to it.

jwe-apple:
Yep, go for it.

jwe-apple:
Actually can you keep the `super` move separate from the ->`init` change, and do separate PRs?

rsanchezsaez:
Will do.

Should I do the homogenization for truly dynamic constraints in a third PR? I plan to consistently use `activateConstraints:`/`deactivateConstraints:` instead of `addConstraints:`/`removeConstraints:`, as the former seems more automatic and less error prone.

jwe-apple:
[...]

PR #343, opened 2015-07-30
Constraint code homogenization

rsanchezsaez:
I went over all classes with constraints and homogenized the constraint code (this is a big pull request). This partially deals with [Issue #110](https://github.com/ResearchKit/ResearchKit/issues/110), and is related to the comments in [Issue #99](https://github.com/ResearchKit/ResearchKit/issues/99).

While I was doing this, I tested every individual class immediately after changing it. I tried not to break anything, but a full UI review would be a good idea.

It takes care of the following items (some are purely stylistic, but others slightly affect how the code works):

*Stylistic changes*
- Always use the property format when doing `view.translatesAutoresizingMaskIntoConstraints = NO` instead of the method format (`setTranslatesAutoresizingMaskIntoConstraints:`).
- Align all constraint code so it doesn't overflow the screen (easier to read).
- Use `CGFloat` literals (`1.0` and `0.0`) on `multiplier` and `constant` parameters when creating constraints. This visually indicates that these are `CGFloat`.
- Use the `ORKScreenMetricMaxDimension` constant instead of magic numbers for constraints requesting a very large width or height.
- Use `(NSLayoutFormatOptions)0` instead of just `0` when passing constraint options (makes argument clearer).
- Use `[array count]` method syntax instead of `array.count` property syntax.

*Code changes*
- All constraints are added and removed by activation/deactivation (`NSLayoutConstraint activateConstraints:` and `NSLayoutConstraint deactivateConstraints`). This is the [preferred way of doing it on *iOS 8* and up](https://developer.apple.com/library/ios/documentation/AppKit/Reference/NSLayoutConstraint_Class/#//apple_ref/occ/instp/NSLayoutConstraint/active) as it adds the constraints at the most optimal place.
- Make sure static constraints are created only once, usually in the `-setUpConstraints` method.
- Variable constraints (which need to be recalculated) are always stored in the `_variableConstraints` ivar. These are always removed and recreated in the same way.
- Constraints which need *constant* updating are stored as ivars (instead of in a `NSDictionary`): less boilerplatery and more type-safe.
- Constraints which need *constant* updating are updated on the `updateConstraintsConstantForWindow:` method. This method is usually called: when there are changes in the view; in `-updateConstraints`; and in `-willMoveToWindow:`. The last one is the reason of the window argument.
- Remove all `_screenType` ivars, which was needless artificial state (error prone). Instead, the constraints are updated correctly by passing the appropriate `UIWindow` to `updateConstraintsConstantForWindow:`.
- `ORKImageSelectionView` structure was simplified by dropping the invisible label (the choice label takes the functionality of making sure the `intrinsicContentSize` is tall enough for the tallest string choice).
- Use internal *ivars* instead of internal properties in some classes (for coherence with most of the code).

I may be forgetting other things that I did, I'll update the PR if something comes up. Let me know if you have any question.

Although these changes will be quite a task to review, I think they make constraint code more homogenous, more stable, easier to read, and easier to maintain.


rsanchezsaez:
I couldn't help myself and, taking advantage of this big PR, I refactored the uses of `array.count` property syntax into the more explicit `[array count]` method syntax. I'll add this recommendation to the *Coding Style Guide* as well.

rsanchezsaez:
Hmm, regarding the last commit on this PR: I just realized that according to the documentation `count` changed from a method to a property on *iOS 8* (I guess this was done for coherence with Swift). Should we then favour the `.count` syntax instead on *Objective-C*?

I can easily overwrite the last commit with the opposite style, and update the *Coding Style Guide* accordingly.

rsanchezsaez:
I deleted the `[array count]` commit from this PR to facilitate reviewing. I'll submit it as a different PR, but with the (now officially documented) `array.count` format. (Unless you tell me otherwise).

YuanZhu-apple:
@rsanchezsaez I think we should follow `.count` syntax in this PR.

rsanchezsaez:
Ok! I'll add some property syntax refactoring commits (for `count`, `length`, `*Value`, and a few others) to this PR.

rsanchezsaez:
Added.

YuanZhu-apple:
@rsanchezsaez Thanks you the code refactor!
Do you think by any chance you can address  #290 "Implement size class support" in this PR?

rsanchezsaez:
Yeah, I'll look into that in a few days.

YuanZhu-apple:
@rsanchezsaez I know you are also taking on the **Charts** PR, do you think you can fix #290 by early next week? 

Or I can create a separate PR to address it. Then this one can be merge with my PR.

rsanchezsaez:
It's a bit of a tight schedule, I'm not sure if I'll have time to do it so quick.

Could you possibly give it a stab starting from my *rsanchezsaez-constraints* branch? That way the merge would be super-easy.

---

Note that the `deactivateConstraints:` *iOS 9* bug discussed [in the ORKTowerOfHanoiView crash PR](https://github.com/ResearchKit/ResearchKit/pull/362) will probably affect other parts of this refactoring. Let's hope they fix it at the *iOS* level soon.

YuanZhu-apple:
I like to make my PR base on your branch, but I don't think it would compile under Xcode7(ios9)?
The main purpose of my PR is to make RK looks nicer under multitask split view in iOS9.

rsanchezsaez:
Good point. Then you could make it against the *iOS 9* branch, and I'll figure how to merge it with my constraint refactoring branch later.

rsanchezsaez:
Hey @YuanZhu-apple, do you think is it a good idea to try and merge this early in the release cycle? That way we'll have time to catch and fix any regressions, if there's any.

YuanZhu-apple:
Sure, I will review this.

ResearchKit/ActiveTasks/ORKActiveStepQuantityView.m
```diff
@@ -111,91 +110,89 @@ - (void)setImage:(UIImage *)image {
...
-    NSLayoutConstraint *zeroWidthConstraint = [NSLayoutConstraint constraintWithItem:self
-                                                                           attribute:NSLayoutAttributeWidth
-                                                                           relatedBy:NSLayoutRelationEqual
-                                                                              toItem:nil
-                                                                           attribute:NSLayoutAttributeNotAnAttribute
-                                                                          multiplier:1.0
-                                                                            constant:0.0];
-    zeroWidthConstraint.priority = UILayoutPriorityRequired-1;
-    _zeroWidthConstraint = zeroWidthConstraint;
-    _zeroWidthConstraint.active = !_enabled;
```
YuanZhu-apple: We don't need this line now?

ResearchKit/ActiveTasks/ORKActiveStepQuantityView.m
```diff
@@ -245,12 +242,12 @@ - (instancetype)initWithFrame:(CGRect)frame {
         [self addSubview:_leftView];
         [self addSubview:_rightView];
         [self addSubview:_metricKeyline];
-        [self setupConstraints];
+        [self setUpConstraints];
```
YuanZhu-apple: Will this code make a loop?

`setUpConstraints` -> `setNeedsUpdateConstraints` -> `updateConstraints` -> `setUpConstraints`

ResearchKit/ActiveTasks/ORKDataLogger.m
```diff
@@ -1075,13 +1075,17 @@ - (instancetype)initWithDirectory:(NSURL *)directory delegate:(id<ORKDataLoggerM
     return self;
 }
 
+ORKDefineStringKey(PendingUploadBytesThresholdKey);
+ORKDefineStringKey(TotalBytesThresholdKey);
+ORKDefineStringKey(LoggerConfigurationsKey);
```
YuanZhu-apple: Would this renaming cause a migration on the existing archive?

ResearchKit/ActiveTasks/ORKSpatialSpanMemoryContentView.m
```diff
@@ -274,29 +272,25 @@ - (void)setBounds:(CGRect)bounds {
...
+                                             metrics:nil
+                                               views:views]];
+    NSLayoutConstraint *gameViewHeightConstraint = [NSLayoutConstraint constraintWithItem:_gameView
+                                                                      attribute:NSLayoutAttributeHeight
+                                                                      relatedBy:NSLayoutRelationEqual
+                                                                         toItem:nil
+                                                                      attribute:NSLayoutAttributeNotAnAttribute
+                                                                     multiplier:1.0
+                                                                       constant:ORKScreenMetricMaxDimension];
+    gameViewHeightConstraint.priority = UILayoutPriorityDefaultLow-1;
```
YuanZhu-apple: Not your code, add spaces around `-`

ResearchKit/Common/ORKImageCaptureStepViewController.m
```diff
@@ -109,26 +121,26 @@ - (void)retakePressed:(void (^ __nullable)())handler {
...
 }
 
 - (void)videoOrientationDidChange:(AVCaptureVideoOrientation)videoOrientation {
     // Keep the output orientation in sync with the input orientation
-    ((AVCaptureConnection *)self.stillImageOutput.connections[0]).videoOrientation = videoOrientation;
+    ((AVCaptureConnection *)_stillImageOutput.connections[0]).videoOrientation = videoOrientation;
 }
 
-- (void)queue_captureImageFromData:(CMSampleBufferRef)imageDataSampleBuffer withHandler:(void (^ __nullable)(BOOL))handler {
+- (void)queueCaptureImageFromData:(CMSampleBufferRef)imageDataSampleBuffer handler:(void (^)(BOOL))handler {
```
YuanZhu-apple: Can you leave the `_` in the middle, I find it is easy to read.

ResearchKit/Common/ORKSurveyAnswerCell.m
```diff
@@ -143,11 +143,11 @@ - (void)keyboardWillAppear:(NSNotification *)aNotification {
     _cachedContentInsets = tableView.contentInset;
     _cachedScrollIndicatorInsets = tableView.scrollIndicatorInsets;
     
-    NSDictionary *info = [aNotification userInfo];
-    CGSize kbSize = [info[UIKeyboardFrameEndUserInfoKey] CGRectValue].size;
-    kbSize.height = kbSize.height-44;
+    NSDictionary *userInfo = aNotification.userInfo;
+    CGSize keyboardSize = ((NSValue *)userInfo[UIKeyboardFrameEndUserInfoKey]).CGRectValue.size;
+    keyboardSize.height = keyboardSize.height-44;
```
YuanZhu-apple: Space around `-`

ResearchKit/ActiveTasks/ORKActiveStepQuantityView.m
rsanchezsaez: This `setUpConstraints` call is on the `init` method, and `updateConstraints` doesn't call `setUpConstraints`. I don't think there's an infinite loop, could you re-check?

ResearchKit/Common/ORKImageCaptureStepViewController.m
rsanchezsaez: Are you sure? Isn't this very much against the [Coding Guidelines for Cocoa](https://developer.apple.com/library/mac/documentation/Cocoa/Conceptual/CodingGuidelines/Articles/NamingMethods.html#//apple_ref/doc/uid/20001282-BCIGIJJF)?

ResearchKit/ActiveTasks/ORKDataLogger.m
rsanchezsaez: Good catch, I'm now reusing the old keys for backwards compatibility.

rsanchezsaez:
I have taken care of some of the comment, replied to others, and added some extra *Coding Style Guide* conformance commits.

ResearchKit/ActiveTasks/ORKActiveStepQuantityView.m
rsanchezsaez: Activation of the `_zeroWidthConstraint` constraint is taken care in `updateConstraints`, which will take place at the most optimal time after `setNeedsUpdateConstraints` is called.

rsanchezsaez:
I have added a bunch of *Coding Style Guide* homogenization commits.

ResearchKit/ActiveTasks/ORKActiveStepQuantityView.m
YuanZhu-apple: Nevermind, my bad, I thought this is from `updateConstraints`.

ResearchKit/Common/ORKImageCaptureStepViewController.m
YuanZhu-apple: I just think this falls into a special use case, to me and @umerkhan-apple,  `_` is more visible.

rsanchezsaez: Sure thing, done. It makes kind of sense for methods that are only called from a *GCD* queue.

YuanZhu-apple:
I found a issue which might not from this PR.


1. Launch ORKTest app in iPad simulator in landscape mode
2. Rotate simulator to Portrait mode
3. All the items form into one column, which is unexpected.

screen shot 2015-09-22 at 5 08 07 pm

rsanchezsaez:
Yeah, that's an unrelated problem. I have submitted a separate [PR](https://github.com/ResearchKit/ResearchKit/issues/474) with a fix.

YuanZhu-apple:
r=me. Thank you @rsanchezsaez !

rsanchezsaez:
Excellent! Let me know if any UI issue comes up in the next full UI review.

Issue #349, opened 2015-08-04
Tower of Hanoi should honor tintColor


rsanchezsaez:
The *Tower of Hanoi Task* should honor the `tintColor` in the same way the *Memory Game Task* does. See screenshots below.

*Note* that after tapping the disks, they get the proper `tintColor`.

---

![ios simulator screen shot 4 aug 2015 15 08 14](https://cloud.githubusercontent.com/assets/444313/9061246/cdd6d428-3aba-11e5-87a7-2e45e00df7a6.png)

---

![ios simulator screen shot 4 aug 2015 15 08 05](https://cloud.githubusercontent.com/assets/444313/9061247/ce0032f0-3aba-11e5-80ae-76a095705dee.png)

---

YuanZhu-apple:
Fix has been merged.

PR #350, opened 2015-08-04
Fix small Tower of Hanoi issues

rsanchezsaez:
This fixes two small Tower of Hanoi issues:
- ORKTest serialization unit test was failing due improper `NSArray` unarchiving.
- The discs were not honoring the `tintColor` on first display ([Issue #349](https://github.com/ResearchKit/ResearchKit/issues/349)).


YuanZhu-apple:
Thanks! 
r=me.

PR #359, opened 2015-08-06
Review Step: API Design

oliverschaefer:
See ORKReviewStep.md


YuanZhu-apple:
Hi @oliverschaefer Can add a .md file with the content in your draft design?

oliverschaefer:
Hi @YuanZhu-apple, I've added that md file and wrote some code today just to test some aspects of the draft

ResearchKit/Common/ORKReviewStep.md
```diff
@@ -0,0 +1,81 @@
...
+  - step results can be changed
+  - review direction defaults to reverse			
+
+<h2>ORKReviewStepViewControllerDelegate</h2>
+<h3>methods</h3>
+    reviewStepViewController: canReviewStep: 
+        called to determine if a step can be reviewed
+    
+    reviewStepViewController: canChangeStep: 
+        called to determine if an answer can be changed during review
```
YuanZhu-apple: What is the usage of  `canReviewStep` and `canChangeStep`?

ResearchKit/Common/ORKReviewStep.md
```diff
@@ -0,0 +1,81 @@
...
+
+    tableView: didSelectRowAtIndexPath:
+        if reviewStepViewController: canChangeStep: returns true, call reviewStepViewController: reviewStep:
+
+<h2>ORKTaskViewController</h2>
+<h3>properties</h3>
+    NSArray *navigationStack
+    
+<h3>methods</h3>
+    pushStepOnNavigationStack:
```
YuanZhu-apple: No need to maintain a stack. Just use `_managedStepIdentifiers` and `stepForIdentifer:` to recover the stack.

ResearchKit/Common/ORKReviewStep.md
```diff
@@ -0,0 +1,81 @@
...
+<h3>properties</h3>
+    NSArray *steps
+    ORKTaskResultSource *resultSource
+    ORKReviewStepReviewDirection *reviewDirection
+<h3>use cases</h3>
+- __outside of a survey = standalone__
+  - steps array is provided upon initialization
+  - review step is at the beginning
+  - step results may not be changed
+  - review direction defaults to forward
```
YuanZhu-apple: Can you explain a bit of the review direction ?

ResearchKit/Common/ORKReviewStep.md
```diff
@@ -0,0 +1,81 @@
...
+    ORKReviewStepReviewDirection reviewDirection
+<h3>methods</h3>
+    discoverSteps: withResult:
+            repeatedly call stepAfterStep: (reviewDirection forward) or stepBeforeStep: (reviewDirection reverse) until upcoming step is nil or another review step is reached
+            for every step in steps array
+                call reviewStepViewController: shouldIncludeStep:
+            reload tableView data			
+
+    tableView: didSelectRowAtIndexPath:
+        if reviewStepViewController: canChangeStep: returns true, call reviewStepViewController: reviewStep:
```
YuanZhu-apple: Should always go the the stepVC, give user a complete view. Instead, make the StepVC immutable base on the need.

ResearchKit/Common/ORKReviewStep.md
```diff
@@ -0,0 +1,81 @@
...
+
+<h2>ORKReviewStepViewController</h2>
+<h3>properties</h3>
+    NSArray *steps
+    ORKTaskResultSource *resultSource
+    ORKReviewStepViewControllerDelegate *reviewDelegate
+    ORKReviewStepReviewDirection reviewDirection
+<h3>methods</h3>
+    discoverSteps: withResult:
+            repeatedly call stepAfterStep: (reviewDirection forward) or stepBeforeStep: (reviewDirection reverse) until upcoming step is nil or another review step is reached
```
YuanZhu-apple: if `ORKReviewStepViewController` has all the steps, then the sequence is in the steps array, right?

oliverschaefer: Exactly

ResearchKit/Common/ORKReviewStep.md
oliverschaefer: Okay, that will require some changes to `ORKStepViewController` as well. I'd suggest a `canChangeStepResult` property. If set to `YES` the behavior is just as it is now. If set to `NO` we need to disable any UI elements that may change the result. Furthermore, `ORKTaskViewController` should not add the result of an immutable step to its' `managedResults` array.

ResearchKit/Common/ORKReviewStep.md
oliverschaefer: We won't need `canChangeStep` anymore, if we add a similar property to `ORKStepViewController` (see my comment below).

`canReviewStep` asks the delegate if a step that was discovered by the `ORKReviewStepViewController` should actually be presented for review. This could be used to just display steps that have not been answered yet. To do so an app would need to subclass `ORKTaskViewController` and override the default implementation of `canReviewStep`.

ResearchKit/Common/ORKReviewStep.md
oliverschaefer: The `reviewDirection` is used by `ORKReviewStepViewController` to discover the steps that need to be reviewed. Within a survey, the review step is usually displayed after the steps that need to be reviewed. To discover those, you need to go in a reverse direction and call `stepBeforeStep:withResult` repeatedly. If an app wants to do different and display a review step at the beginning, the review direction needs to be changed to forward. That way steps will be discovered by calling `stepAfterStep:withResult:` instead.

If `ORKReviewStep` is used standalone, steps won't need to be discovered. So `reviewDirection` will be ignored.

ResearchKit/Common/ORKReviewStep.md
oliverschaefer: Okay. But calling `stepWithIdentifier` will only work for steps whose task property is set to a task which implements this method. If the review step is used standalone, that won't be the case. To solve this problem, we would need to change `ORKOrderedTask`s default implementation of `stepWithIdentifier:`, wouldn't we?

ResearchKit/Common/ORKReviewStep.md
YuanZhu-apple: Agree

ResearchKit/Common/ORKReviewStep.md
YuanZhu-apple: If the review step is used standalone, the steps stack has to be on `review step` already.

ResearchKit/Common/ORKReviewStep.md
YuanZhu-apple: I just think the steps are provided by `TaskVC`, so `ORKReviewStepViewController` don't need to discover anything.

oliverschaefer: Yep, you are right. `reviewDirection` should be a property of `ORKReviewStep` instead. The discovery should be performed by the `taskViewController`

oliverschaefer:
Hi @YuanZhu-apple  I've implemented most of what we talked about last week. I thought we should look at some code to move on. That's why I've created a new repo at https://github.com/oliverschaefer/ReviewStepTest and added you as an collaborator. I've also added `ORKTest`examples. Please check out my changes and take a look. Would be nice if we could work on the code together to finally get this one ready.

oliverschaefer:
Hey @YuanZhu-apple, did you get a chance to look at the code mentioned above? I've spent the weekend with some refinements and bug fixes. I've just pushed these changes.

YuanZhu-apple:
I will review it today.

ResearchKit/Common/ORKStep.h
```diff
@@ -167,6 +166,8 @@ ORK_CLASS_AVAILABLE
  */
 - (void)validateParameters;
 
+@property (nonatomic) BOOL isBeingReviewed;
```
YuanZhu-apple: I don't think any state should be on a step (model should not be touched)

Testing/ORKTest/ORKTest/MainViewController.m
```diff
@@ -2162,6 +2183,50 @@ - (IBAction)showCustomNavigationItemTask:(id)sender {
...
+    ORKQuestionStep *step1 = [[ORKQuestionStep alloc] initWithIdentifier:@"reviewStepStandaloneTask.step1"];
+    step1.title = @"step 1 title";
+    step1.text = @"step 1 text";
+    step1.answerFormat = [ORKAnswerFormat booleanAnswerFormat];
+    ORKQuestionStep *step2 = [[ORKQuestionStep alloc] initWithIdentifier:@"reviewStepStandaloneTask.step2"];
+    step2.title = @"step 2 title";
+    step2.text = @"step 2 text";
+    step2.answerFormat = [ORKAnswerFormat booleanAnswerFormat];
+    ORKReviewStep *reviewStep = [[ORKReviewStep alloc] initWithIdentifier:@"reviewStepStandaloneTask.reviewStep"
+                                                                    steps:@[step1, step2]
```
YuanZhu-apple: In this scenario, shouldn't a nil array be provided? 
E.g. the steps can be dynamic.

ResearchKit/Common/ORKTaskViewController.m
```diff
@@ -956,6 +971,16 @@ - (BOOL)shouldPresentStep:(ORKStep *)step {
     return shouldPresent;
 }
 
+- (BOOL)shouldReviewStep:(ORKStep *)step {
+    BOOL shouldReview = ([step isKindOfClass:[ORKQuestionStep class]] || [step isKindOfClass:[ORKFormItem class]]);
```
YuanZhu-apple: Should be `[step isKindOfClass:[ORKFormStep class]]`

ResearchKit/Common/ORKTaskViewController.m
```diff
@@ -956,6 +971,16 @@ - (BOOL)shouldPresentStep:(ORKStep *)step {
...
+    
+    if (shouldReview && [self.delegate respondsToSelector:@selector(taskViewController:shouldReviewStep:)]) {
+        shouldReview = [self.delegate taskViewController:self shouldReviewStep:step];
```
YuanZhu-apple: I am not total convinced to add this delegate method.
Can you give a use case when a developer want to exclude a step from review? 
If this delegate has a good use, then it would be nice to include the actual reviewStep as a parameter (multiple review step in a task?)

ResearchKit/Common/ORKTaskViewController.m
```diff
@@ -1021,6 +1042,39 @@ - (ORKStepViewController *)viewControllerForStep:(ORKStep *)step {
...
+            if (reviewStepViewController.reviewStep.isStandalone) {
+                for (ORKStep *step in reviewStepViewController.reviewStep.steps) {
+                    if ([self shouldReviewStep:step]) {
+                        [steps addObject:step];
+                    }
+                }
+            } else {
+                ORKStep *nextStep = step;
+                do {
+                    switch (reviewStepViewController.reviewStep.reviewDirection) {
```
YuanZhu-apple: Can you use `_managedStepIdentifiers` to gather the steps? instead of using  `stepAfterStep` and `stepBeforeStep`.

ResearchKit/Common/ORKStepViewController.m
```diff
@@ -284,6 +284,18 @@ - (BOOL)hasBeenPresented {
...
+- (BOOL)canChangeStepResult {
+    if (!self.step || !self.taskViewController) {
+        return NO;
+    }
+    BOOL isReviewStepStandalone = NO;
+    ORKReviewStep *reviewStep = [self.taskViewController reviewStepForStep:self.step];
+    if (reviewStep) {
+        isReviewStepStandalone = reviewStep.isStandalone;
+    }
+    return self.step.isBeingReviewed ? !isReviewStepStandalone : YES;
```
YuanZhu-apple: Is it easier to let taskVC is figure out `canChangeStepResult` and `isBeingReviewed` and assign them to the stepVC?

YuanZhu-apple:
@oliverschaefer Thanks for doing this! Just realized you almost make it working.

Testing/ORKTest/ORKTest/MainViewController.m
oliverschaefer: I think there are 2 szenarions: (1) review step within survey = steps are provided by the task view controller (array is nil) (2) review step standalone = steps are provided upon initialization (array is non nil)

ResearchKit/Common/ORKTaskViewController.m
oliverschaefer: Yep, my mistake.

ResearchKit/Common/ORKTaskViewController.m
oliverschaefer: I actually do have a good use case: Use it if you want to filter the steps by some criterion (e.g. only optional steps). One could provide the review step view controller with a segmented control as a custom title view. Based on the selected segment only some of the steps get displayed, the others get filtered out. I think that adding the actual reviewStep is a nice idea for all the delegate methods, not just this one.

ResearchKit/Common/ORKStepViewController.m
oliverschaefer: The taskVC already assigns `isBeingReviewed` to the step. I will change that property to belong to the stepVC as you suggested above.

If we let the taskVC assign `canChangeStepResult` as well, we should take care of proper state restoration. I think it can be done but won't make things a lot easier. However, the present approach might be a little easier to read for anyone who's trying to understand what's going on.

ResearchKit/Common/ORKStep.h
oliverschaefer: You are right. I will change that (see my comment below)

ResearchKit/Common/ORKTaskViewController.m
oliverschaefer: That has 2 major downsides: (1) Since `_managedStepIdentifiers` only contains the identifiers, we need to call `stepWithIdentifier:` to get the actual step. That will only work if `stepWithIdentifier:` is implemented. (2) If you want to start a task with a review step, `managedStepIdentifiers` only contains the review step but none of the other steps that should be reviewed. I would keep the current implementation instead.

oliverschaefer:
Hey @YuanZhu-apple I've just realized you probably reviewed an outdated diff. I've pushed my latest changes (including everything mentioned above) to my forked repository as well as to https://github.com/oliverschaefer/ReviewStepTest. Sorry for that.

Testing/ORKTest/ORKTest/MainViewController.m
YuanZhu-apple: Sorry, my bad. Off course, this is standalone case.

ResearchKit/Common/ORKAnswerFormat.h
```diff
@@ -560,6 +560,9 @@ ORK_CLASS_AVAILABLE
 ORK_CLASS_AVAILABLE
 @interface ORKBooleanAnswerFormat : ORKAnswerFormat
 
+@property (nullable) NSString *yesText;
```
YuanZhu-apple: No need to add this. They should use ORKTextChoiceAnswerFormat in this case.

ResearchKit/Common/ORKReviewStep.h
```diff
@@ -0,0 +1,62 @@
...
+ORK_CLASS_AVAILABLE
+@interface ORKReviewStep : ORKStep
+
+@property (nonatomic) ORKReviewStepReviewDirection reviewDirection;
+
+@property (nonatomic, readonly, nullable) NSArray *steps;
+
+@property (nonatomic, readonly, nullable) id<ORKTaskResultSource> resultSource;
+
+@property (nonatomic, readonly) BOOL isStandalone;
```
YuanZhu-apple: Can you move this property to internal header, less noise for developers.

ResearchKit/Common/ORKReviewStep.m
```diff
@@ -0,0 +1,84 @@
...
+
+- (void)encodeWithCoder:(NSCoder *)aCoder {
+    [super encodeWithCoder:aCoder];
+    ORK_ENCODE_OBJ(aCoder, steps);
+    ORK_ENCODE_INTEGER(aCoder, reviewDirection);
+}
+
+- (BOOL)isEqual:(id)object {
+    __typeof(self) castObject = object;
+    return [super isEqual:object] && ORKEqualObjects(self.steps, castObject.steps) && ORKEqualObjects(self.resultSource, castObject.resultSource) && self.reviewDirection == castObject.reviewDirection;
```
YuanZhu-apple: Wrap the line.

ResearchKit/Common/ORKReviewStepViewController.h
```diff
@@ -0,0 +1,73 @@
...
+ORK_CLASS_AVAILABLE
+@interface ORKReviewStepViewController : ORKStepViewController
+
+@property (nonatomic, nullable) id<ORKReviewStepViewControllerDelegate> reviewDelegate;
+
+@property (nonatomic, nullable) NSArray *steps;
+
+@property (nonatomic, nullable) id<ORKTaskResultSource> resultSource;
+
+@property (nonatomic, nullable) NSString *noStepsAvailableText;
```
YuanZhu-apple: Should we just use the title and text from ORKReviewStep?

ResearchKit/Common/ORKReviewStepViewController.h
```diff
@@ -0,0 +1,73 @@
...
+                shouldReviewStep:(ORKStep*)step;
+
+- (void)reviewStepViewController:(ORKReviewStepViewController*)reviewStepViewController
+                      willReviewStep:(ORKStep*)step;
+@end
+
+ORK_CLASS_AVAILABLE
+@interface ORKReviewStepViewController : ORKStepViewController
+
+@property (nonatomic, nullable) id<ORKReviewStepViewControllerDelegate> reviewDelegate;
```
YuanZhu-apple: I'd like to have a init method to take:
1. reviewStep
2. steps (nullable)
3. resultSource (nullable)

ResearchKit/Common/ORKReviewStepViewController.h
```diff
@@ -0,0 +1,73 @@
...
+
+
+NS_ASSUME_NONNULL_BEGIN
+
+@class ORKReviewStepViewController;
+
+@protocol ORKReviewStepViewControllerDelegate <NSObject>
+
+- (NSArray*)reviewStepViewController:(ORKReviewStepViewController*)reviewStepViewController
+                  stepsForReviewStep:(ORKReviewStep*)reviewStep;
```
YuanZhu-apple: Aren't the steps supplied during initialization?

ResearchKit/Common/ORKTaskViewController.m
```diff
@@ -970,7 +989,7 @@ - (ORKStep *)nextStep {
 - (ORKStep *)prevStep {
     ORKStep *step = nil;
     
-    if ([self.task respondsToSelector:@selector(stepBeforeStep:withResult:)]) {
+    if (step == nil && [self.task respondsToSelector:@selector(stepBeforeStep:withResult:)]) {
```
YuanZhu-apple: step is always `nil` in this case, right? Why bother to check it?

ResearchKit/Common/ORKTaskViewController.m
```diff
@@ -981,10 +1000,6 @@ - (void)collectToolbarItemsFromViewController:(UIViewController *)viewController
     _pageViewController.toolbarItems = viewController.toolbarItems;
     _pageViewController.navigationItem.leftBarButtonItem = viewController.navigationItem.leftBarButtonItem;
     _pageViewController.navigationItem.rightBarButtonItem = viewController.navigationItem.rightBarButtonItem;
-    if (![self shouldDisplayProgressLabel]) {
-        _pageViewController.navigationItem.title = viewController.navigationItem.title;
-        _pageViewController.navigationItem.titleView = viewController.navigationItem.titleView;
-    }
```
YuanZhu-apple: Why move the code? I think we should make sure `viewController.navigationItem.title` changing can trigger `collectToolbarItemsFromViewController`

ResearchKit/Common/ORKTaskViewController.m
```diff
@@ -1021,6 +1036,11 @@ - (ORKStepViewController *)viewControllerForStep:(ORKStep *)step {
         
         stepViewController.restorationIdentifier = step.identifier;
         stepViewController.restorationClass = stepViewControllerClass;
+        if ([step isKindOfClass:[ORKReviewStep class]] && [stepViewController isKindOfClass:[ORKReviewStepViewController class]]) {
+            ORKReviewStepViewController *reviewStepViewController = (ORKReviewStepViewController *) stepViewController;
+            reviewStepViewController.resultSource = self.result;
```
YuanZhu-apple: Aren't supposed to only supply `resultSource` when ORKReviewStep is not carrying one?

ResearchKit/Common/ORKTaskViewController.m
```diff
@@ -1050,6 +1069,29 @@ - (BOOL)shouldDisplayProgressLabel {
     return self.showsProgressInNavigationBar && [_task respondsToSelector:@selector(progressOfCurrentStep:withResult:)];
 }
 
+- (nullable ORKReviewStep*)reviewStepForStep:(ORKStep *)step {
```
YuanZhu-apple: I think just create a `lastReviewStep` would be sufficient. Or we can just save the reviewStep on ORKStepVC to save the time of searching.

ResearchKit/Common/ORKTaskViewController.m
```diff
@@ -1050,6 +1069,29 @@ - (BOOL)shouldDisplayProgressLabel {
...
+    }];
+    return [reviewStep isKindOfClass:[ORKReviewStep class]] && reviewStep != step ? (ORKReviewStep*) reviewStep : nil;
+}
+
+- (BOOL)isReviewStepCompleted:(ORKReviewStep*)reviewStep {
+    if (!reviewStep) {
+        return NO;
+    }
+    ORKStepResult *stepResult = _managedResults[reviewStep.identifier];
+    if (stepResult.endDate) {
```
YuanZhu-apple: You might get a non-nil endDate if the user revisits a reviewStep.

ResearchKit/Common/ORKTaskViewController.m
```diff
@@ -1207,6 +1257,8 @@ - (void)stepViewControllerResultDidChange:(ORKStepViewController *)stepViewContr
     if ([strongDelegate respondsToSelector:@selector(taskViewController:didChangeResult:)]) {
         [strongDelegate taskViewController:self didChangeResult: [self result]];
     }
+    
+    [self updateCustomProgressLabel];
```
YuanZhu-apple: Why add this here?

ResearchKit/Common/ORKTaskViewController.m
```diff
@@ -1245,6 +1304,51 @@ - (ORKStep *)stepAfterStep:(ORKStep *)step {
...
 }
 
+#pragma mark - ORKReviewStepViewControllerDelegate
+
+- (nonnull NSArray *)reviewStepViewController:(ORKReviewStepViewController *)reviewStepViewController
+                           stepsForReviewStep:(ORKReviewStep *)reviewStep {
+    NSMutableArray *steps = [[NSMutableArray alloc] init];
+    if (reviewStep.isStandalone) {
+        for (ORKStep *step in reviewStep.steps) {
+            if ([reviewStepViewController.reviewDelegate respondsToSelector:@selector(reviewStepViewController:shouldReviewStep:)] && [reviewStepViewController.reviewDelegate reviewStepViewController:reviewStepViewController shouldReviewStep:step]) {
```
YuanZhu-apple: I don't like the idea to call another delegate method from one delegate method

ResearchKit/Common/ORKTaskViewController.m
```diff
@@ -1245,6 +1304,51 @@ - (ORKStep *)stepAfterStep:(ORKStep *)step {
...
+                }
+            }
+        } while (nextStep || [nextStep isKindOfClass:[ORKReviewStep class]]);
+    }
+    return (reviewStep.reviewDirection == ORKReviewStepReviewDirectionReverse) ? [[steps reverseObjectEnumerator] allObjects] : [steps copy];
+}
+
+- (BOOL)reviewStepViewController:(ORKReviewStepViewController *)reviewStepViewController
+                shouldReviewStep:(ORKStep *)step {
+    return ![step isKindOfClass:[ORKReviewStep class]] || ![step isKindOfClass:[ORKActiveStep class]];
```
YuanZhu-apple: 1. only questionStep and formStep are allowed
2. should expose this `shouldReviewStep:` to taskVC delegate to opt out any step the developer don't want to include.

ResearchKit/Common/ORKTaskViewController.m
```diff
@@ -1245,6 +1304,51 @@ - (ORKStep *)stepAfterStep:(ORKStep *)step {
...
+}
+
+- (void)reviewStepViewController:(ORKReviewStepViewController *)reviewStepViewController
+                      willReviewStep:(ORKStep *)step {
+    ORKStepViewController *stepViewController = [self viewControllerForStep:step];
+    NSAssert(stepViewController != nil, @"A non-nil step should always generate a step view controller");
```
YuanZhu-apple: I think here is a good chance to set `isBeingReviewed` and `canChangeResult`. 
Maybe store `reviewStep` on `ORKStepViewController` at here as well.

ResearchKit/Common/ORKReviewStep.m
```diff
@@ -0,0 +1,84 @@
...
+@implementation ORKReviewStep
+
+- (nonnull instancetype)initWithIdentifier:(nonnull NSString *)identifier
+                                     steps:(nullable NSArray *)steps
+                              resultSource:(nullable id<ORKTaskResultSource>)resultSource {
+    self = [super initWithIdentifier:identifier];
+    if (self) {
+        _steps = steps;
+        _resultSource = resultSource;
+        _reviewDirection = self.isStandalone ? ORKReviewStepReviewDirectionForward : ORKReviewStepReviewDirectionReverse;
```
YuanZhu-apple: Why the `ReviewDirection` is decide by `isStandalone` ? Shouldn't the sequence just be the order of steps array?

oliverschaefer:
Hi @YuanZhu-apple I will be on vacation in the next 3 weeks touring California. I will continue to work on this once I return in mid-September.

YuanZhu-apple:
@oliverschaefer Have a nice trip!
Let me know if you are traveling near by Cupertino CA, maybe we can meet:)

oliverschaefer:
Hi @YuanZhu-apple we are visiting Cupertino tomorrow afternoon. Would be nice to meet you :-)

YuanZhu-apple:
@oliverschaefer We are at 20705 Valley Green Drive, when do like to meet? Will 2pm work for you?

oliverschaefer:
@YuanZhu-apple Great. I'm looking forward to seeing you. 2pm is perfect. In case something goes wrong, this is my mobile number: +1 (415) 960-4806

oliverschaefer:
@YuanZhu-apple @umerkhan-apple  Hi guys, thanks again for meeting. It was a great experience. Now I'm back at home and finally got to work on the review step again. Since there have been so many commits in the last 4 weeks, I've completely refactored my existing code so it can be merged again. To avoid confusion, I close this PR and open a new one shortly.

PR #360, opened 2015-08-06
ORKTest: add form step to Navigable Ordered Task example

rsanchezsaez:
This PR makes `ORKTest`'s *Navigable Ordered Task* example more complete by adding a form step and a predicate rule that matches both form step results.


YuanZhu-apple:
r=me.

umerkhan-apple:
r=m3

PR #361, opened 2015-08-06
Fix a crash case in HanoiTowerView for iOS9

YuanZhu-apple:
The crash is caused by `[NSLayoutConstraint deactivateConstraints:_currentConstraints];`


YuanZhu-apple:
Against wrong branch. I will reopen another one.

PR #362, opened 2015-08-06
Fix a crash case in HanoiTowerView for iOS9

YuanZhu-apple:
Caused by `[NSLayoutConstraint deactivateConstraints:_currentConstraints];`

### DIFF
--- a/ResearchKit/ActiveTasks/ORKTowerOfHanoiTowerView.m
+++ b/ResearchKit/ActiveTasks/ORKTowerOfHanoiTowerView.m
@@ -66,7 +66,8 @@ static const CGFloat kBaseSpacing = 10;
 
 - (void)updateConstraints {
     if (_currentConstraints != nil) {
-        [NSLayoutConstraint deactivateConstraints:_currentConstraints];
+        [self removeConstraints:_currentConstraints];
+        _currentConstraints = nil;
     }
     NSMutableArray *newConstraints = [NSMutableArray new];
     CGFloat height = (kDiskHeight * _maximumNumberOfDisks) + (kDiskSpacing * _maximumNumberOfDisks);


ResearchKit/ActiveTasks/ORKTowerOfHanoiTowerView.m
```diff
@@ -66,7 +66,8 @@ - (instancetype)initWithFrame:(CGRect)frame maximumNumberOfDisks:(NSUInteger)max
 
 - (void)updateConstraints {
     if (_currentConstraints != nil) {
-        [NSLayoutConstraint deactivateConstraints:_currentConstraints];
+        [self removeConstraints:_currentConstraints];
```
rsanchezsaez: I reproduced the crash but I don't see a good reason for it happening. Doesn't it actually look like an *iOS 9 API* bug that should be fixed at the *iOS* level?

This fix goes against my constraint refactor (activate/deactivate is the recommended way of adding/removing constraints [by Apple here](https://developer.apple.com/library/ios/documentation/AppKit/Reference/NSLayoutConstraint_Class/#//apple_ref/occ/instp/NSLayoutConstraint/active)).

YuanZhu-apple: Use `removeConstraints:` to be consistency with other UIView subclass like `ORKSurveyAnswerCellForPicker`. 
And happens to fixed this crash.

YuanZhu-apple: This could be a temporary fix to not block other work.

rsanchezsaez: I see. Have you submitted a Radar regarding the iOS 9 API? If not, I can submit it together with a RK-based sample project.

YuanZhu-apple: Not yet, glad if you can do it.

rsanchezsaez: Done! rdar://22179283

I attached a minimum example Xcode project showing the crash. You can also see it [here](https://github.com/rsanchezsaez/iOS9-Constraint-Crash). I didn't have to include *ResearchKit*, this example reproduces the crash with very little code.

It happens when you call `deactiveConstraints:` with an array that has constraints for a view that was removed.

---

rsanchezsaez:
@YuanZhu-apple Not sure if GitHub sends notification for merged pull requests. Adding your name just in case: I added a the *rdar* reference for tracking, and some comments.

ResearchKit/ActiveTasks/ORKTowerOfHanoiTowerView.m
YuanZhu-apple: Thanks for opening the radar.
I like the idea of using a for loop, but it still crashes:

screen shot 2015-08-06 at 4 26 12 pm

rsanchezsaez: Weird, in my super-simple aforementioned Xcode example it did crash when using `deactivateConstraints:`, but it did not when using the loop. I guess the bug is more subtle that I thought.

Again, let's hope they fix it on the *iOS 9* codebase. ;-)

PR #366, opened 2015-08-07
Fix ORKResultPredicate bug

rugheid:
The predicate would fail when the task identifier was empty.
I've ran all tests again and they all pass.


rsanchezsaez:
I think that using the empty string as an *identifier* is not within *ResearchKit* expectations. Just curious, why are you doing that?

Instead of altering the predicates, maybe we should throw an exception if you try to pass a `@""` for a *task*, *step*, *recorder* or *formItem* identifier,

rugheid:
I was implementing the task and didn't know what to fill in there, so I filled in an empty string to change it later.

Throwing an exception would indeed be a more logical solution. Shall I change it and submit a new pull request?

rsanchezsaez:
I can't think of any reason for allowing the empty string as an identifier, but let's wait for @YuanZhu-apple comments to be extra sure. ;-)

rugheid:
Ok, perfect! Thanks!

YuanZhu-apple:
For the debugging perspective, I think it is make more sense to throw an exception on empty identifiers. 
@rugheid You can just update this PR.

YuanZhu-apple:
Hi @rugheid, it seems you committed body graph code into this PR.

rugheid:
Oops, my bad! Thanks for letting me know! :)

rugheid:
I'll just close this pull request for now and submit a new one when I've had time to work on the bug...

PR #370, opened 2015-08-09
iOS9 iPad split view support

YuanZhu-apple:
Fixed layout problem on iPad with iOS 9. (Tested in simulator)

Changes:
- Determine screen type from two dimensions: `vertical` and `horizontal`  
Renamed `ORKGetScreenTypeForWindow` to `ORKGetVerticalScreenTypeForWindow`
And introduced `ORKGetHorizontalScreenTypeForWindow`

       Use horizontal screen type to determine a few metrics:
       ` ORKScreenMetricContinueButtonWidth`
       `ORKScreenMetricHeadlineSideMargin`
        `ORKScreenMetricLearnMoreButtonSideMargin`
        ~~`ORKScreenMetricTopToIllustration`~~

- Update `layoutMargin` in `ORKVerticalContainerView`'s `updateConstraintConstants`

### DIFF
--- a/ResearchKit/Common/ORKSkin.m
+++ b/ResearchKit/Common/ORKSkin.m
@@ -96,7 +96,7 @@ const CGSize ORKiPhone6ScreenSize = (CGSize){375, 667};
 const CGSize ORKiPhone6PlusScreenSize = (CGSize){414, 736};
 const CGSize ORKiPadScreenSize = (CGSize){768, 1024};
 
-ORKScreenType ORKGetScreenTypeForBounds(CGRect bounds) {
+ORKScreenType ORKGetVerticalScreenTypeForBounds(CGRect bounds) {
     ORKScreenType screenType = ORKScreenTypeiPhone6;
     CGFloat maximumDimension = MAX(bounds.size.width, bounds.size.height);
     if (maximumDimension < ORKiPhone4ScreenSize.height + 1) {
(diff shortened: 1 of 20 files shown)


YuanZhu-apple:
@umerkhan-apple @rsanchezsaez Can you review this?

umerkhan-apple:
Tested on hardware. r=m3

ResearchKit/Common/ORKSkin.m
```diff
@@ -171,7 +198,22 @@ CGFloat ORKGetMetricForScreenType(ORKScreenMetric metric, ORKScreenType screenTy
...
 
 CGFloat ORKGetMetricForWindow(ORKScreenMetric metric, UIWindow *window) {
-    return ORKGetMetricForScreenType(metric, ORKGetScreenTypeForWindow(window));
+    
+    CGFloat ret = 0;
+    switch (metric) {
+        case ORKScreenMetricContinueButtonWidth:
+        case ORKScreenMetricHeadlineSideMargin:
+        case ORKScreenMetricLearnMoreButtonSideMargin:
+        case ORKScreenMetricTopToIllustration:
```
rsanchezsaez: Isn't `ORKScreenMetricTopToIllustration` a vertical metric?

ResearchKit/Common/ORKSkin.m
```diff
@@ -205,10 +247,17 @@ CGFloat ORKStandardHorizMarginForView(UIView *view) {
...
+            // Use adaptive side margin, if view is wider than iPhone6 Plus.
+            // Min Marign = ORKLayoutMarginWidthThinBezelRegular, Max Marign = ORKLayoutMarginWidthiPad
+            CGFloat ratio =  (view.bounds.size.width - ORKiPhone6PlusScreenSize.width)/(ORKiPadScreenSize.width - ORKiPhone6PlusScreenSize.width);
+            ratio = MIN(1.0, ratio);
+            ratio = MAX(0.0, ratio);
+            margin = ORKLayoutMarginWidthThinBezelRegular + (ORKLayoutMarginWidthiPad - ORKLayoutMarginWidthThinBezelRegular)*ratio;
             break;
+        }
     }
+    NSLog(@"margin = %@ [%@]", @(margin), @(view.bounds.size.width));
```
rsanchezsaez: Maybe remove this `NSLog`?

ResearchKit/Common/ORKSkin.m
YuanZhu-apple: Agree, updated.

ResearchKit/Common/ORKSkin.m
```diff
@@ -113,21 +113,48 @@ ORKScreenType ORKGetScreenTypeForBounds(CGRect bounds) {
...
+        screenType = ORKScreenTypeiPhone6Plus;
+    } else {
+        screenType = ORKScreenTypeiPad;
+    }
+    return screenType;
+}
+
+ORKScreenType ORKGetVerticalScreenTypeForWindow(UIWindow *window) {
     if (!window) {
         window = [[[UIApplication sharedApplication] windows] firstObject];
```
rsanchezsaez: Sorry for the late comment: wouldn't using `[UIApplication sharedApplication].keyWindow` here be more correct?

rsanchezsaez:
I was merging `master` into my [constraint refactor PR](https://github.com/ResearchKit/ResearchKit/pull/343) and I noticed that this PR seems to break the *wide iPad margins* feature when not using multi-tasking. The problem seems to be in [this line](https://github.com/ResearchKit/ResearchKit/pull/370/files#diff-a102b42394b359447ca686f03bed4954R252). Before, the method didn't rely on `view.bounds`, which are `CGRectZero` in many instances when views are being set up.

I'm working on a fix at the moment.

ResearchKit/Common/ORKSkin.m
rsanchezsaez: I figured that you use the window array method instead of `UIApplication`'s `keyWindow` or `UIApplication`'s `delegate`'s `window` because we may need the `window` before the `keyWindow` is set (e.g., if a view controller loads programmatically on the app delegate to be assigned as the root view controller).

YuanZhu-apple:
@rsanchezsaez Good catch, it should fall back to iPad-Margin if `view.bounds == CGRectZero`.

PR #372, opened 2015-08-10
iOS9: Split view fix for image capture step.

umerkhan-apple:
In iOS9 the image capture step does not function as intended when being used with split view. The `capture preview` freezes, but the `capture image` button is still visible. Tapping the `capture image` button causes the app to crash.

This fix removes the `capture image` button and the `capture preview` from the screen and displays a message for the user when in split view.


YuanZhu-apple:
r=me.

PR #375, opened 2015-08-14
Accessibility enhancement of Tower of Hanoi

YuanZhu-apple:


ResearchKit/ActiveTasks/ORKTowerOfHanoiStepViewController.m
```diff
@@ -143,6 +143,10 @@ - (void)towerOfHanoiTowerViewWasSelected:(ORKTowerOfHanoiTowerView *)towerView {
     [self evaluatePuzzle];
 }
 
+- (ORKTowerOfHanoiTowerView *)towerOfHanoiHighlightedTowerView {
+    return (_selectedIndex != nil ? _towerViews[_selectedIndex.integerValue] : nil);
+}
+
 #pragma Mark -- ORKTowerOfHanoiViewController
```
rsanchezsaez: I have not reviewed the accessibility changes per-se (busy with the graphs) but I spotted there are several incorrect `#pragma Mark -- ...` lines on the tower of hanoi files. This PR might be a good change to fix those. They should be ` #pragma mark - ...`.

YuanZhu-apple: Good idea.

umerkhan-apple:
r=m3

ResearchKit/ActiveTasks/ORKTowerOfHanoiStepViewController.m
```diff
@@ -57,7 +57,7 @@ @implementation ORKTowerOfHanoiViewController {
     NSDate *_firstMoveDate;
 }
 
-#pragma Mark -- UIViewController
+#pragma mark -- UIViewController
```
rsanchezsaez: For the separator line to work it needs to be one dash character (`-`) instead of two (`--`). I should have mentioned this explicitly in my earlier comment, sorry!

YuanZhu-apple: Good catch. Updated.

PR #377, opened 2015-08-17
Consent section: avoid headers without body

p2:
With this change, `ORKConsentSectionFormatter` will return an empty string if a section does not have content.

In our consenting we have 2 or 3 steps that do not have a more info text. These would show up as only a header in the generated HTML.


ResearchKit/Consent/ORKConsentSectionFormatter.m
```diff
@@ -35,9 +35,12 @@ OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 @implementation ORKConsentSectionFormatter
 
 - (NSString *)HTMLForSection:(ORKConsentSection *)section {
-    NSString *title = [NSString stringWithFormat:@"<h4>%@</h4>", section.formalTitle?:(section.title?:@"")];
-    NSString *content = [NSString stringWithFormat:@"<p>%@</p>", section.htmlContent?:(section.escapedContent?:@"")];
-    return [NSString stringWithFormat:@"%@%@", title, content];
+	NSString *content = section.htmlContent ?: (section.escapedContent ?: @"");
+	if (0 == content.length) {
+		return @"";
```
YuanZhu-apple: As I understand:
```
NSString *content = section.htmlContent ?: (section.escapedContent ?: @"");
```
`content` cannot be nil, then why need the `if` block?

YuanZhu-apple:
@p2 
Do you think it is better to add a boolean to indicate the section to be excluded from generated HTML?

p2:
I thought about that, but in our case it works just as well. The steps to exclude, like _Study Dashboard_, explain in one sentence something about the app. No need to put those in the consent, and since there is no “Learn More” it's excluded. Otherwise there would be a heading _Study Dashboard_ and no content, so at least in these cases it works well without bool.

The question arises if you have some sections that you indeed want to exclude. You could make the point to just override the whole htmlContent in that case, but then you lose the ability to add the signature to the generated PDF. Unless that part is changed... :)

YuanZhu-apple:
But I feel some people still want to keep the title itself in the HTML file. 

Imagine people want to exclude a section from HTML, but the section has both non empty title and content. 

Providing a boolean can accommodate different use cases.

p2:
Only the title looks kinda bad (it's the reason I made this change), but I agree with your second point.

YuanZhu-apple:
r=me.

YuanZhu-apple:
@p2 Can you update this patch? I want to merge it.

p2:
Sure; done.

YuanZhu-apple:
Thanks!

PR #378, opened 2015-08-17
Performance, Consolidation of Operations, and Formatting

newbyman:
- Framework Review
- Simplification of Operations
- Performance consolidation
- Performance breaks and returns 
- Code formatting updates in accordance with the provided coding style guide

### DIFF
--- a/ResearchKit/ActiveTasks/ORKReactionTimeViewController.m
+++ b/ResearchKit/ActiveTasks/ORKReactionTimeViewController.m
@@ -86,7 +86,7 @@ static const NSTimeInterval OutcomeAnimationDuration = 0.3;
 #if TARGET_IPHONE_SIMULATOR
 - (void)motionBegan:(UIEventSubtype)motion withEvent:(UIEvent *)event
 {
-    if(event.type == UIEventSubtypeMotionShake)
+    if (event.type == UIEventSubtypeMotionShake)
     {
         if (_validResult) {
             ORKReactionTimeResult *reactionTimeResult = [[ORKReactionTimeResult alloc] initWithIdentifier:self.step.identifier];
--- a/ResearchKit/ActiveTasks/ORKReactionTimeViewController.m
+++ b/ResearchKit/ActiveTasks/ORKReactionTimeViewController.m
@@ -84,10 +84,8 @@ static const NSTimeInterval OutcomeAnimationDuration = 0.3;
 }
 
 #if TARGET_IPHONE_SIMULATOR
-- (void)motionBegan:(UIEventSubtype)motion withEvent:(UIEvent *)event
-{
-    if (event.type == UIEventSubtypeMotionShake)
-    {
+- (void)motionBegan:(UIEventSubtype)motion withEvent:(UIEvent *)event {
+    if (event.type == UIEventSubtypeMotionShake) {
         if (_validResult) {
             ORKReactionTimeResult *reactionTimeResult = [[ORKReactionTimeResult alloc] initWithIdentifier:self.step.identifier];
             reactionTimeResult.timestamp = _stimulusTimestamp;
--- a/ResearchKit/Common/ORKFormItemCell.m
+++ b/ResearchKit/Common/ORKFormItemCell.m
@@ -831,7 +831,7 @@ static const CGFloat kHMargin = 15.0;
 }
 
 - (id<ORKScaleAnswerFormatProvider>)formatProvider {
-    if (_formatProvider == nil){
+    if (_formatProvider == nil) {
         _formatProvider = (id<ORKScaleAnswerFormatProvider>)[self.formItem.answerFormat impliedAnswerFormat];
     }
     return _formatProvider;
--- a/ResearchKit/Common/ORKFormItemCell.m
+++ b/ResearchKit/Common/ORKFormItemCell.m
@@ -831,7 +831,7 @@ static const CGFloat kHMargin = 15.0;
 }
 
 - (id<ORKScaleAnswerFormatProvider>)formatProvider {
-    if(_formatProvider == nil){
+    if (_formatProvider == nil){
         _formatProvider = (id<ORKScaleAnswerFormatProvider>)[self.formItem.answerFormat impliedAnswerFormat];
     }
     return _formatProvider;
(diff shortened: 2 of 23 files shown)


ResearchKit/Consent/ORKConsentLearnMoreViewController.m
```diff
@@ -54,11 +54,9 @@ - (instancetype)initWithHTMLContent:(NSString *)content {
     return self;
 }
 
--(instancetype)initWithContentURL:(NSURL * __nonnull)contentURL
-{
+- (instancetype)initWithContentURL:(NSURL * __nonnull)contentURL {
```
rsanchezsaez: We don't normally add nullability annotations in implementation files. Could you take the chance to review if there are more around, and remove them?

I'll add this to the *Coding Style Guideline* later.

ResearchKit/Consent/ORKConsentLearnMoreViewController.m
```diff
@@ -54,11 +54,9 @@ - (instancetype)initWithHTMLContent:(NSString *)content {
...
     self = [super init];
-    if(self)
-    {
+    if(self) {
```
rsanchezsaez: Missing space after `if`. You could also fix this everywhere: a simple search-and-replace will do.

ResearchKit/Consent/ORKConsentLearnMoreViewController.m
```diff
@@ -81,8 +79,7 @@ - (void)viewDidLoad {
     _webView.scrollView.scrollIndicatorInsets = (UIEdgeInsets){.left = -horizMargin, .right = -horizMargin};
     _webView.opaque = NO; // If opaque is set to YES, _webView shows a black right margin during transition when modally presented. This is an artifact due to disabling clipsToBounds to be able to show the scroll indicator outside the view.
     
-    if(_contentURL)
-    {
+    if(_contentURL) {
```
rsanchezsaez: Same.

ResearchKit/Consent/ORKVisualConsentStepViewController.m
```diff
@@ -661,13 +662,12 @@ - (NSUInteger)indexOfViewController:(UIViewController *)viewController {
         return NSNotFound;
     }
     
-    NSUInteger index = NSNotFound;
     for (NSNumber *key in _viewControllers) {
         if (_viewControllers[key] == viewController) {
-            index = [key unsignedIntegerValue];
+            return [key unsignedIntegerValue];
```
rsanchezsaez: I personally prefer having a `break` here, instead of two `return`s (a single method exit point is easier to read). Like you did in `-scrollView`. ;-)

rsanchezsaez:
Nice PR. r=me.

ResearchKit/ActiveTasks/ORKReactionTimeViewController.m
```diff
@@ -86,7 +86,7 @@ - (void)start {
 #if TARGET_IPHONE_SIMULATOR
 - (void)motionBegan:(UIEventSubtype)motion withEvent:(UIEvent *)event
 {
-    if(event.type == UIEventSubtypeMotionShake)
+    if (event.type == UIEventSubtypeMotionShake)
     {
```
umerkhan-apple: Move `{` to same line as `if statement`

ResearchKit/Common/ORKFormItemCell.m
```diff
@@ -831,7 +831,7 @@ @implementation ORKFormItemScaleCell {
 }
 
 - (id<ORKScaleAnswerFormatProvider>)formatProvider {
-    if(_formatProvider == nil){
+    if (_formatProvider == nil){
         _formatProvider = (id<ORKScaleAnswerFormatProvider>)[self.formItem.answerFormat impliedAnswerFormat];
```
umerkhan-apple: Space after `(`

ResearchKit/Consent/ORKConsentSceneViewController.m
```diff
@@ -192,10 +192,9 @@ - (void)scrollToTopAnimated:(BOOL)animated completion:(void (^)(BOOL finished))c
 - (IBAction)showContent:(id)sender {
     ORKConsentLearnMoreViewController *viewController = nil;
     
-    if(_section.contentURL)
-    {
+    if (_section.contentURL) {
         viewController = [[ORKConsentLearnMoreViewController alloc] initWithContentURL:_section.contentURL];
-    }else{
+    } else{
```
umerkhan-apple: Add space between `else` and `{`

ResearchKit/Common/ORKFormItemCell.m
rsanchezsaez: But we usually don't add a space between a *cast* and the *casted value*, do we?

umerkhan-apple: No I was referring to the `if` statement. It should look like `if (_formatProvider == nil) {`

ResearchKit/Consent/ORKVisualConsentStepViewController.m
newbyman: We have three logical returns within the scope of the method.  To put the break back in on this one, we would need to restore the NSUInteger to serves as a primitive container, vs the latter (hair lighter on memory and faster) saves us the container, setter, break, and return vs return.   (May not be needed within this scope, however adopted pattern long time ago from mathematically modeling and realized performance gains.)  Intention was just to save the "NSUInteger index = NSNotFound;, that may not be needed.

rsanchezsaez: That's fair enough!

However, again, I don't think that exit points in the middle of methods is good style, as could be error prone. This method is simple enough, but with longer ones they can lead to "*code blindness*".

I prefer that all method exit points are either at the top (because of initial conditions, like the first one here), or at the bottom. That way you can follow the flow of what the function does easily.

I think legibility is more important than performance in this case (and probably the compiler would optimize the extra variable away).

@YuanZhu-apple decides. ;-)

YuanZhu-apple: I prefer `break` over second `return`.

newbyman: No problem.  PR updated.

YuanZhu-apple:
r=me.

PR #380, opened 2015-08-18
Alternate fix for Tower of Hanoi iOS 9 constraint crash

rsanchezsaez:
I figured an alternate fix for the *iOS 9* crash when you call `NSLayoutConstraint deactivateConstraints:` on constraints that reference removed views.

It adds the `void ORKRemoveConstraintsForRemovedViews(NSMutableArray *constraints, NSMutableArray *removedViews)` function to `ORKHelpers`. You have to call this whenever you remove subviews on a view that has a managed `_variableConstraints` mutable array.

Benefits of this fix:
- It allows to continue using the recommended `[NSLayoutConstraint deactivateConstraints:]`.
- It is easily compatible with my *constraint refactor* [pull request](https://github.com/ResearchKit/ResearchKit/pull/343), without having to switch back and forward between `[NSLayoutConstraint deactiveConstraints:]` and `[view removeConstraints:]`.
- When the bug is fixed, it'll be very easy to remove the fix by removing the new function and all of its calls, and the code will remain clean and in working order.

If, in the future, you merge the aforementioned *constraint refactor* PR before the bug is fixed at the iOS level, I can check it add calls to this function everywhere is needed.

### DIFF
--- a/ResearchKit/ActiveTasks/ORKTowerOfHanoiTowerView.m
+++ b/ResearchKit/ActiveTasks/ORKTowerOfHanoiTowerView.m
@@ -33,17 +33,19 @@
 #import "ORKActiveStepView.h"
 #import "ORKSkin.h"
 #import "ORKAccessibility.h"
+#import "ORKHelpers.h"
 
-static const CGFloat kDiskHeight = 10;
-static const CGFloat kDiskSpacing = 8;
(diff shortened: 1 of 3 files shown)


YuanZhu-apple:
Thanks! and  r=me.

Issue #224, opened 2015-05-25
Step navigation rule initializer should have Swift overlay to use tuples


jwe-apple:
The Step navigation initializer and the step navigation rule overall should use tuples in Swift. The Objective-C API is not ideal, but acceptable given we can improve it in Swift. This issue is for adding a Swift overlay initializer and any other necessary components, and potentially hiding the Objective-C initializer in Swift.

rsanchezsaez:
This can be closed as the feature was implemented and merged.

YuanZhu-apple:
Yes.

PR #383, opened 2015-08-19
Add Swift tuple initializer overlay to ORKPredicateStepNavigationRule

rsanchezsaez:
This adds a *Swift* overlay for `ORKPredicateStepNavigationRule` which takes an array of `(NSPredicate, String)` tuples (*result predicate*, *destination step identifier*) instead of two separate arrays. ([Issue #224](https://github.com/ResearchKit/ResearchKit/issues/224))

The *Navigable Ordered Task* creation code in `ORKTest` was converted to *Swift*, and it now uses the overlay *Swift* initializer for creating the *predicate step navigation rules*.

Note that this sets `CLANG_ENABLE_MODULES` to `YES` in the *ResearchKit* project, I hope it doesn’t introduce any unintended side-effects.

~~On `iOS 9`, we’ll be able to use [`NS_SWIFT_UNAVAILABLE(“”)`](https://developer.apple.com/library/prerelease/ios/documentation/Swift/Conceptual/BuildingCocoaApps/MixandMatch.html#//apple_ref/doc/uid/TP40014216-CH10-ID178) to disallow the raw *Objective-C* initializers from *Swift*.~~ Since we're now supporting *Xcode 7* on `master`, I used the macro to made the *ObjC* initializers unavailable from *Swift*.

Is this the correct approach for creating a *Swift* overlay?


rsanchezsaez:
Hey @YuanZhu-apple, do you think this could be merged any time soon?

I'm looking at the backwards navigation problem upon `ORKNavigableOrderedTask` restoration and it would be nice to have this in to avoid conflicts.

Testing/ORKTest/ORKTest/TestHelpers.swift
```diff
@@ -0,0 +1,168 @@
...
+SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*/
+
+import Foundation
+import ResearchKit
+
+@objc class TestHelpers : NSObject {
```
YuanZhu-apple: Is `TaskFactory` a better name for this class?

YuanZhu-apple:
Thanks @rsanchezsaez, the Swift tuple initializer is more naturally than the Ojc version.

Testing/ORKTest/ORKTest/MainViewController.m
```diff
@@ -475,7 +479,7 @@ - (UICollectionViewCell *)collectionView:(UICollectionView *)collectionView cell
                                                               rotated:NO
                                                             timeLimit:300.0
                                                               options:ORKPredefinedTaskOptionNone];
-    } else if ([identifier isEqualToString:StepNavigationTaskIdentifier]) {
+    } else if ([identifier isEqualToString:NavigableOrderedTaskIdentifier]) {
         return [self makeNavigableOrderedTask];
```
YuanZhu-apple: Call swift factory method directly from here?

rsanchezsaez:
Thanks for the review, I'll make the suggested changes soon.

rsanchezsaez:
Changes done, and synced to master. I also fixed a problem that prevented *RK's unit tests* to run due to the new embedded Swift code.

YuanZhu-apple:
R=me.
Thanks @rsanchezsaez, merging this one.

chrisnowak:
Hey @rsanchezsaez - I'm not 100% sure this is the root issue, but I think this new Swift overlay is not included in the ResearchKit cocoapod due to Swift files not being part of the "source" in the podspec here:
https://github.com/ResearchKit/ResearchKit/blob/master/ResearchKit.podspec#L11

Whenever I try to reference master as a cocoapod (I'm trying to play with some of the new onboarding stuff slated for next release), it can't find the "resultPredicatesAndDestinationStepIdentifiers" method.

Issue #379, opened 2015-08-17
Chart classes improvements


rsanchezsaez:
Some possible `ORK*GraphChartView` improvements:
- ~~Add an optional delegate method to provide colors for secondary graphs on multi-plot graphs. Currently, the main plot is painted of the view's `tintColor`, and all the secondary plots are painted of the `referenceLineColor`.~~ *Implemented and merged.*
- ~~Add an optional delegate method to allow to select which graph to scrub over on multi-plot graphs (currently, it always scrubs over the first graph).~~ *Implemented and merged.*
- Add support for graphs with a large number of points, by adding a property for setting an interval on the *xAxis* labels and reference lines (currently the *xAxis* gets very crowded if you add lots of points).

rsanchezsaez:
All items implemented and merged by the referenced pull requests, closing.

PR #385, opened 2015-08-20
Charts improvements

rsanchezsaez:
Several improvements to the charts.

*Graph Charts*
- Add optional delegate method to provide colors for secondary graphs on multi-plot graphs ([Issue #379](https://github.com/ResearchKit/ResearchKit/issues/379)).
- Add optional delegate method to allow to select which graph to scrub over on multi-plot graphs  ([Issue #379](https://github.com/ResearchKit/ResearchKit/issues/379)).
- Fix scrubbing animation glitch.

*Pie Chart*
- Use shades of current `tintColor` (instead of shades of black) when no colors are supplied to the pie chart view.
- Fix animation not working if `showsPercentageLabels` was set to `NO`.

*All*
- Properties with default values can be reset to their default value if set to `nil` (and declare them as `null_resetable`).
- Documentation improvements.
- Other minor fixes.


rsanchezsaez:
@YuanZhu-apple: This PR have some chart fixes that I thing ought to be in `convergence`:
- Fix *pie chart* animation not working if `showsPercentageLabels` was set to `NO`.
- Fix *pie chart* legend height not being updated properly after a rotation.
- Fix *pie chart* legend possibly being added several times if you change a valid delegate to a different valid one.
- Workaround for *pie chart* *iOS 9* constraint crash (rdar://22179283, same as what happened on the *Tower Of Hanoi* task).

Do you want me to cherry pick the fixes without the enhancements (new delegate methods, pie chart `tintColor`, and `null_resettable` properties) and submit a new PR against `convergence`?

YuanZhu-apple:
Yes! I like to have a separated PR against convergence, cherry picks the fixes without the enhancements.

rsanchezsaez:
I'm working on this.

But, are you sure we don't want to include the enhancements? Since the next release will be the first one with charts, isn't it a shame that we don't include the improvements from the get go? Specially, the conversion of the charts properties to`null_resettable` would be a notable miss from the base API.

YuanZhu-apple:
I am OK with the enhancements to go into convergence as long as they do not require additional accessibility and localization.

rsanchezsaez:
I think they don't. ;-) I'll submit a PR against convergence in some moments.

PR #393, opened 2015-08-24
Improve localization string for accessibility of Hanoi Tower

YuanZhu-apple:
Put `%@` in the loc string.


YuanZhu-apple:
Reviewed by internal.

PR #396, opened 2015-08-25
Refactoring consent localization keys

umerkhan-apple:
Refactored the localization keys for consent first name and consent last name to consent given name and consent family name.


YuanZhu-apple:
r=me.

PR #398, opened 2015-08-26
Fix a crash by adding a boundary check

YuanZhu-apple:


YuanZhu-apple:
reviewed by @jwe-apple

PR #407, opened 2015-08-27
Convergence fix multiline desc on slider

YuanZhu-apple:
- Make slider's view bottom greater than slider's bottom, also greater than lower descriptionLabel's bottom.
- Make current value above slider's top, also above upper description Label's top.


YuanZhu-apple:

umerkhan-apple:
r=m3.

PR #409, opened 2015-08-28
Fix NSRangeException in Hanoi Task

YuanZhu-apple:
Tower could be empty, so it is not safe to call `tower.disks[index];` without a check.
Accessibility announcement is only applicable when the both donor and recipient are valid.


YuanZhu-apple:

umerkhan-apple:
r=m3.

PR #411, opened 2015-08-28
Fix text label layout problem in right-to-left UI.

YuanZhu-apple:
- Make the checkmark always in the choiceCell, hide it when the cell is unselected.
So layout code get consistent bounds to calculate label's rect.

iphone_timed_walk_2_1_6


umerkhan-apple:
r=m3

PR #413, opened 2015-08-28
Show family name first for Vietnamese

YuanZhu-apple:


ResearchKit/Common/ORKHelpers.m
```diff
@@ -327,7 +327,7 @@ BOOL ORKCurrentLocalePresentsFamilyNameFirst() {
     static dispatch_once_t onceToken;
```
rsanchezsaez: Should we remove the space between the `*` and language?

ResearchKit/Common/ORKHelpers.m
```diff
@@ -327,7 +327,7 @@ BOOL ORKCurrentLocalePresentsFamilyNameFirst() {
...
     static NSArray *familyNameFirstLangs = nil;
```
rsanchezsaez: Should we rename the variable to `familyNameFirstLanguages`? `Lang` is not accepted abbreviation [here](https://developer.apple.com/library/mac/documentation/Cocoa/Conceptual/CodingGuidelines/Articles/APIAbbreviations.html).

rsanchezsaez:
Reviewed. Two nits, otherwise it looks good!

rsanchezsaez:
By the way, I'm [reading](https://en.wikipedia.org/wiki/Family_name#Overview) that the family name also comes first in *Hungarian* and *Romanian*. I don't have access to any natives to check it out. It may be worth confirming to also add those two if that's the case.

YuanZhu-apple:
Maybe not this time for Hungarian and Romanian. Rather wait until there is radar from our QA.

oliverschaefer:
Hi @YuanZhu-apple @rsanchezsaez my mother is Hungarian so I'm at least a 50% native :-). The family name comes first. Hope that helps

PR #414, opened 2015-08-29
Fix placeholder get truncated in FormItemCell.

YuanZhu-apple:
Take _textFieldView's estimatedWidth into consideration when building the constraints for textFieldBasedformItemCell.

- implemented _textFieldView.estimatedWidth to return the larger value of text+unit and placeholder+unit 

screen shot 2015-08-28 at 6 07 26 pm


umerkhan-apple:
r=m3

Issue #401, opened 2015-08-26
ORKResultPredicate for skipped steps or form items


rsanchezsaez:
Investigate the possibility of adding explicit support for detecting *skipped steps* or *skipped form items* through predicates.

Probably it's a good idea to solve [Issue #354](https://github.com/ResearchKit/ResearchKit/issues/354) before tackling this one.

rsanchezsaez:
Solved by the linked PR.

Issue #354, opened 2015-08-05
ORKResultPredicate for form step


rugheid:
Is it possible to define navigation rules for a form step for an ORKNavigableOrderedTask? If so, how would I go about doing that?

Thanks

jwe-apple:
What do you mean exactly? Do you want the form step itself to change? Or merely to change what the next step is based on the results of the form? You can definitely do the latter; I believe you can't do the former at present, though there has been some discussion about it.

rugheid:
I want to change the next step based on the results of the form. Every result predicate needs a result identifier, so what to I need to fill in there? Once I have the result predicate I assume I do it as I would with a question step. I just create a navigation rule using one or more predicates and set it on the task.

Thanks for your help!

rsanchezsaez:
@rugheid: Yeah, result predicates do work with forms: you need to pass the `ORKFormItem`'s `identifier` as the predicate `resultIdentifier`. To exemplify this I added a *form step* to `ORKTest`'s *Navigable Ordered Task* ([PR #360](https://github.com/ResearchKit/ResearchKit/pull/360)).

---

@jwe-apple: This made me realize a potential issue. Given this result hierarchy:

    taskResult
     |
     |_ stepResult
     |    |
     |    |_ questionResult
     |
     |_ formStepResult
          |
          |_ formQuestionResultA
          |
          |_ formQuestionResultB

*ResearchKit* allows `questionResult`and either `formQuestionResultA` or `formQuestionResultB` to have the same *identifier* (since they are scoped by the step identifier).

For *single-question steps*, the *questionResult* identifier is always the same as the *stepResult* identifier, but for *forms* the child *question results* get the identifiers from the *formItems*. Currently, *result predicates* only look for leaf *question results*, ignoring whether they came from *question* or *form steps*. So if you have duplicate *question result identifiers*, the predicates would match both.

The same problem arises if you have several forms on the same survey (child question results could have repeated identifiers between the different forms).

Two course of actions come to mind:
- Make sure the current predicates only match single-question steps (by also matching the *questionStepResult* identifier) and add a whole new set of methods for constructing *form result predicates*, in which you pass both the *form step identifier* and the *item identifier*. Basically, the number of public *result predicate*-building methods would duplicate. 
- Leave the code as is and document the issue in the headers (that'd place the burden on the user by recommending not to use duplicate identifiers between *form items* and *single-question steps*).

Thoughts?

rugheid:
[...]

PR #416, opened 2015-08-30
Result predicates: add support for form item results and nil results 

rsanchezsaez:
This PR adds explicit support for *form item results* ([Issue #354](https://github.com/ResearchKit/ResearchKit/issues/354)) and for *results with a `nil` answer* ([Issue #401](https://github.com/ResearchKit/ResearchKit/issues/401)).

A new optional `stepIdentifier:(nullable NSString *)stepIdentifier` argument has been added to all of the predicate-constructing methods (along with new convenience methods). For matching *form item results*, the *form step identifier* must be passed through this argument. For matching *question results within single-question steps*, this argument can be omitted. Predicates without a `stepIdentifier` no longer match *form item results*.

Also, a family of three `+predicateForNilQuestionResultWithTaskIdentifier:stepIdentifier:resultIdentifier:` methods have been added. These can be used to match *results with a `nil` answer* (produced by skipped steps or unanswered form questions).

I have added tests to `ORKTest` covering these two features.


ResearchKit/Common/ORKResultPredicate.h
```diff
@@ -196,6 +319,24 @@ ORK_CLASS_AVAILABLE
...
+                                assumes the step identifier is the same as the result identifier.
+ @param resultIdentifier    The identifier of the question result you are interested in.
+ @param expectedString      The expected string answer.
+ 
+ @return A result predicate.
+ */
++ (NSPredicate *)predicateForChoiceQuestionResultWithTaskIdentifier:(nullable NSString *)taskIdentifier
+                                                     stepIdentifier:(nullable NSString *)stepIdentifier
+                                                   resultIdentifier:(NSString *)resultIdentifier
+                                                     expectedString:(NSString *)expectedString;
```
YuanZhu-apple: The `value` property of an `ORKTextChoice` is not limited to `NSString`.

```
@property (copy, readonly) id<NSCopying, NSCoding, NSObject> value;
```
And ORKChoiceQuestionResult only carries an array of the value object. Does this predicate require the textChoice to use a string as value object?

ResearchKit/Common/ORKResultPredicate.h
```diff
@@ -250,6 +409,24 @@ ORK_CLASS_AVAILABLE
...
+ @param stepIdentifier      The identifier of the step whose result you want to match. Passing `nil`
+                                assumes the step identifier is the same as the result identifier.
+ @param resultIdentifier    The identifier of the question result you are interested in.
+ @param pattern             An ICU-compliant regular expression pattern that matches the answer string.
+ 
+ @return A result predicate.
+ */
++ (NSPredicate *)predicateForChoiceQuestionResultWithTaskIdentifier:(nullable NSString *)taskIdentifier
+                                                     stepIdentifier:(nullable NSString *)stepIdentifier
+                                                   resultIdentifier:(NSString *)resultIdentifier
```
YuanZhu-apple: Maybe we can create a seperate class to contain `taskIdentifier`, `stepIdentifier`, and `resultIdentifier`??
Then the number of init method can be reduced in this class.

rsanchezsaez: Good idea!

The only downside is that the new methods won't be backwards compatible with the current `stable` branch, but I don't think that's too much of a problem. Is it?

ResearchKit/Common/ORKResultPredicate.h
```diff
@@ -508,20 +838,48 @@ ORK_CLASS_AVAILABLE
...
                                                       resultIdentifier:(NSString *)resultIdentifier
-                                             minimumExpectedAnswerHour:(NSInteger)minimumExpectedAnswerHour
-                                           minimumExpectedAnswerMinute:(NSInteger)minimumExpectedAnswerMinute
-                                             maximumExpectedAnswerHour:(NSInteger)maximumExpectedAnswerHour
-                                           maximumExpectedAnswerMinute:(NSInteger)maximumExpectedAnswerMinute;
+                                                   minimumExpectedHour:(NSInteger)minimumExpectedHour
+                                                 minimumExpectedMinute:(NSInteger)minimumExpectedMinute
+                                                   maximumExpectedHour:(NSInteger)maximumExpectedHour
+                                                 maximumExpectedMinute:(NSInteger)maximumExpectedMinute;
+
```
YuanZhu-apple: To be consistent with `ORKTimeOfDayQuestionResult`.
Can you use dateComponents as the values?

ResearchKit/Common/ORKResultPredicate.h
rsanchezsaez: Good catch.

I can generalize this method to just take an `id<NSCopying, NSCoding, NSObject>` expected value. And keep the other pattern matching method as is (`predicateForChoiceQuestionResultWithResultIdentifier:(NSString *)matchingPatterns:(NSArray<NSString *> *)`) if you want to use regular expressions for the specific case of a text-containing choice question result (which I guess it's a common use case).

ResearchKit/Common/ORKResultPredicate.h
rsanchezsaez: The problem here is that if we directly use `NSDateComponents` objects on predicates, you cannot use the *greater than* and *less than* operators. You have to manually access their components to compare them. And I think `NSDateComponents` has too many fields to provide comprehensive support.

So, after discussing with John, we ended leaving this bare-bones method for the very basic (and hopefully common) use case.

Discussion in the [original PR](https://github.com/ResearchKit/ResearchKit/pull/151#issuecomment-102458319):

> [Me]: Yeah, there are too many `NSDateComponents` for covering all of them. But I did this because of a comment in `ORKTimeOfDayQuestionResult`'s *HeaderDoc*: `/** Typically only hour, minute, and AM/PM data are of interest. */`. Couldn't this method be useful anyway? If the developer needs more, she can just define further methods or manually create more predicates.

And we ended up deciding to keep it in an offline discussion.

ResearchKit/Common/ORKResultPredicate.h
YuanZhu-apple: Not backwards compatible is fine:)

ResearchKit/Common/ORKResultPredicate.h
YuanZhu-apple: Then, It is good.

rsanchezsaez:
PR updated. Introducing the `ORKResultIdentifier` wrapper was an excellent idea! It greatly simplified the API and made maintenance much more manageable.

ResearchKit/Common/ORKResultPredicate.h
```diff
@@ -41,6 +41,133 @@ ORK_EXTERN NSString *const ORKResultPredicateTaskIdentifierVariableName ORK_AVAI
 #define ORKIgnoreTimeIntervalValue (ORKIgnoreDoubleValue)
 
 /**
+ The `ORKResultIdentifier` class provides a simple wrapper object containing the `taskIdentifier`,
+ `stepIdentifier` and `resultIdentifier` strings. You use this object specify the question result
+ you are interested in when building result predicates. See `ORKResultPredicate` for more
+ information.
+ */
+ORK_CLASS_AVAILABLE
+@interface ORKResultIdentifier : NSObject <NSSecureCoding, NSCopying>
```
YuanZhu-apple: I feels `[ORKResultIdentifier  identifierWithResultIdentifier: ]` is a little bit confusing. Two `resultIdentifer` actually have different meanings. 

Rename  `ORKResultIdentifier` to 
`ORKResultPredicateIdentfier`? 
`ORKResultPredicateReference`? 
`ORKResultReference`? 
`ORKResultDescriptor`? 

I am also open to other names as well.

rsanchezsaez:
How about `ORKQuestionResultIdentifier`? I also like `ORKResultPredicateIdentifier`, but it can be confusing because the *identifier wrapper object* is not for a *predicate*, but for a *result*.

The other suggestions feel a little bit cryptic.

rsanchezsaez:
I renamed it to `ORKQuestionResultIdentifier`. If you actually prefer other name, let me know and I'll refactor it tomorrow.

YuanZhu-apple:
`ORKQuestionResultIdentifier` is quite good, which matches `ORKQuestionResult` and its subclasses.

But do you like `ORKUniqueResultIdentifier` better? To emphasize it can uniquely identified the the result object.

rsanchezsaez:
I see the intention of using the `Unique` prefix but, isn't that misleading given that all of *RK* identifiers must be unique within their scope?

Which brought a new suggestion: what about `ORKScopedResultIdentifier`?

YuanZhu-apple:
Maybe `ORKResultSelector`? :)

rsanchezsaez:
I like both `ORKScopedResultIdentifier` and `ORKResultSelector`. Tell me your final preference and I'll update the PR now. ;-)

YuanZhu-apple:
@umerkhan-apple and I like `ORKResultSelector` better:)

rsanchezsaez:
I have updated the refactor commit with the `ORKResultSelector` name. In hindsight it was the better option due to its compactness. It grows on you. ;-)

ResearchKit/Common/ORKResultPredicate.h
```diff
@@ -41,24 +41,161 @@ ORK_EXTERN NSString *const ORKResultPredicateTaskIdentifierVariableName ORK_AVAI
...
+@interface ORKResultSelector : NSObject <NSSecureCoding, NSCopying>
+
+/*
+ The `init` and `new` methods are unavailable.
+ */
++ (instancetype)new NS_UNAVAILABLE;
+- (instancetype)init NS_UNAVAILABLE;
+
+/**
+ Returns a result identifier initialized with the specified arguments.
```
YuanZhu-apple: Should be:
Return a result ~~identifier~~ selector

ResearchKit/Common/ORKResultPredicate.h
```diff
@@ -41,24 +41,161 @@ ORK_EXTERN NSString *const ORKResultPredicateTaskIdentifierVariableName ORK_AVAI
...
+ @param resultIdentifier    The result identifier string.
+ 
+ @return A result selector.
+ */
++ (instancetype)selectorWithTaskIdentifier:(nullable NSString *)taskIdentifier
+                            stepIdentifier:(nullable NSString *)stepIdentifier
+                          resultIdentifier:(NSString *)resultIdentifier;
+
+/**
+ Returns a result identifier initialized with the specified arguments.
```
YuanZhu-apple: Here also.

ResearchKit/Common/ORKResultPredicate.h
```diff
@@ -41,24 +41,161 @@ ORK_EXTERN NSString *const ORKResultPredicateTaskIdentifierVariableName ORK_AVAI
...
+ @param taskIdentifier      An optional task identifier string.
+ @param resultIdentifier    The result identifier string.
+ 
+ @return A result selector.
+ */
++ (instancetype)selectorWithTaskIdentifier:(nullable NSString *)taskIdentifier
+                          resultIdentifier:(NSString *)resultIdentifier;
+
+/**
+ Returns a result identifier initialized with the specified arguments.
```
YuanZhu-apple: Here

ResearchKit/Common/ORKResultPredicate.h
```diff
@@ -41,24 +41,161 @@ ORK_EXTERN NSString *const ORKResultPredicateTaskIdentifierVariableName ORK_AVAI
...
+ @param stepIdentifier      An optional step identifier string.
+ @param resultIdentifier    The result identifier string.
+ 
+ @return A result selector.
+ */
++ (instancetype)selectorWithStepIdentifier:(nullable NSString *)stepIdentifier
+                          resultIdentifier:(NSString *)resultIdentifier;
+
+/**
+ Returns a result identifier initialized with the specified arguments.
```
YuanZhu-apple: Here

ResearchKit/Common/ORKResultPredicate.h
```diff
@@ -41,24 +41,161 @@ ORK_EXTERN NSString *const ORKResultPredicateTaskIdentifierVariableName ORK_AVAI
...
+ 
+ @param resultIdentifier    The result identifier string.
+ 
+ @return A result selector.
+ */
++ (instancetype)selectorWithResultIdentifier:(NSString *)resultIdentifier;
+
+/**
+ Returns a result identifier initialized with the specified arguments.
```
YuanZhu-apple: Here

ResearchKit/Common/ORKResultPredicate.h
```diff
@@ -41,24 +41,161 @@ ORK_EXTERN NSString *const ORKResultPredicateTaskIdentifierVariableName ORK_AVAI
...
+ @param resultIdentifier    The result identifier string.
+ 
+ @return A result selector.
+ */
+- (instancetype)initWithTaskIdentifier:(nullable NSString *)taskIdentifier
+                        stepIdentifier:(nullable NSString *)stepIdentifier
+                      resultIdentifier:(NSString *)resultIdentifier NS_DESIGNATED_INITIALIZER;
+
+/**
+ Returns a result identifier initialized with the specified arguments.
```
YuanZhu-apple: Here

ResearchKit/Common/ORKResultPredicate.h
```diff
@@ -41,24 +41,161 @@ ORK_EXTERN NSString *const ORKResultPredicateTaskIdentifierVariableName ORK_AVAI
...
+ @param taskIdentifier      An optional task identifier string.
+ @param resultIdentifier    The result identifier string.
+ 
+ @return A result selector.
+ */
+- (instancetype)initWithTaskIdentifier:(nullable NSString *)taskIdentifier
+                      resultIdentifier:(NSString *)resultIdentifier;
+
+/**
+ Returns a result identifier initialized with the specified arguments.
```
YuanZhu-apple: Here

ResearchKit/Common/ORKResultPredicate.h
```diff
@@ -41,24 +41,161 @@ ORK_EXTERN NSString *const ORKResultPredicateTaskIdentifierVariableName ORK_AVAI
...
+ @param stepIdentifier      An optional step identifier string.
+ @param resultIdentifier    The result identifier string.
+ 
+ @return A result selector.
+ */
+- (instancetype)initWithStepIdentifier:(nullable NSString *)stepIdentifier
+                      resultIdentifier:(NSString *)resultIdentifier;
+
+/**
+ Returns a result identifier initialized with the specified arguments.
```
YuanZhu-apple: Here

rsanchezsaez:
Well spotted! Updated.

YuanZhu-apple:
r=me. THanks!

PR #417, opened 2015-09-01
 Should adjust PageViewControllerNavigationDirection for RTL

YuanZhu-apple:
For right to left languages.


umerkhan-apple:
r=m3.

PR #418, opened 2015-09-01
Convergence fix NaN crash in chart

YuanZhu-apple:
The type of numberOfXAxisPoints should be `NSInteger`.


umerkhan-apple:
r=m3!

PR #420, opened 2015-09-01
Reorganize 'ORKCatalog' into subsections

rsanchezsaez:
Since `ORKCatalog` is also growing, I have reorganized it into alphabetically-sorted sections mirroring the [project's readme](https://github.com/ResearchKit/ResearchKit/blob/master/README.md) structure.

Let me know what you think.


YuanZhu-apple:
Overall it is very good.

Is the alignment difference on purpose?
screen shot 2015-09-01 at 4 26 13 pm

rsanchezsaez:
Yes, to differentiate between the *container survey/form examples*, and the
*individual question examples* which can be used in either one (except
*image capture*, which is survey-only, I think).

YuanZhu-apple:
Not quite like alignment difference.
For the questions, maybe better to create a separate section for them? What do you think?

rsanchezsaez:
Yeah, I was not completely convinced either. I've added new commit moving the *Survey Questions* to a new section.

YuanZhu-apple:
r=me.

YuanZhu-apple:
@rsanchezsaez Can you get a chance to update this PR? So I can merge it.

rsanchezsaez:
@YuanZhu-apple: Did you mean reorganizing the files as I just did in `ORKTest`? I just added commits doing that.

Also, I noticed that additional localization languages are not added in `ORKCatalog` so all tasks show in English. Should we add as they are in `ORKTest`?

YuanZhu-apple:
Sorry, I meant making this PR mergeable.

YuanZhu-apple:
Or you want do `files reorganizing` in this PR as well?

rsanchezsaez:
Ah, I see. Ok, I'll merge this with master, plus the *files reorganization*.

Let me fix something first: I'm going to remove the `samples` to `Samples` folder rename commit, as it messes with *Xcode* history when using the *Version Editor*, so it's not worth it.

rsanchezsaez:
Ok:
- Merged `master`.
- Reorganized project files.
- Updated `ResultTableProvides` to match `TaskListRow`.
- Updated `readme.md` to add formatting and information about sections.

YuanZhu-apple:
r=me.

PR #422, opened 2015-09-02
Fix recorder file naming

YuanZhu-apple:
Should use [_recorderUUID uuidString] instead of [_recorderUUID description]

### DIFF
--- a/ResearchKit/ActiveTasks/ORKRecorder.m
+++ b/ResearchKit/ActiveTasks/ORKRecorder.m
@@ -180,7 +180,7 @@
 }
 
 - (NSString *)logName {
-    return [NSString stringWithFormat:@"%@_%@", [self recorderType], _recorderUUID];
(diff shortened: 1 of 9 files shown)


ResearchKit/ActiveTasks/ORKRecorder.m
```diff
@@ -180,7 +180,7 @@ - (NSString *)recorderType {
 }
 
 - (NSString *)logName {
-    return [NSString stringWithFormat:@"%@_%@", [self recorderType], _recorderUUID];
```
rsanchezsaez: Should we use the *property syntax* as per the documentation? Otherwise it looks good.

YuanZhu-apple: Sure, will do.

YuanZhu-apple:
Updated to use property syntax.

PR #423, opened 2015-09-02
Fixed the graph view and time label overlapping.

YuanZhu-apple:
This issue happens on iPhone only.
Some how, it works after I removed the constraints in GraphView.

img_0808


umerkhan-apple:
r=m3!

PR #425, opened 2015-09-02
Fix for starting image capture step in split view.

umerkhan-apple:
Added a fix for image capture step if entering the preview step while already in split view.


YuanZhu-apple:
r=me. If it has been tested.

PR #431, opened 2015-09-03
uialertview tintColor fix

matteocrippa:
I noticed that setting a tintColor for ORKConsentReview does not affect the tintColor of the alert.
This commit forces uialertview to use the same tint color of the view in order to avoid color issue

### DIFF
--- a/ResearchKit/Consent/ORKConsentReviewController.m
+++ b/ResearchKit/Consent/ORKConsentReviewController.m
@@ -148,6 +148,8 @@
                                                                    message:self.localizedReasonForConsent
                                                             preferredStyle:UIAlertControllerStyleAlert];
     
+    alert.view.tintColor = self.view.tintColor;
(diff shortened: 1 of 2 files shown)


YuanZhu-apple:
I think it works without your PR.
See this:
screen shot 2015-09-03 at 1 53 12 pm

matteocrippa:
@YuanZhu-apple did you set the tintColor via global tint via storyboard or via code? I'll investigate it and get back with updates.

In the meanwhile I noticed a couple of issue in PDFs generation, sorry but I was editing via github online editor and it managed to list them there instead of a dedicated PR.

YuanZhu-apple:
@matteocrippa To set a tintColor:
```
 [UIView appearance].tintColor = [UIColor greenColor];
```
Sure. Please open issue in PDFs generation.

PR #433, opened 2015-09-03
fix (null) value in pdf

matteocrippa:
Avoid pdf to be rendered with (null) value for CONSENT_DOC_LINE_PRINTED_NAME and CONSENT_DOC_LINE_SIGNATURE

### DIFF
--- a/ResearchKit/Consent/ORKConsentSignatureFormatter.m
+++ b/ResearchKit/Consent/ORKConsentSignatureFormatter.m
@@ -46,11 +46,11 @@
     BOOL addedSig = NO;
 
     NSMutableArray *signatureElements = [NSMutableArray array];
+    NSString *nameStr = @"&nbsp;";
 
     // Signature
     if (signature.requiresName || signature.familyName || signature.givenName) {
         addedSig = YES;
-        NSString *nameStr = @"&nbsp;";
         if (signature.familyName || signature.givenName) {
             NSMutableArray *names = [NSMutableArray array];
             if (signature.givenName) {
@@ -65,7 +65,7 @@
             nameStr = [names componentsJoinedByString:@"&nbsp;"];
         }
 
-        NSString *titleFormat = ORKLocalizedString(@"CONSENT_DOC_LINE_PRINTED_NAME", nil);
+        NSString *titleFormat = [NSString stringWithFormat:ORKLocalizedString(@"CONSENT_DOC_LINE_PRINTED_NAME", nil),nameStr];
         [signatureElements addObject:[NSString stringWithFormat:signatureElementWrapper, nameStr, hr, [NSString stringWithFormat:titleFormat,signature.title]]];
     }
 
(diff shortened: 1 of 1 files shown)


ResearchKit/Consent/ORKConsentSignatureFormatter.m
```diff
@@ -65,7 +65,7 @@ - (NSString *)HTMLForSignature:(ORKConsentSignature *)signature {
             nameStr = [names componentsJoinedByString:@"&nbsp;"];
         }
 
-        NSString *titleFormat = ORKLocalizedString(@"CONSENT_DOC_LINE_PRINTED_NAME", nil);
+        NSString *titleFormat = [NSString stringWithFormat:ORKLocalizedString(@"CONSENT_DOC_LINE_PRINTED_NAME", nil),nameStr];
```
YuanZhu-apple: This make `signature.title` be ignored?

YuanZhu-apple:
`@"CONSENT_DOC_LINE_PRINTED_NAME"` should accept a title not a name.

YuanZhu-apple:
The real fix should be provide a signature.title in the model.

matteocrippa:
I haven't noticed the real idea is to add a text like "(?)'s Signature" instead of forcing the real name/surname of the user written down (sometimes it happens that in some contracts this is used to).

By the way prob the best fix is to check if the string is null and avoid to inject it in the text in order not to have (null) written, I'll give a check to that adjusting the PR according this,do you agree?

YuanZhu-apple:
Generally, showing `Researcher's Signature` is recommended.
Leave it `(NULL)'s Signature`  is easy for developer to catch the issue.

If you really want  "(?)'s Signature", just set `(?)` on `signature.title`.

YuanZhu-apple:
Or shall we throw an exception when `signature.title` is nil?

matteocrippa:
Prob an exception will be better instead of running into a (null) in the pdf once it is rendered

YuanZhu-apple:
Alternatively, make the title `nonnull` in `ORKConsentSignature`?

rsanchezsaez:
Even if the title is annotated to `nonnull`, it may be nice to validate and throw the exception, in case the developer assigns it to a *previously nulled variable* which won't be cached by the compiler.

YuanZhu-apple:
This could apply to all `nonnull` parameters. Don't think we can throw exception on all of them. Right?

rsanchezsaez:
Haha, no. ;-)

But since some developers were encountering PDFs with the `(null)` string on them, maybe this one is worth it to avoid passing that onto users.

YuanZhu-apple:
I just realized not all apps use both `ORKConsentSignature` and `PDF export`. So it is making sense to leave `title` nullable . 
Sure, let's do an `exception`.

YuanZhu-apple:
@matteocrippa Are you going to update this code soon?

matteocrippa:
@YuanZhu-apple hope to try to update during late of this week

matteocrippa:
new ref #481

PR #435, opened 2015-09-04
Add another convenient constructor for ORKQuestionStep

YuanZhu-apple:
The `text` attribute is often missed by developers.


umerkhan-apple:
r=m3

PR #439, opened 2015-09-07
ORKGraphChartView: fix accessibility crash when 'graphChartView:titleForXAxisAtPointIndex:' is not implemented

rsanchezsaez:
This PR fixes a crash when `graphChartView:titleForXAxisAtPointIndex:` (which is an optional method) is not implemented. Unfortunately, this fix needs new localized strings.

### DIFF
--- a/ResearchKit/Charts/ORKGraphChartView.m
+++ b/ResearchKit/Charts/ORKGraphChartView.m
@@ -1066,7 +1066,11 @@ inline static CALayer *graphPointLayerWithColor(UIColor *color) {
             }
         }
         
-        element.accessibilityLabel = [self.dataSource graphChartView:self titleForXAxisAtIndex:pointIndex];
+        if ([_dataSource respondsToSelector:@selector(pieChartView:titleForSegmentAtIndex:)]) {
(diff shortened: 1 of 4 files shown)


ResearchKit/Charts/ORKGraphChartView.m
```diff
@@ -1066,7 +1066,11 @@ - (void)_axCreateAccessibilityElements {
             }
         }
         
-        element.accessibilityLabel = [self.dataSource graphChartView:self titleForXAxisAtIndex:pointIndex];
+        if ([_dataSource respondsToSelector:@selector(pieChartView:titleForSegmentAtIndex:)]) {
+            element.accessibilityLabel = [self.dataSource graphChartView:self titleForXAxisAtPointIndex:pointIndex];
+        } else {
+            element.accessibilityLabel = [[NSString alloc] initWithFormat:ORKLocalizedString(@"AX_CHART_POINT_%@", nil), @(pointIndex).stringValue];
```
YuanZhu-apple: Should use `ORKLocalizedStringFromNumber()` to localize the number.

YuanZhu-apple:
r=me, except number localization. 
Need to get this in now. Will address it in another PR.

PR #441, opened 2015-09-08
Use `ORKLocalizedStringFromNumber()`

YuanZhu-apple:
Thanks @rsanchezsaez 's  #439 
This is a following up PR to fix a minor issue.


YuanZhu-apple:
R=me, and tested.

PR #442, opened 2015-09-08
Charts: update ORKCatalog to use new data source method (convergence)

rsanchezsaez:
Sorry, I forgot to update `ORKCatalog` as well.


YuanZhu-apple:
Yes!

Issue #421, opened 2015-09-01
Add 'optional' property to form items


rsanchezsaez:
As discussed on the [mailing list](http://lists.apple.com/archives/researchkit-users/2015/Aug/msg00018.html), adding an `optional` property to `ORKFormItem` would be desirable.

The *form step* would enable the *Done/Next* button after all the *non-optional form items* have been completed by the user. If all the contained *form items* are optional, the *Done/Next* button is enabled at all times, even when the *form step* is non-optional.

This is to allow researchers to get results from non-optional *form steps*, which could then be partially completed by the user (since skipping an optional *form step* purposefully nils all the currently entered data).

rsanchezsaez:
This has been implemented and merged by the referenced pull request, closing issue.

PR #443, opened 2015-09-08
ORKFormItem: implement 'optional' property

rsanchezsaez:
Implements `ORKFormItem`'s `optional` property ([Issue #421](https://github.com/ResearchKit/ResearchKit/issues/421)). Also adds corresponding `ORKCatalog` example.

Now the *Continue/Done* button of form steps, irregardless of them being optional, only enables if:
- At least one form item has an answer.
- All answered form items are valid.
- All the non-optional form items have answers.

The previous behaviour was that the *Continue/Done button* was enabled for *optional form steps* if at least one question was answered; and was enabled for *non-optional form step* if all answers were answered. I changed this behaviour for homogeneity between forms, and because the implemented *form step optionality* provides maximum flexibility.

The *Skip button* behaviour remains unchanged: show it for *optional form steps* and hide it for *non-optional form steps*.

### DIFF
--- a/ResearchKit/Common/ORKFormStep.h
+++ b/ResearchKit/Common/ORKFormStep.h
@@ -46,8 +46,15 @@ NS_ASSUME_NONNULL_BEGIN
  When the task completes, the user's answers are encoded in the result hierarchy
  in the task view controller.
  
- Each question in the form is represented by an `ORKFormItem` object. The form
- can be broken into sections by using an `ORKFormItem` object that includes only a section title.
+ Each question in the form is represented by an `ORKFormItem` object. The form items have an
+ `optional` property that defaults to `YES`. All non-optional questions need to be answered for the
+ Continue button to be enabled. If all the form items are optional, at least one question needs to
+ be answered for the Continue button to be enabled. You can allow the user to completely skip a
+ form step using the Skip button, even if it has non-optional form items, by setting the form step
+ `optional` property to yes.
+ 
+ The form can be broken into sections by using an `ORKFormItem` object that includes only a section
+ title.
  
  The result of a form step is an `ORKStepResult` object that includes a child `ORKQuestionResult`
  object for each form item.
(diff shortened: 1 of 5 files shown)


ResearchKit/Common/ORKFormStep.h
```diff
@@ -129,6 +129,14 @@ ORK_CLASS_AVAILABLE
 @property (nonatomic, copy, readonly) NSString *identifier;
 
 /**
+ A Boolean value indicating whether the form item is optional.
+ 
+ The default value of this property is `NO`. When the value is `YES`, this form item doesn't need
```
YuanZhu-apple: Shall we make default value to be `YES`? 
To align with `ORKStep`'s `optional`.

rsanchezsaez: I was thinking that defaulting it to `NO` was more coherent with the previous behaviour where all *form items* were non-optional.

But let me know if you prefer it to default to `YES` and I'll change it.

YuanZhu-apple: I see your point. 

In this PR we need fill a matrix:

| Step         | FormItem           | Require to answer?  |
| ------------- |:-------------:| -----:|
| optional    | not-optional | ? |
| optional    | optional      |   NO  |
| not-optional    | not-optional     |  YES  |
| not-optional    | optional      |   ? |

And since we have the granularity control for each `formItem`, shall we ignore `optional` setting on step level in a form step? If so, default value `YES` make more sense.

rsanchezsaez: That's almost right.

Basically, the `optional` setting in a *form step* only controls if the *Skip button* shows or not (skipping a form step nullifies all previously provided answers, so it's a slightly different use case).

Regardless of the *form step* optionality, then we have:

FormItems  | *Done button* enabled
------------- | -------------
All non-optional  | If all items are answered
Mixed  | If all non-optional items are answered
All optional | If at least one item is answered

Tapping the *Done button* does not nullify any of the provided answers.

YuanZhu-apple: Thanks for the explanation.

Before this PR, the formItems are `All optional` or `All non-optional` controlled by form step optionality. (Not anymore in this PR)

So regardless the default value `YES` or `NO` we give to formItem optionality, we cannot make all existing behavior unchanged in the existing apps.

Like I said previously, to align with ORKStep's optionality for questionStep, make default value `YES` is better. Agree?

rsanchezsaez: Yes, the only detail is that even if you set all *form items* as non-optional, they can still be overridden by using the *Skip button* if the *form step* is optional. This part is equivalent to what happened before.

But yeah, maybe setting the *form item optional default* value to `YES` is more homogenous, I'll do just that.

YuanZhu-apple:
Looks perfect! r=me.

PR #445, opened 2015-09-09
[ORKTextField] Coherent placeholder handling

rsanchezsaez:
This PR makes the handing of the placeholder in `ORKTextField` more similar to what `UITextField` does.

It comprises of two changes:
- Remove extra spacing between unit number and placeholder text. This makes the space between the unit and the placeholder to be consistent between *editing* and *non editing* modes.
- Do not hide placeholder when the editing text is empty (same as with `UITextField`).

Also reuses form item cells in `ORKStepViewController` so cells are not created twice when calculating their height.

---

Before: 

simulator screen shot 9 sep 2015 02 46 13
simulator screen shot 9 sep 2015 02 59 19
simulator screen shot 9 sep 2015 02 46 20

---

After:

simulator screen shot 9 sep 2015 03 05 42
simulator screen shot 9 sep 2015 03 05 47
simulator screen shot 9 sep 2015 03 05 53


umerkhan-apple:
Did we want to extend this same functionality to the multiline textfields? For example, check out the *Message* form item under *Mini Form* in *ORKTest*.

rsanchezsaez:
@umerkhan-apple: I'll look into this in the following days.

umerkhan-apple:
@rsanchezsaez So did you get a chance to look at multiline textfields? Do we want to extend this functionality to them, or should we just leave it to single line textfields?

rsanchezsaez:
Hey Umer, 

Sorry, I missed the GitHub notification earlier.

I agree that the *placeholder* on *text views* (*multiline text fields*) should behave like the *text field* one. I'm looking into this.

rsanchezsaez:
@umerkhan-apple: Done, I have implemented the same placeholder behavior for `ORKAnswerTextView`.

umerkhan-apple:
@rsanchezsaez Awesome, I will test the changes shortly.

ResearchKit/Common/ORKAnswerTextView.m
```diff
@@ -30,44 +31,83 @@ OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
...
     return self;
 }
 
-- (void)init_ORKAnswerTextView {
+- (void)commonInit {
     [[NSNotificationCenter defaultCenter] addObserver:self
                                              selector:@selector(updateAppearance)
                                                  name:UIContentSizeCategoryDidChangeNotification
                                                object:nil];
+    [[NSNotificationCenter defaultCenter]  addObserver:self selector:@selector(textViewTextDidChange:)
```
umerkhan-apple: Move selector argument onto a new line.

rsanchezsaez:
Adding the placeholder to the `ORKAnswerTextView` was tricky.

First, I tried using a plain `UILabel` but getting the positioning to be coherent with the typed in text was not easy.

So I settled to adding a non-interactive `UITextView` as a child of `ORKAnswerTextView`. For this to layout correctly on form steps, I needed to both set the frame manually (on layout subviews) and add constraints. I find this very weird because normally you either set the frame or use constraints, but never both. Removing one or the other makes it to stop being laid out correctly. If you have a suggestion for cleanest implementation, I'm all ears.

ResearchKit/Common/ORKAnswerTextView.m
```diff
@@ -30,44 +31,83 @@ OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
...
+    [self insertSubview:_placeholderTextView atIndex:0];
     
+    [self setUpConstraints];
     
     [self updateAppearance];
 }
 
+- (void)layoutSubviews {
+    [super layoutSubviews];
+    // Setting the frame directly causes a layout error on a form step (it looks like an iOS bug, as setting the frame should produce the same effect as settings the bounds and the center)
```
umerkhan-apple: Typo: settings -> setting

umerkhan-apple:
@rsanchezsaez I agree that this is a bit of a tricky implementation. I will look into finding a cleaner approach if possible.

Currently, the framework is crashing if I change orientation from Portrait to Landscape inside `ORKTest` running `Mini Form` task.

##### Console log
> 2015-10-20 11:16:55.377 ORKTest[614:106204] -[NSIndexPath setExpectedLayoutWidth:]: unrecognized selector sent to instance 0x175dfda0 

_Tested on iPhone 5 and iPhone 6S._

rsanchezsaez:
I've fixed the crash, that was an unrelated silly mistake on my end.

Let me know if you come up with an alternate solution for the *text view placeholder*.

YuanZhu-apple:
screen shot 2015-10-21 at 2 54 21 pm

Not form your code, I just noticed the textview is misaligned in formStep. Can you fix it in this PR, thanks!

ResearchKit/Common/ORKTextFieldView.m
```diff
@@ -185,7 +184,7 @@ - (void)ork_setSuffix:(NSString *)suffix withColor:(UIColor *)color {
     if (suffix.length == 0) {
         return;
     }
-    _suffixLabel = [self ork_createTextLabelWithTextColor:color ?: [UIColor grayColor]];
+    _suffixLabel = [self ork_createTextLabelWithTextColor:(color ?: [UIColor ork_midGrayTintColor])];
```
YuanZhu-apple: Intention of this change?

rsanchezsaez: This color is sometimes used for the *unit suffix label* when editing an `ORKTextFieldView` with no placeholder. You can see this if you revert this change, tap on the first text field of the first step on *Mini Form*, tap on the second field, and then tap on the first field again (the unit suffix label turns darker).

`ork_midGrayTintColor` is used as the color in all other `placeholder`s and *unit suffix* instances, so I figured it should be used here too.

But maybe we should also review the actual color-chosing logic, because it's weird that this color only gets used the second time you tap on the field but not the first one.

rsanchezsaez:
I'll later look into the alignment of the *text view*.

umerkhan-apple:
@rsanchezsaez After internal discussion, we have come to the following conclusion:
- When there is a unit specified on the text field, keep the behavior as it is in current RK.
- When there is no unit specified (basically, when it is functioning like a basic uitextfield) then show the placeholder until editing begins to stay consistent with iOS's uitextfield (your PR's implementation).

rsanchezsaez:
Ok. But why do that for first the case? Don't you think keeping the placeholder until some text is typed is more coherent and consistent?

umerkhan-apple:
The main reason for that conclusion was that having two placeholders seemed untidy and slightly confusing for the user.

rsanchezsaez:
But, in my opinion, the user perception is that, when editing an empty unit-suffixed *text field*, the unit label becomes part of the placeholder text and then, when they start typing, the "long placeholder" transforms to show only the unit suffix. I think this worked well after reducing the placeholder <> unit separation (see the before/after screenshots above). Don't you think?

If you remain unconvinced let me know and I'll update it to the behavior you requested. :)

YuanZhu-apple:
@rsanchezsaez I think the original set a more clear distinction between `placeholder` and `unit string`, see the margin between them when the textfield is inactive.

And when the textfield become active, showing both of them makes the field a little noisy.

[Original] 
screen shot 2015-10-21 at 3 30 19 pm
screen shot 2015-10-21 at 3 24 48 pm


[This PR] 
screen shot 2015-10-21 at 3 32 20 pm
screen shot 2015-10-21 at 3 15 17 pm

rsanchezsaez:
I have updated this PR against the latest master, with the requested behavior:

- When there is a *suffix unit* specified on the text field, keep the behavior as it was.
- When there is no *suffix unit* specified, show the placeholder until the user writes text, to stay consistent with *iOS*'s `UITextField`.

I also added fixed a regression in which the user was being prevented leaving a form text field blank (a validation alert was being shown, you can try in `ORKTest`'s *Mini Form* by editing the first field and trying to move to the second without typing anything).

Let me know if you want additional changes. I think the animation glitch and the suffix color mismatch fix are worthwhile improvements as well.

rsanchezsaez:
Also, the *TextView* misalign shown in the above comment screenshot no longer seems to be an issue. Maybe *iOS 9.2* fixed it? I don't see any RK code change related to that.

YuanZhu-apple:
R=ME:) @umerkhan-apple Can you take a look as well?

YuanZhu-apple:
@umerkhan-apple Get a chance to try this?

rsanchezsaez:
@umerkhan-apple Ping? ;-)

PR #448, opened 2015-09-10
Add a single workspace that includes RK,  ORKTest, and ORKCatalog for convenience

YuanZhu-apple:
Save developer's effort in switching between project files.

### DIFF
--- a/RKWorkspace.xcworkspace/contents.xcworkspacedata
+++ b/RKWorkspace.xcworkspace/contents.xcworkspacedata
@@ -1,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Workspace
+   version = "1.0">
+   <FileRef
+      location = "group:ResearchKit.xcodeproj">
+   </FileRef>
+   <FileRef
+      location = "group:Testing/ORKTest/ORKTest.xcodeproj">
+   </FileRef>
+   <FileRef
+      location = "group:samples/ORKCatalog/ORKCatalog.xcodeproj">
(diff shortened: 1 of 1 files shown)


YuanZhu-apple:
@rsanchezsaez What do you think?

rsanchezsaez:
This is very convenient indeed. I was constantly opening and closing `ORKCatalog` and `ORKTest` before. ;-)

I wonder if we should remove the `ResearchKit` subproject from the individual `ORKCatalog` and `ORKTest` projects. But maybe not, because it doesn't seem to have any ill effects, and that way those can still be opened individually without using the *workspace*.

All targets and tests work ok when run from the workspace.

YuanZhu-apple:
Cool!

rsanchezsaez:
@YuanZhu-apple: A small nit I missed: wouldn't `ResearchKit.xcworkspace` be a better name than `RKWorkspace.xcworkspace`? Adding *filetype information* to a file name doesn't feel right.

umerkhan-apple:
@rsanchezsaez I think RKWorkspace is a bit more definitive, since the `ResearchKit` title is used for the framework already. We do not require the entire workspace to be included in order to use ResearchKit, so this helps differentiate between the workspace and the actual framework.

rsanchezsaez:
@umerkhan-apple That makes sense.

I was also thinking about a names like `RKSamples` instead of  `RKWorkspace`, but it's not ideal either because `ORKCatalog` is a sample, but `ORKTest` is more of a manual test suite.

Issue #256, opened 2015-06-02
Revise logging implementation


jwe-apple:
ORK_Log_Debug and ORK_Log_Oops need some control developers can exert besides using DEBUG / NDEBUG.

p2:
+1 from me! I've changed the flag to `ORK_DEBUG` in my checkouts just to silence all the view-did-appear logs which obscure my own logging.

rsanchezsaez:
I'll be working on this.

rsanchezsaez:
@p2: PR submitted. Any feedback is welcome.

rsanchezsaez:
This can be closed.

PR #449, opened 2015-09-10
Revamp logging system

rsanchezsaez:
This PR deals with [Issue #256](https://github.com/ResearchKit/ResearchKit/issues/256).

It adds four logging levels:
- `ORK_LOG_LEVEL_NONE`
- `ORK_LOG_LEVEL_DEBUG`, `ORK_Log_Debug()`
- `ORK_LOG_LEVEL_WARNING`, `ORK_Log_Warning()`
- `ORK_LOG_LEVEL_ERROR`, `ORK_Log_Error()`

Also:
- Changes some (mildly) serious error messages from the `DEBUG` to the `WARNING` level. Please let me know if any of these promotions should be avoided, or if any of those should be promoted to the `ERROR` level.
- Remove the line number from logging as it's noisy (we already have the function name).

I figured these three levels are more than enough, but let me know if you prefer to have more levels such as  `Emergency`, `Alert`, `Critical`, `Notice` or `Info`, as explained [here](https://developer.apple.com/library/mac/documentation/MacOSX/Conceptual/BPSystemStartup/Chapters/LoggingErrorsAndWarnings.html#//apple_ref/doc/uid/10000172i-SW8-SW3).

---

This is the documentation, which I added as a new section to the *ResearchKit Framework Programming Guide*:

## Logging Errors and Warnings

The ResearchKit framework supports four log levels, controlled by four preprocessor macros and their corresponding *`NSLog()`-like* logging macros:
* `ORK_LOG_LEVEL_NONE`
* `ORK_LOG_LEVEL_DEBUG`, `ORK_Log_Debug()`
* `ORK_LOG_LEVEL_WARNING`, `ORK_Log_Warning()`
* `ORK_LOG_LEVEL_ERROR`, `ORK_Log_Error()`

Setting the ResearchKit's framework `ORK_LOG_LEVEL_NONE` macro to `1` completely silences all ResearchKit logs, overriding any other specified log level. Setting `ORK_LOG_LEVEL_DEBUG`, `ORK_LOG_LEVEL_WARNING`, or `ORK_LOG_LEVEL_ERROR` to `1` enables logging at that level and at those of higher seriousness.

If you do not explicitly set a log level, `ORK_LOG_LEVEL_WARNING=1` is used by default.

You have to set any of these preprocessor macros in your ResearchKit subproject, not in your main project. Within *Xcode*, you can do so by setting any of them in the `Preprocessor Macros` list on the `Build Settings` of your `ResearchKit` framework target.

See these resources if you are using ResearchKit through CocoaPods and need to change the log level: [[1]](http://stackoverflow.com/a/30038120/269753) [[2]](http://www.mokacoding.com/blog/cocoapods-and-custom-build-configurations/).


p2:
I greatly appreciate having control over RK‘s log level. Not sure about removing the line number though, usually file name+line number is what I orient myself after, not so much the method name. But not a biggie either.

rsanchezsaez:
I can think of two main use cases for the logging system:
- A *ResearchKit framework* developer using the `DEBUG` level to check if everything is working as expected. In that case, she just added several `ORK_Log_Debug()` lines, so the code that is being changed is fresh on her mind, or already open in *Xcode*..
- A *ResearchKit* user opens the console and looks for `WARNING` or `ERROR` messages to file a bug report or to know what's happening. It's not frequent that his user will dive into RK's code to make changes or to look at the line where that happened.

In both of these cases, I find the line number a bit superfluous and distracting. Usually, RK's methods are pretty small: that's why I thought the method name was enough. I only log the method name I my other projects as well. ;-)

p2:
Different preferences I guess; I only log file name and line number. :) In this case we should definitely go with your preference, given all your work already and me not having a solid argument, other than taste.

ResearchKit/Common/ORKHelpers.h
```diff
@@ -34,6 +35,37 @@
...
 
 
+#if ( defined(ORK_LOG_LEVEL_NONE) && ORK_LOG_LEVEL_NONE )
+#  undef ORK_LOG_LEVEL_DEBUG
+#  undef ORK_LOG_LEVEL_WARNING
+#  undef ORK_LOG_LEVEL_ERROR
+#endif
+
+#if ( !defined(ORK_LOG_LEVEL_NONE) && !defined(ORK_LOG_LEVEL_DEBUG) && !defined(ORK_LOG_LEVEL_WARNING) && !defined(ORK_LOG_LEVEL_ERROR) )
+#  define ORK_LOG_LEVEL_WARNING 1
```
YuanZhu-apple: I could be wrong, `ORK_LOG_LEVEL_WARNING` is the default value (fallback) ? If the developer leave `Preprocessor Macro` blank? For example in the release config? 
screen shot 2015-09-23 at 2 00 49 pm

rsanchezsaez: That is correct, if none of the *log levels* is specified, it defaults to `ORK_LOG_LEVEL_WARNING`. I think this is saner than defaulting to `ORK_LOG_LEVEL_LOG` or to `ORK_LOG_LEVEL_NONE`.

I have detailed this on the documentation.

What do you think?

YuanZhu-apple: Thanks, use `ORK_LOG_LEVEL_WARNING` as the default is good. 
Shall we update RK's project file to include our recommended configurations: 

Debug | ORK_LOG_LEVEL_WARNING=1
Release | ORK_LOG_LEVEL_NONE=1

rsanchezsaez: Should we default `Release` to `ORK_LOG_LEVEL_NONE`? I think it's kind of nice to default `Release` to *warnings*, or at least to *errors*. Other Apple frameworks are quite verbose on the console log regarding potentially dangerous situations.

Current *ResearchKit* warnings are quite sparse, and they can be potentially important for the developer or the user to see.

YuanZhu-apple: Can the developers retrieve those logs from shipping apps? The user won't be able to see the console logs.

rsanchezsaez: Well, a user guided by the developer could use `Xcode -> Devices` to see the *iPhone* console, where logs from shipping apps do show.

YuanZhu-apple: Only the `crash logs` can be seen, right? I cannot find general console log.

rsanchezsaez: No, within *Xcode* you can see the full real-time console of a connected device:

screen shot 2015-09-23 at 22 52 50

YuanZhu-apple: Learned:)
You are right, the first party apps are leaving log on for `warnings` and `errors`.
So let's make it explicit in the project file?

Debug | ORK_LOG_LEVEL_WARNING=1
Release | ORK_LOG_LEVEL_WARNING=1

rsanchezsaez: Sure, added. ;-)

YuanZhu-apple:
r=me,

PR #457, opened 2015-09-14
Fix `blank` consent html review page.

YuanZhu-apple:
To reproduce:
1. Run ORKTest
2. Select Consent
3. Navigate to consent review step

Somehow `updateViewConstraints` is not being called. Have to force it.
Also added  super call to `viewWillTransitionToSize`

 
screen shot 2015-09-14 at 1 59 28 pmscreen shot 2015-09-14 at 1 59 11 pm

### DIFF
--- a/ResearchKit/Consent/ORKConsentReviewController.m
+++ b/ResearchKit/Consent/ORKConsentReviewController.m
@@ -87,6 +87,8 @@
     
     [self.view addSubview:_webView];
     [self.view addSubview:_toolbar];
+    
+    [self.view setNeedsUpdateConstraints];
 }
 
 - (void)updateLayoutMargins {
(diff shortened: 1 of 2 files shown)


YuanZhu-apple:
@rsanchezsaez Can you help me review this one?
I am also considering cherry picking this patch to stable branch, because it is quite critical.

rsanchezsaez:
Reviewed. Reproduced bug without the patch, the issue looks fixed after applying the patch.

Code looks good: it makes sense to call `setNeedsUpdateConstraints` after adding additional subviews from the view controller.

YuanZhu-apple:
Thanks!

YuanZhu-apple:
It is weird that without the patch, `Consent Review` task works fine.

rsanchezsaez:
@YuanZhu-apple Yeah, that’s quite strange.

It seems that, starting on iOS 8, `updateViewConstraints` isn’t always called after the view loads. I wonder why it’s called in one case but not in the other. Maybe an iOS-level *autolayout* bug?

Issue #455, opened 2015-09-14
Animations for Consent Sections not playing after leaving visual consent step


jamiedaniel:
To replicate:

1. Run ORKTest
2. Select Consent
3. Navigate to consent review step
4. Navigate backward 2-4 times
5. Navigate forward and notice that the animation in not visible but played the correct amount of time.

YuanZhu-apple:
@umerkhan-apple is investigating.

YuanZhu-apple:
Fixed by #459

PR #459, opened 2015-09-16
Fix consent animation.

YuanZhu-apple:
Fix for #455

Root cause: old playerView's dealloc happens on the new EAGLContext(wrong) of new playerView.

Save/Restore EAGLContext in each of the methods which do GL drawing.
This make more than one GLContext lives together.

### DIFF
--- a/ResearchKit/Consent/ORKEAGLMoviePlayerView.m
+++ b/ResearchKit/Consent/ORKEAGLMoviePlayerView.m
@@ -93,6 +93,7 @@ static const GLfloat kColorConversion709[] = {
     GLint _backingHeight;
     
     EAGLContext *_context;
+    NSMutableArray *_contextStack;
     CVOpenGLESTextureRef _lumaTexture;
     CVOpenGLESTextureRef _chromaTexture;
     CVOpenGLESTextureCacheRef _videoTextureCache;
@@ -144,7 +145,7 @@ const GLfloat DefaultPreferredRotation = 0;
         // Set the context into which the frames will be drawn.
         _context = [[EAGLContext alloc] initWithAPI:kEAGLRenderingAPIOpenGLES2];
         
-        if (!_context || ![EAGLContext setCurrentContext:_context] || ![self loadShaders]) {
+        if (!_context || ![self loadShaders]) {
             return nil;
         }
         
@@ -161,7 +162,8 @@ const GLfloat DefaultPreferredRotation = 0;
     }
     _glIsSetup = YES;
 
-    [EAGLContext setCurrentContext:_context];
+    [self saveGLContext];
+    
     glDisable(GL_DEPTH_TEST);
     [self setupBuffers];
     
@@ -189,6 +191,7 @@ const GLfloat DefaultPreferredRotation = 0;
     glGenVertexArraysOES(1, &_vertexArrayHandle);
     glGenBuffers(1, &_vertexBufferHandle);
 
+    [self restoreGLContext];
     ORKEAGLLog(@"");
 }
 
@@ -198,6 +201,8 @@ const GLfloat DefaultPreferredRotation = 0;
     if (!_glIsSetup) {
         return;
     }
+    
+    [self saveGLContext];
     
     [self deleteBuffers];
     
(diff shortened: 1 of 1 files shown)


YuanZhu-apple:
@jwe-apple Can you review? Thanks!

rsanchezsaez:
The fix looks sensible and it works great.

But, wouldn't just using a `_currentContext` variable instead of a *stack* be simpler? We could assume that the context is correct in inner methods like `compileShader`, `linkProgram`, `cleanUpTextures`, etc., and don't save/restore the context on those (since they'd always be called from somewhere else which has an outer context save/restore) .

I'm conflicted: what you did is probably safer and more comprehensive, but it makes avoidable calls to `saveGLContext`/`restoreGLContext`. ;-)

YuanZhu-apple:
My solution is very like: `CGContextSaveGState` and `CGContextRestoreGState`.  
There are avoidable calls, but this code is easy to maintain(no need to keep track which one is inner method).

rsanchezsaez:
Yeah, you're right, I'm now convinced that your solution is better. I'm all-in for simplicity over premature optimizations any time. The extra calls for maintaining the stack should be negligible.

I was kind of thinking loudly (because I had seen the local variable solution before).

ResearchKit/Consent/ORKEAGLMoviePlayerView.m
```diff
@@ -580,14 +662,18 @@ - (BOOL)compileShader:(GLuint *)shader type:(GLenum)type URL:(NSURL *)URL {
     
     glGetShaderiv(*shader, GL_COMPILE_STATUS, &status);
     if (status == 0) {
+        [self restoreGLContext];
```
jwe-apple: Can we rearrange this slightly to avoid the early return? Better to have a single exit path.

ResearchKit/Consent/ORKEAGLMoviePlayerView.m
```diff
@@ -527,7 +606,7 @@ - (BOOL)loadShaders {
             glDeleteProgram(_programHandle);
             _programHandle = 0;
         }
-        
+        [self restoreGLContext];
```
jwe-apple: Also arrange this to avoid the early return if we can?

ResearchKit/Consent/ORKEAGLMoviePlayerView.m
```diff
@@ -246,10 +258,57 @@ - (void)cleanUpTextures {
...
+
+- (BOOL)restoreGLContext {
+    BOOL success = YES;
+    
+    id lastObject = [_contextStack lastObject];
+    
+    if (lastObject) {
+        EAGLContext *contextToBeRestored = (lastObject != [NSNull null]) ? lastObject : nil;
+        
+        // Switch context only when necessory
```
jwe-apple: spelling

ResearchKit/Consent/ORKEAGLMoviePlayerView.m
```diff
@@ -246,10 +258,57 @@ - (void)cleanUpTextures {
...
+    EAGLContext *currentContext = [EAGLContext currentContext];
+    
+    // Switch context only when necessory
+    if (_context != currentContext) {
+        glFlush();
+        success = [EAGLContext setCurrentContext:_context];
+    }
+    
+    // Always push
+    [_contextStack addObject:currentContext? : [NSNull null]];
```
jwe-apple: spacing

ResearchKit/Consent/ORKEAGLMoviePlayerView.m
```diff
@@ -246,10 +258,57 @@ - (void)cleanUpTextures {
     
     // Periodic texture cache flush every frame
     CVOpenGLESTextureCacheFlush(_videoTextureCache, 0);
+    
+    [self restoreGLContext];
+}
+
+- (BOOL)saveGLContext {
```
jwe-apple: I'm not sure I like "save" as the name here. "Save" normally means you don't change anything - but here "save" has the side effect of setting the current context.

Could we instead call these push / pop? e.g. pushGLContext / popGLContext

YuanZhu-apple:
Thanks @jwe-apple

PR #474, opened 2015-09-23
ORKTest: fix column width not being updated after rotation

rsanchezsaez:
This PR fixes the column width not being updated after rotation in `ORKTest` (issue originally reported in [PR #343](https://github.com/ResearchKit/ResearchKit/pull/343).

It also makes the scroll view to use 3 columns for view widths equal or greater than the *iPhone 6* in landscape mode (*667 points*).


YuanZhu-apple:
r=me. Thanks!

PR #483, opened 2015-09-30
Make the test for ORKESerialization to catch unregistered classes.

YuanZhu-apple:
Also added missed class registrations.


YuanZhu-apple:
@umerkhan-apple This one, thanks!

umerkhan-apple:
r=m3! :+1:

Issue #482, opened 2015-09-29
Audio task layout bug


umerkhan-apple:
The Audio Task in ORKTest and ORKCatalog has a constraint bug that causes the contents to be shifted to the left side of the screen for Step 4. (Screenshot attached).

![simulator screen shot sep 28 2015 5 22 03 pm](https://cloud.githubusercontent.com/assets/13040996/10152305/95ee6fdc-6605-11e5-9f42-a21dc4fc8e02.png)

PR #486, opened 2015-09-30
Fix audio task layout problem

YuanZhu-apple:
For #482, fixed by removing a constraint.

screen shot 2015-09-30 at 11 21 03 am

### DIFF
--- a/ResearchKit/ActiveTasks/ORKAudioContentView.m
+++ b/ResearchKit/ActiveTasks/ORKAudioContentView.m
@@ -74,14 +74,6 @@ static const CGFloat ValueLineMargin = 1.5;
 }
 
 - (void)setUpConstraints {
-    NSLayoutConstraint *widthConstraint = [NSLayoutConstraint constraintWithItem:self
-                                                                       attribute:NSLayoutAttributeWidth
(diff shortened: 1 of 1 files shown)


matteocrippa:
@YuanZhu-apple I was looking at this issue when I noticed that you already fixed it.
What do you think to move the counter in the middle center under the progress? Do you think isn't better for the UX to have that information in the center of the view instead of on the right?

YuanZhu-apple:
@matteocrippa The lastest amplitude are coming in from right side, which is next to the `time label`. Current design is good for user who monitor his/her performance and check the time at the same time.

YuanZhu-apple:

umerkhan-apple:
r=m3! :+1:

rsanchezsaez:
@YuanZhu-apple: I think a better solution would be to restore the deleted constant, and use `ORKScreenMetricMaxDimension` instead of `CGFLOAT_MAX` (in both the *width* and *height* `ORKAudioGraphView` constraints). That fixes the issue.

Using `CGFLOAT_MAX`, although not too reasonable, wasn't a problem on *iOS 8*. I guess *iOS 9* is more sensitive to this boundary value.

Sorry to comment after it was merged! I was just doing some experimentation on this.

Issue #495, opened 2015-10-01
Localization fallback language


umerkhan-apple:
Localization does not have any fallback language set.
If translations are not available, it should automatically revert to the U.S. English.
Currently, it shows the localization key instead of the U.S. English's value for the key.

LukeJamesFitzpatrick:
This comment on Stack Over Flow should help you out about default localization language fallback http://stackoverflow.com/questions/3263859/localizing-strings-in-ios-default-fallback-language 

Also, you can use a service like http:www.localizer.co to localize your website / app, setup is simple, post a string of code into your web app, and it's done.

PR #498, opened 2015-10-01
IOI # 495 - Localization fallback language

umerkhan-apple:
#### Localization fallback language
Currently, ORKLocalizedString() returns the key as a string if a value is not found for that key.
> For example, if we are in the German language setting, then it will display "PASSCODE_PROMPT_MESSAGE".

This PR redirects ORKLocalizedString() to find the value for the key in U.S. English and return that instead. If there is no value in the U.S. English either, then it will just return the key.
> With this update, if we are in the German language setting, then it will display "Enter passcode", if there is no translation available in German.

Issue #495

### DIFF
--- a/ResearchKit/Common/ORKHelpers.m
+++ b/ResearchKit/Common/ORKHelpers.m
@@ -262,7 +262,8 @@ NSBundle *ORKDefaultLocaleBundle() {
     
     static dispatch_once_t onceToken;
     dispatch_once(&onceToken, ^{
-        __bundle = [NSBundle bundleWithPath:[ORKBundle() pathForResource:@"en" ofType:@"lproj"]];
+        NSString *path = [ORKBundle() pathForResource:[ORKBundle() objectForInfoDictionaryKey:@"CFBundleDevelopmentRegion"] ofType:@"lproj"];
+        __bundle = [NSBundle bundleWithPath:path];
     });
     
     return __bundle;
--- a/ResearchKit/Common/ORKHelpers.m
+++ b/ResearchKit/Common/ORKHelpers.m
@@ -247,14 +247,25 @@ NSDateFormatter *ORKTimeOfDayLabelFormatter() {
 }
 
 NSBundle *ORKBundle() {
-    NSBundle *bundle = [NSBundle bundleForClass:[ORKStep class]];
-    return bundle;
+    static NSBundle *__bundle;
+    
+    static dispatch_once_t onceToken;
+    dispatch_once(&onceToken, ^{
+        __bundle = [NSBundle bundleForClass:[ORKStep class]];
+    });
+    
+    return __bundle;
 }
 
 NSBundle *ORKDefaultLocaleBundle() {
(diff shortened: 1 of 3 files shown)


ResearchKit/Common/ORKHelpers.m
```diff
@@ -251,6 +251,12 @@ void ORKEnableAutoLayoutForViews(NSArray *views) {
     return bundle;
 }
 
+NSBundle *ORKDefaultLocaleBundle() {
+    NSString * path = [ORKBundle() pathForResource:@"en" ofType:@"lproj"];
+    NSBundle * bundle = [NSBundle bundleWithPath:path];
```
YuanZhu-apple: Cache this bundle?

YuanZhu-apple:
r=m3! With one suggestion.

ResearchKit/Common/ORKHelpers.m
rsanchezsaez: Wouldn't it be nicer to get the default localization (`en`) by using `[[NSBundle mainBundle] objectForInfoDictionaryKey:@"CFBundleDevelopmentRegion"]` rather than hardcoding it?

Also, remove spaces between `*` and variable name.

YuanZhu-apple: Like the idea to use `objectForInfoDictionaryKey:@"CFBundleDevelopmentRegion"`. But this is for RK's localization, we might want to use `ORKBundle` instead of `mainBundle`.

rsanchezsaez: Good point.

oliverschaefer:
@umerkhan-apple Nice. Could we integrate a mechanism to specify a specific locale to be used? Maybe via a proxy approach similar to UIView? That would allow to change user facing strings at runtime, which would be helpful. Certainly this is an issue of its own, just wanted to mention it here since you were actively dealing with it at the moment :-)

umerkhan-apple:
@YuanZhu-apple done!

oliverschaefer:
@umerkhan-apple I've opened PR #500 and added some code to your localization fix, to allow for changing the locale at runtime.

YuanZhu-apple:
Great.

ResearchKit/Common/ORKHelpers.m
```diff
@@ -247,8 +247,26 @@ void ORKEnableAutoLayoutForViews(NSArray *views) {
 }
 
 NSBundle *ORKBundle() {
-    NSBundle *bundle = [NSBundle bundleForClass:[ORKStep class]];
-    return bundle;
+    static NSBundle *__bundle;
```
rsanchezsaez: Shouldn't we avoid the *double underscore prefix*? It's normally [reserved](http://stackoverflow.com/a/10687114/269753) in *C* and *C++*.

It's used on several variable names on `ORKHelpers.m`, we could review the file and remove them.

YuanZhu-apple: Agree.

PR #512, opened 2015-10-07
RestorationData

md0u80c9:
Amended ORKTaskViewController initWithTask to take a nullable restorationData field.
Updated ORKTest to allow us to test this out.
tested with NavigableOrderedTask.


ResearchKit/Common/ORKTaskViewController.m
```diff
@@ -302,7 +302,7 @@ - (instancetype)initWithTask:(id<ORKTask>)task restorationData:(NSData *)data de
     
     self = [self initWithTask:task taskRunUUID:nil];
     
-    if (self) {
+    if (self && (data != nil)) {
         self.delegate = delegate;
```
YuanZhu-apple: Should accept `delegate` regardless of restoration. 
Shall we move `self.delegate = delegate;` out of `if block`?

md0u80c9:
Agreed! Come to think of it - shouldn't there be a convenience initialiser for taskRunUUID which accepts delegate as a parameter too?

ResearchKit/Common/ORKTaskViewController.m
```diff
@@ -304,10 +304,13 @@ - (instancetype)initWithTask:(id<ORKTask>)task restorationData:(NSData *)data de
     
     if (self) {
         self.delegate = delegate;
-        self.restorationClass = [self class];
-        NSKeyedUnarchiver *unarchiver = [[NSKeyedUnarchiver alloc] initForReadingWithData:data];
-        [self decodeRestorableStateWithCoder:unarchiver];
-        [self applicationFinishedRestoringState];
+        if (data != nil) {
+            self.delegate = delegate;
```
YuanZhu-apple: Remove this extra assignment.

YuanZhu-apple:
`delegate` can play a very important role in the restoration process (`applicationFinishedRestoringState`) during init. I think we can leave out `delegate` in the other initializer.

md0u80c9:
Agreed.

YuanZhu-apple:
@md0u80c9 Thanks! Merging it.

Issue #502, opened 2015-10-05
RestorationData prevents going backwards in Survey


md0u80c9:
- I have an ORKNavigableOrderedTask with multiple FormSteps in.
- Start the task.
- Complete to second or third page of the form.
- Hit cancel. taskViewControllerSupportsSaveAndRestore is supported, so we 'Save for later'.
- Task is reopened using  taskViewController = ORKTaskViewController(task: StrokeNurseAdmission, restorationData: draft, delegate: self)

Result: The task opens on the page we were last working with. However, back navigation has been removed, so the last completed page is now the first available step in the navigation hierarchy.

NB I haven't managed to test whether this just applies to ORKNavigableOrderedTask or to ORKOrderedTask.

rsanchezsaez:
This is expected behavior, as it was agreed that the ivar that the accessory that `ORKNavigableOrderedTask` uses for maintaining backwards navigation state (`_stepIdentifierStack`) was not a core part of a navigable ordered task, and as such it was nor persisted when using *NSCoding* (which is used when creating or reading restoration data).

We should improve the documentation to make this clear.

---

Or we could change course and save/restore the step identifier stack after all.

md0u80c9:
Hi,

I guess this depends on the use case. I think the current situation may be inconsistent for users: you can go forwards and backwards through a survey, but if you 'save draft' when you return you can't go backwards.

For complex forms being able to traverse the form is very useful.

You may gather my application is looking at stroke care management: the stroke nurse form is a complex one looking at the immediate post-stroke care (to then produce a printable record and then submission to the UK Sentinel Stroke Audit) - and for them it is invaluable to be able to go back and modify their answers if need be. I can see for simpler cases this wouldn't be needed (but then again of course if it's simpler you also may not need to save work in progress at all!)

YuanZhu-apple:
@rsanchezsaez Thanks for joining this discussion. 

`_stepIdentifierStack` is a *state* which is not normally be a part of `Model Object` (e.g. task or step). And all the states should be living on `ORKTaskViewController` and be persisted during restoration.

I talked with @jwe-apple, we would like let `ORKTaskViewController` maintain this identifier stack, and use it to provide a hint in the protocol call. It requires to change API `stepBeforeStep:` to something like `stepBeforeStep:hintStep:`. Then the `ORKTask instance` can decide how to use it.

What do you think?

rsanchezsaez:
[...]

PR #513, opened 2015-10-07
Use results for backwards navigation in 'ORKNavigableOrderedTask'

rsanchezsaez:
This PR improves the situation described on [Issue #502](https://github.com/ResearchKit/ResearchKit/issues/502).

It removes the `ORKNavigableOrderedTask`'s `_stepIdentifierStack` ivar (used for backwards navigation) and uses the *managed task result* instead. The *managed task result* will only show results for the steps that have been previously visited, so it can be used to simulate the backwards navigation stack.

The `ORKTaskViewController`'s `_managedStepIdentifiers` ivar has been converted into an `NSMutableOrderedSet`. When an already visited step is shown (either by going back, or by following a loop in a navigable ordered task), all the managed step identifiers which come later than the step being shown are removed.

This commit also updates `ORKTest` by removing the *Save For Later* navigable ordered task archiving (which is no longer needed) and adds an `Ordered Loop Task` example which can be used to test backwards navigation when looping.

Some minor style refactor was also done on `ORKTest`, and `ORKResult` `description` methods were extended and homogenized for easier debugging.

### DIFF
--- a/ResearchKit/Common/ORKHelpers.m
+++ b/ResearchKit/Common/ORKHelpers.m
@@ -547,10 +547,5 @@ void ORKAdjustPageViewControllerNavigationDirectionForRTL(UIPageViewControllerNa
 }
 
 NSString *ORKPaddingWithNumberOfSpaces(NSUInteger numberOfPaddingSpaces) {
-    NSMutableString *padding = [NSMutableString new];
-    while (numberOfPaddingSpaces) {
-        [padding appendString:@" "];
-        numberOfPaddingSpaces--;
-    }
-    return padding;
+    return [@"" stringByPaddingToLength:numberOfPaddingSpaces withString:@" " startingAtIndex:0];
 }
--- a/ResearchKit/Common/ORKHelpers.m
+++ b/ResearchKit/Common/ORKHelpers.m
@@ -545,3 +545,12 @@ void ORKAdjustPageViewControllerNavigationDirectionForRTL(UIPageViewControllerNa
         *direction = (*direction == UIPageViewControllerNavigationDirectionForward) ? UIPageViewControllerNavigationDirectionReverse : UIPageViewControllerNavigationDirectionForward;
     }
 }
+
+NSString *ORKPaddingWithNumberOfSpaces(NSUInteger numberOfPaddingSpaces) {
+    NSMutableString *padding = [NSMutableString new];
+    while (numberOfPaddingSpaces) {
+        [padding appendString:@" "];
+        numberOfPaddingSpaces--;
+    }
+    return padding;
+}
--- a/ResearchKit/Common/ORKNavigableOrderedTask.m
+++ b/ResearchKit/Common/ORKNavigableOrderedTask.m
@@ -81,11 +81,9 @@
         if (nextStepIdentifier) {
             nextStep = [self stepWithIdentifier:nextStepIdentifier];
             
-            #if defined(DEBUG) && DEBUG
             if (step && nextStep && [self indexOfStep:nextStep] <= [self indexOfStep:step]) {
-                ORK_Log_Warning(@"Index of next step (\"%@\") is equal or lower than index of current step (\"%@\") in ordered task. Make sure this is intentional as you could loop idefinitely without appropriate navigation rules.", nextStep.identifier, step.identifier);
+                ORK_Log_Warning(@"Index of next step (\"%@\") is equal or lower than index of current step (\"%@\") in ordered task. Make sure this is intentional as you could loop idefinitely without appropriate navigation rules. Also please note that you'll get duplicate result entries each time you loop over the same step.", nextStep.identifier, step.identifier);
             }
-            #endif
         } else {
             nextStep = [super stepAfterStep:step withResult:result];
         }
@@ -96,9 +94,10 @@
 - (ORKStep *)stepBeforeStep:(ORKStep *)step withResult:(ORKTaskResult *)result {
     ORKStep *previousStep = nil;
     __block NSInteger indexOfCurrentStepResult = -1;
-    [result.results enumerateObjectsUsingBlock:^(ORKResult *result, NSUInteger idx, BOOL * _Nonnull stop) {
+    [result.results enumerateObjectsWithOptions:NSEnumerationReverse usingBlock:^(ORKResult *result, NSUInteger idx, BOOL *stop) {
         if ([result.identifier isEqualToString:step.identifier]) {
             indexOfCurrentStepResult = idx;
+            *stop = YES;
         }
     }];
     if (indexOfCurrentStepResult != -1 && indexOfCurrentStepResult != 0) {
--- a/ResearchKit/Common/ORKNavigableOrderedTask.m
+++ b/ResearchKit/Common/ORKNavigableOrderedTask.m
@@ -41,14 +41,12 @@
 
 @implementation ORKNavigableOrderedTask {
     NSMutableDictionary *_stepNavigationRules;
-    NSMutableOrderedSet *_stepIdentifierStack;
 }
 
 - (instancetype)initWithIdentifier:(NSString *)identifier steps:(NSArray<ORKStep *> *)steps {
     self = [super initWithIdentifier:identifier steps:steps];
     if (self) {
         _stepNavigationRules = nil;
-        _stepIdentifierStack = nil;
     }
     return self;
 }
@@ -91,52 +89,20 @@
         } else {
             nextStep = [super stepAfterStep:step withResult:result];
         }
-        if (nextStep) {
-            [self updateStepIdentifierStackWithSourceIdentifier:step.identifier destinationIdentifier:nextStep.identifier];
-        }
     }
     return nextStep;
-}
-    
-- (void)updateStepIdentifierStackWithSourceIdentifier:(NSString *)sourceIdentifier destinationIdentifier:(NSString *)destinationIdentifier {
-    NSParameterAssert(destinationIdentifier);
-    
-    if (!_stepIdentifierStack) {
-        _stepIdentifierStack = [NSMutableOrderedSet new];
-        // sourceIdentifier is nil if the task starts fresh,
-        // but can have a value if the task is being restored to a specific step
-        if (sourceIdentifier) {
-            [_stepIdentifierStack addObject:sourceIdentifier];
-        }
-        [_stepIdentifierStack addObject:destinationIdentifier];
-        return;
-    }
-    
-    NSUInteger indexOfSource = [_stepIdentifierStack indexOfObject:sourceIdentifier];
-    if (indexOfSource == NSNotFound) {
-        ORK_Log_Warning(@"You are calling an out of on order step in an ongoing task (\"%@\" -> \"%@\"). Clearing navigation stack.", sourceIdentifier, destinationIdentifier);
-        [_stepIdentifierStack removeAllObjects];
-        if (sourceIdentifier) {
-            [_stepIdentifierStack addObject:sourceIdentifier];
-        }
-        [_stepIdentifierStack addObject:destinationIdentifier];
-        return;
-    }
-    
-    NSUInteger stackCount = _stepIdentifierStack.count;
-    if (indexOfSource != stackCount - 1) {
-        [_stepIdentifierStack removeObjectsInRange:NSMakeRange(indexOfSource + 1, stackCount - (indexOfSource + 1))];
-    }
-    [_stepIdentifierStack addObject:destinationIdentifier];
 }
 
 - (ORKStep *)stepBeforeStep:(ORKStep *)step withResult:(ORKTaskResult *)result {
     ORKStep *previousStep = nil;
-    if (_stepIdentifierStack) {
-        NSUInteger indexOfSource = [_stepIdentifierStack indexOfObject:step.identifier];
-        if (indexOfSource != NSNotFound && indexOfSource >= 1) {
-            previousStep = [self stepWithIdentifier:_stepIdentifierStack[indexOfSource - 1]];
+    __block NSInteger indexOfCurrentStepResult = -1;
+    [result.results enumerateObjectsUsingBlock:^(ORKResult *result, NSUInteger idx, BOOL * _Nonnull stop) {
+        if ([result.identifier isEqualToString:step.identifier]) {
+            indexOfCurrentStepResult = idx;
         }
+    }];
+    if (indexOfCurrentStepResult != -1 && indexOfCurrentStepResult != 0) {
+        previousStep = [self stepWithIdentifier:result.results[indexOfCurrentStepResult - 1].identifier];
     }
     return previousStep;
 }
@@ -161,7 +127,6 @@
     self = [super initWithCoder:aDecoder];
     if (self) {
         ORK_DECODE_OBJ_MUTABLE_DICTIONARY(aDecoder, stepNavigationRules, NSString, ORKStepNavigationRule);
-        ORK_DECODE_OBJ_MUTABLE_ORDERED_SET(aDecoder, stepIdentifierStack, NSString);
     }
     return self;
 }
@@ -170,7 +135,6 @@
     [super encodeWithCoder:aCoder];
     
     ORK_ENCODE_OBJ(aCoder, stepNavigationRules);
-    ORK_ENCODE_OBJ(aCoder, stepIdentifierStack);
 }
 
 #pragma mark NSCopying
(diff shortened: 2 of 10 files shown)


ResearchKit/Common/ORKNavigableOrderedTask.m
```diff
@@ -91,52 +89,20 @@ - (ORKStep *)stepAfterStep:(ORKStep *)step withResult:(ORKTaskResult *)result {
...
 - (ORKStep *)stepBeforeStep:(ORKStep *)step withResult:(ORKTaskResult *)result {
     ORKStep *previousStep = nil;
-    if (_stepIdentifierStack) {
-        NSUInteger indexOfSource = [_stepIdentifierStack indexOfObject:step.identifier];
-        if (indexOfSource != NSNotFound && indexOfSource >= 1) {
-            previousStep = [self stepWithIdentifier:_stepIdentifierStack[indexOfSource - 1]];
+    __block NSInteger indexOfCurrentStepResult = -1;
+    [result.results enumerateObjectsUsingBlock:^(ORKResult *result, NSUInteger idx, BOOL * _Nonnull stop) {
+        if ([result.identifier isEqualToString:step.identifier]) {
+            indexOfCurrentStepResult = idx;
```
YuanZhu-apple: Call `stop` after finding it?

ResearchKit/Common/ORKTaskViewController.m
```diff
@@ -1162,19 +1171,10 @@ - (IBAction)flipToPreviousPageFrom:(ORKStepViewController *)fromController {
...
     if ([self shouldPresentStep:step]) {
-        ORKStep *currentStep = _currentStepViewController.step;
-        NSString *itemId = currentStep.identifier;
-        
         stepViewController = [self viewControllerForStep:step];
-        
         if (stepViewController) {
-            // Remove the identifier from the list
-            assert([itemId isEqualToString:_managedStepIdentifiers.lastObject]);
-            [_managedStepIdentifiers removeLastObject];
```
YuanZhu-apple: The intend of this action is exclude the `currentStep` from taskVC's result reporting.
Any reason to not remove current identifier from `_managedStepIdentifiers` when a user hit `back` button?

ResearchKit/Common/ORKTaskViewController.m
```diff
@@ -893,9 +893,18 @@ - (void)showViewController:(ORKStepViewController *)viewController goForward:(BO
...
-        [_managedStepIdentifiers addObject:step.identifier];
+    NSString *stepIdentifier = step.identifier;
+    if (stepIdentifier) {
+        NSUInteger indexOfStep = [_managedStepIdentifiers indexOfObject:stepIdentifier];
+        if (indexOfStep == NSNotFound) {
+            [_managedStepIdentifiers addObject:stepIdentifier];
+        } else {
+            // If the step being shown has been visited before, remove all the managed step identifiers visited after it
+            NSUInteger numberManagedOfStepIdentifiers = _managedStepIdentifiers.count;
+            [_managedStepIdentifiers removeObjectsInRange:NSMakeRange(indexOfStep + 1, numberManagedOfStepIdentifiers - (indexOfStep + 1))];
```
YuanZhu-apple: Can you explain a little bit more about removing the steps in the loop from `_managedStepIdentifiers`?

YuanZhu-apple:
Just a reminder, not forget to update `ORKESerialization`.

rsanchezsaez:
I will make changes and update the unit tests soon.

ResearchKit/Common/ORKTaskViewController.m
rsanchezsaez: This case is taken care by the new `showViewController:`'s `_managedStepIdentifiers` handling code.

ResearchKit/Common/ORKTaskViewController.m
rsanchezsaez: The idea here is that if you loop in a *navigable ordered task*, you don't want to keep duplicate managed step identifiers.

Before this change, managed steps identifiers were being duplicated if you followed a loop, so managed results were showing duplicate entries.

The result of this is that once you take a forward navigation step jump that goes to a previously visited step, all the history after that step is removed (and backwards navigation only takes into account the steps that made you arrive to that step in the first place).

That way, we avoid potentially redundant and very large managed results (and backwards navigation steps).

Navigable ordered tasks with no loops and regular ordered tasks should be unaffected by this change.

ResearchKit/Common/ORKNavigableOrderedTask.m
rsanchezsaez: Will do.

ResearchKit/Common/ORKTaskViewController.m
YuanZhu-apple: in the first place, I don't think a loop should be allowed or in any type of tasks, because:

- If a step was showed to user more than once (not because use click `back` button), it is equivalent to two steps sharing a **same** identifier. Which is strictly forbidden in RK framework.

Current `NSArray` based `_managedStepIdentifiers` does not prevent the loop happening, but it keeps the actual history of the steps has been visited. It is more like a web browser.  

With the implementation of removing current step from `_managedStepIdentifiers` stack when user hit `back button`,  the stack won't keep very large managed identifiers.

ResearchKit/Common/ORKTaskViewController.m
YuanZhu-apple: This is managed as the browser's back stack, remove current item from stack when user hit `back button`.

ResearchKit/Common/ORKTaskViewController.m
YuanZhu-apple: Maybe the code itself is not that obvious.
Currently we use `push`(when user go forward) and `pop`(when user go backward) to maintain the history stack(`_managedStepIdentifiers`). That's why the code is reside in two places.

And I finally figure it out, your solution is trying to consolidate the code into one place instead of two.

It would works identical with the old code when there is no `designed loop` in a task.

But if there is one a designed loop, for example: 

1.  First visit `step1` 
2. Go forward to `step2`  
3. Go forward to `step3` 
4. Go forward to `step2`.

With your code the final results will only include `step1` and `step2`, and `step3` is omitted, which is incorrect.

Let me know what do you think.

rsanchezsaez: I agree that loops are a bit controversial.

Originally, @jwe-apple and I agreed to allow loops and leave its usage as the developer responsibility. We log a warning to the console when a loop is detected: `Index of next step (\"%@\") is equal or lower than index of current step (\"%@\") in ordered task. Make sure this is intentional as you could loop idefinitely without appropriate navigation rules`.

So, do we still want to allow loops or not? I think it's better to have them for maximum flexibility. There might be some edge use cases for which they could be useful. Imagine a question like this: *"In light of this new piece of information, do you wish to change your answers?"*. If the user responds yes, the loop takes it back to a previous point in the survey.

If we do keep them, is it your contention that the `_managedStepIdentifiers` should hold all the history including loops? Aren't you worried about `managedResults` returning duplicate results for all the times the user looped over (even though all results point to the same result item)? Do we want to keep the whole backwards navigation stack including loops?

YuanZhu-apple: A dictionary type `managedResults` stores the actual results, which is used to fill in user's previous answer on previously visited steps. A previous answer cannot be determined if we store multiple copies of answer for same step. Right now, I cannot think of a better way to replace the `managedResults`. Then returning duplicate results for all the times the user looped over is inevitable.

Being said, currently solution is not perfect. If there is a loop,  `managedResults` preserves user's path, but returning same results are for the same steps.

One solution could be to detect the loop within ORKTaskViewController and log warning for the developer and explain the consequences(returning duplicate results).

rsanchezsaez: Ok, I have reverted the behaviour to the previous push/pop stack model, so all the navigation stack (including loops) is preserved. As discussed, the results for revisited steps appear duplicated on the final task result, but all of them point to the same result instance (which correspond to the user's latest answer).

I have also extended the developer warning mentioning this (produced from `ORKNavigableOrderedTask`), and added some *headerdoc* paragraphs explaining this.

rsanchezsaez:
The feedback has been taken care of and the *unit tests* have been updated.

YuanZhu-apple:
I see this PR is trying to extend `ORKResult description`.

But now it is not easy to read in the console. Can you format them like in `NSArray` or `NSDictionary` to include indent and newline breaks?

screen shot 2015-10-14 at 6 01 49 pm

rsanchezsaez:
Sure, I'll do this in the following days.

rsanchezsaez:
Done.

*Before*

    2015-10-19 00:50:11.378 ORKTest[13883:2136732] [ORKTest] task results: <ORKTaskResult: 0x7f8b08f28760; identifier: "NavigableLoopTaskIdentifier"; results: {<ORKStepResult: 0x7f8b08f37870; identifier: "introStep"; results: {}>, <ORKStepResult: 0x7f8b08f62040; identifier: "loopAStep"; results: {}>, <ORKStepResult: 0x7f8b08e91b50; identifier: "branchingStep"; results: {<ORKChoiceQuestionResult: 0x7f8b08ea6da0; identifier: "branchingStep"; answer: (
    scale
    )>}>, <ORKStepResult: 0x7f8b08e8d2c0; identifier: "scaleStep"; results: {<ORKScaleQuestionResult: 0x7f8b08e706f0; identifier: "scaleStep"; answer: 8.725>}>, <ORKStepResult: 0x7f8b08fbb060; identifier: "loopBStep"; results: {<ORKBooleanQuestionResult: 0x7f8b08f763c0; identifier: "loopBStep"; answer: 0>}>, <ORKStepResult: 0x7f8b08f69d20; identifier: "endStep"; results: {}>}>


*After*

    2015-10-19 00:46:32.931 ORKTest[13039:2129957] [ORKTest] task results: <ORKTaskResult: 0x7fe708691740; identifier: "NavigableLoopTaskIdentifier"; results: (
        <ORKStepResult: 0x7fe708689f40; identifier: "introStep"; results: ()>,
        <ORKStepResult: 0x7fe70866ce70; identifier: "loopAStep"; results: ()>,
        <ORKStepResult: 0x7fe70873d7a0; identifier: "branchingStep"; results: (
            <ORKChoiceQuestionResult: 0x7fe70873d7d0; identifier: "branchingStep"; answer:
    (
        scale
    )
            >
        )>,
        <ORKStepResult: 0x7fe70876f910; identifier: "scaleStep"; results: (
            <ORKScaleQuestionResult: 0x7fe70876ff70; identifier: "scaleStep"; answer: 8.725>
        )>,
        <ORKStepResult: 0x7fe70868f9f0; identifier: "loopBStep"; results: (
            <ORKBooleanQuestionResult: 0x7fe708695f60; identifier: "loopBStep"; answer: 0>
        )>,
        <ORKStepResult: 0x7fe7086f5f70; identifier: "endStep"; results: ()>
    )>

ResearchKit/Common/ORKHelpers.m
```diff
@@ -545,3 +545,12 @@ void ORKAdjustPageViewControllerNavigationDirectionForRTL(UIPageViewControllerNa
         *direction = (*direction == UIPageViewControllerNavigationDirectionForward) ? UIPageViewControllerNavigationDirectionReverse : UIPageViewControllerNavigationDirectionForward;
     }
 }
+
+NSString *ORKPaddingWithNumberOfSpaces(NSUInteger numberOfPaddingSpaces) {
+    NSMutableString *padding = [NSMutableString new];
+    while (numberOfPaddingSpaces) {
```
YuanZhu-apple: Can you try this API: 
https://developer.apple.com/library/mac/documentation/Cocoa/Reference/Foundation/Classes/NSString_Class/#//apple_ref/occ/instm/NSString/stringByPaddingToLength:withString:startingAtIndex:

ResearchKit/Common/ORKResult.m
```diff
@@ -1226,6 +1260,21 @@ - (id)answer {
...
 }
 
+- (NSString *)descriptionWithNumberOfPaddingSpaces:(NSUInteger)numberOfPaddingSpaces {
+    NSMutableString *description = [NSMutableString stringWithFormat:@"%@; answer:", [self descriptionPrefixWithNumberOfPaddingSpaces:numberOfPaddingSpaces]];
+    id answer = self.answer;
+    if ([answer isKindOfClass:[NSArray class]]
+        || [answer isKindOfClass:[NSDictionary class]]
+        || [answer isKindOfClass:[NSSet class]]
+        || [answer isKindOfClass:[NSOrderedSet class]]) {
+        [description appendFormat:@"\n%@\n%@>", answer, ORKPaddingWithNumberOfSpaces(numberOfPaddingSpaces)];
```
YuanZhu-apple: It could look nicer if we can apply indentation to native type printings:)

screen shot 2015-10-19 at 2 33 36 pm

ResearchKit/Common/ORKResult.m
```diff
@@ -1773,6 +1822,26 @@ - (ORKResult *)firstResult {
...
 
+- (NSString *)descriptionWithNumberOfPaddingSpaces:(NSUInteger)numberOfPaddingSpaces {
+    NSMutableString *description = [NSMutableString stringWithFormat:@"%@; results: (", [self descriptionPrefixWithNumberOfPaddingSpaces:numberOfPaddingSpaces]];
+    
+    NSUInteger numberOfResults = self.results.count;
+    [self.results enumerateObjectsUsingBlock:^(ORKResult *result, NSUInteger idx, BOOL * _Nonnull stop) {
+        if (idx == 0) {
+            [description appendString:@"\n"];
+        }
+        [description appendFormat:@"%@", [result descriptionWithNumberOfPaddingSpaces:numberOfPaddingSpaces + 4]];
```
YuanZhu-apple: Declare `4` as a constant?

YuanZhu-apple:
Thanks @rsanchezsaez, now the console log is much nicer.

ResearchKit/Common/ORKResult.m
rsanchezsaez: Agreed, but I'm not sure how to do that effectively.

From outside, you can only control indentation of the first line when calling `description` on collection instances (i.e., `NSArray`, `NSDictionary`, `NSSet`, and `NSOrderedSet`).

A way to do this that I can think of is to implement `ork_descriptionWithNumberOfPaddingSpaces:` in extensions for the collection types, which would manually iterate the contents and avoid using `description`.

What do you think? Should I try this approach or do you have any more elegant solution in mind?

YuanZhu-apple: I would do:

```
NSString *description = [answer description];
NSArray *lines= [description componentsSeparatedByString:@"\n"];

// Pad each line
// Join them back to one string

```

rsanchezsaez: Good idea! I'll do this soon.

rsanchezsaez:
Feedback taken care of.

Final result:

    2015-10-20 02:39:29.378 ORKTest[41408:3424955] [ORKTest] task results: <ORKTaskResult: 0x7fef83001000; identifier: "NavigableLoopTaskIdentifier"; results: (
        <ORKStepResult: 0x7fef81cef160; identifier: "introStep"; results: ()>,
        <ORKStepResult: 0x7fef83009820; identifier: "loopAStep"; results: ()>,
        <ORKStepResult: 0x7fef81eaae10; identifier: "branchingStep"; results: (
            <ORKChoiceQuestionResult: 0x7fef81e9a710; identifier: "branchingStep"; answer:
                (
                    scale
                )>
        )>,
        <ORKStepResult: 0x7fef81f58220; identifier: "scaleStep"; results: (
            <ORKScaleQuestionResult: 0x7fef81f5a060; identifier: "scaleStep"; answer: 8.725>
        )>,
        <ORKStepResult: 0x7fef81ce87f0; identifier: "loopBStep"; results: (
            <ORKBooleanQuestionResult: 0x7fef81cad750; identifier: "loopBStep"; answer: 0>
        )>,
        <ORKStepResult: 0x7fef81ecf960; identifier: "endStep"; results: ()>
    )>

YuanZhu-apple:
Now the formatting looks great. 
This PR is good to go for me!
@umerkhan-apple Any feedback you would like to add?

md0u80c9:
rsanchezsaez: Just a quick thought based on some work this evening. Would the work you have done on this permit an enhancement to be able to reopen a completed survey at a specified page or identifier?

My app is now working using a series of form-based tasks for a patient. It is conceivable that they may need to correct information on one of the values on the form. The restoration data would give us a reconstruction of the task's states. But clearly if you revised a form you would be starting at the last step: more commonly you would want to rewind to the first step.

rsanchezsaez:
Have you seen the [Review Step PR](https://github.com/ResearchKit/ResearchKit/pull/463)? It looks like what you need.

Otherwise, you could use some `ORKTaskViewControllerDelegate` method to save the task view controller's `restorationData` at different times, and use those to restore the forms later on.

PR #515, opened 2015-10-08
Clear unexpected exceptions from unit tests.

YuanZhu-apple:
`HTMLForSignature:` would throw exception on a nil `signature.title`


umerkhan-apple:
r=m3!

Issue #516, opened 2015-10-08
Consent required name logic is invalid


umerkhan-apple:
## Consent Bug
The consent review step requires First Name and Last Name, but only entering the First Name allows the user to proceed.

#### Steps to reproduce
1. Open `ORKTest`
2. Tap `Consent Review`
3. Continue until you're asked to enter your name.
4. Type in `First Name`
5. `Next` button becomes enabled.

_* This can also be reproduced in `ORKCatalog`._

![simulator screen shot oct 8 2015 2 40 12 pm](https://cloud.githubusercontent.com/assets/13040996/10380819/17916ee6-6dcb-11e5-8b88-f574a1091273.png)

YuanZhu-apple:
Taking on it.

PR #518, opened 2015-10-08
Fix Issue #516: Consent required name logic is invalid

YuanZhu-apple:
Fix for #516

Regression caused by #443 (ORKFormItem: implement 'optional' property).


umerkhan-apple:
r=m3! :+1:

PR #527, opened 2015-10-13
ORKGraphChartView - Easier requery of dataSource

huguesbr:
Re-query datasource on dataSource update (legacy) but also on animateWithDuration and on setNeedsLayout (new method to requery dataSource).

It's not obvious that the only way to requery dataSource is to update the dataSource.
I propose to do it on dataSource update but also on animateWithDuration and setNeedsLayout. 
We could also create a new methods (update) but didn't want to pollute the API.

### DIFF
--- a/ResearchKit/Common/ORKTaskViewController.h
+++ b/ResearchKit/Common/ORKTaskViewController.h
@@ -307,7 +307,7 @@ ORK_CLASS_AVAILABLE
  
  @return A new task view controller.
  */
-- (instancetype)initWithTask:(nullable id<ORKTask>)task restorationData:(NSData *)data delegate:(nullable id<ORKTaskViewControllerDelegate>)delegate;
+- (instancetype)initWithTask:(nullable id<ORKTask>)task restorationData:(nullable NSData *)data delegate:(nullable id<ORKTaskViewControllerDelegate>)delegate;
 
 /**
  The delegate for the task view controller.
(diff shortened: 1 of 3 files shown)


YuanZhu-apple:
@rsanchezsaez Can you help review this one?

rsanchezsaez:
I'd advise against doing this change. The current implementation separates the *drawn layers set up* (which as you found out is only done when you set the `dataSource`), from the *layers layout* (which is only done on the view layout). This is done to try to improve the performance when rotating the device (the *`dataSource` data* is not expected to change on a device rotation, but the view layout is).

I have submitted a new [PR here](https://github.com/ResearchKit/ResearchKit/pull/528) which adds a new `Charts Performance Test` to `ORKTest` in order to assess this, and I think this change negatively affects performance.

I suggest that a better solution is to more explicitly document that if the developer expects any changes in the data provided by the `dataSource` midway through, they need to set the `dataSource` manually again to produce an update (or add a new `reloadData` method, mimicking `UITableView`'s).

huguesbr:
@rsanchezsaez, I totally agree. Can then add a reloadData method?

rsanchezsaez:
Sure, if you want to add a `reloadData` method that'd be fine. Don't forget to add it for the *pie chart* as well, and add some headerdoc documentation comments. :-)

rsanchezsaez:
In case this still of interest, I implemented the `reloadData` methods on the just merged [Pull Request #602](https://github.com/ResearchKit/ResearchKit/pull/602).

huguesbr:
@rsanchezsaez thx you for this!

PR #538, opened 2015-10-19
Onboarding - Account Module

umerkhan-apple:
## Introducing Account Module.

This PR includes a registration step, a verification step, a login step, and two view controllers.
- The registration step is used to create a new account. This is to be used along with the eligibility screening and informed consent process.
 - The form items can be customized using the `ORKRegistrationOptions` parameter.
- The verification step is used to inform the user that email verification is pending.
  - The verification view controller allows for custom navigation logic for button actions including `Wrong email address?`, `Resend Verification Email`, and the `Continue` button.
  - A recommended use example using a registration step, wait step, and verification step together is available in `ORKCatalog`.
- The login step is used to provide an interface for existing users. 
  - The login view controller allows for custom navigation logic for the `Forgot password?` button.

#### Screenshots.
simulator screen shot oct 19 2015 11 07 45 am simulator screen shot oct 19 2015 11 08 30 am simulator screen shot oct 19 2015 11 08 36 am


md0u80c9:
This looks awesome. Really useful. I'll drop it into my app now and test it out :-).

ResearchKit/Common/ORKAnswerFormat.m
```diff
@@ -1920,6 +1921,14 @@ - (ORKQuestionType)questionType {
     return ORKQuestionTypeText;
 }
 
+- (void)validateParameters {
+    [super validateParameters];
+    
+    NSAssert(((!self.regex && !self.invalidMessage) || (self.regex && self.invalidMessage)),
```
YuanZhu-apple: To be consistent, use `@throw exception`:)

ResearchKit/Onboarding/ORKRegistrationStep.m
```diff
@@ -0,0 +1,258 @@
...
+                                                               text:ORKLocalizedString(@"CONSENT_NAME_FAMILY", nil)
+                                                       answerFormat:answerFormat];
+        item.placeholder = ORKLocalizedString(@"FAMILY_NAME_ITEM_PLACEHOLDER", nil);
+        item.optional = NO;
+        
+        [formItems addObject:item];
+    }
+    
+    if (_options & ORKRegistrationStepIncludeGender) {
+        NSArray *textChoices = @[[ORKTextChoice choiceWithText:ORKLocalizedString(@"GENDER_MALE", nil) value:@0],
```
YuanZhu-apple: Without looking at the code, it is a little hard for developer to figure out the meaning of `0/1/2` ...

ResearchKit/Onboarding/ORKRegistrationStep.m
```diff
@@ -0,0 +1,258 @@
...
+        item.optional = NO;
+       
+        [formItems addObject:item];
+    }
+    
+    if (_options & ORKRegistrationStepIncludeDOB) {
+        // Calculate default date (20 years from now).
+        NSDateComponents *minusTwentyYears = [NSDateComponents new];
+        minusTwentyYears.year = -20;
+        NSDate *defaultDate = [[NSCalendar currentCalendar] dateByAddingComponents:minusTwentyYears
```
YuanZhu-apple:
```
calendar  dateByAddingUnit:(NSCalendarUnit)unit
                       value:(NSInteger)value
                      toDate:(NSDate *)date
                     options:(NSCalendarOptions)options
```

ResearchKit/Onboarding/ORKRegistrationStep.m
```diff
@@ -0,0 +1,258 @@
...
+                                                                                toDate:[NSDate date]
+                                                                               options:0];
+        
+        ORKDateAnswerFormat *answerFormat = [ORKAnswerFormat dateAnswerFormatWithDefaultDate:defaultDate
```
YuanZhu-apple: Use `ORKHealthAnswerFormat` to auto load DOB from HK?

ResearchKit/Onboarding/ORKRegistrationStep.m
```diff
@@ -0,0 +1,258 @@
...
+}
+
+- (ORKTextAnswerFormat *)passwordAnswerFormat {
+    ORKFormItem *passwordFormItem = ORKFindInArrayByFormItemId(self.formItems, ORKRegistrationFormItemPassword);
+    ORKTextAnswerFormat *passwordAnswerFormat = (ORKTextAnswerFormat *)passwordFormItem.answerFormat;
+    return passwordAnswerFormat;
+}
+
+- (NSString *)passcodeValidationRegex {
+    return [self passwordAnswerFormat].regex;
```
YuanZhu-apple: Not from your code, rename `regex` to `validationRegex` on `ORKTextAnswerFormat` too? Just feel `validationRegex` is more descriptive.

ResearchKit/Common/ORKAnswerFormat.m
```diff
@@ -2062,6 +2071,66 @@ - (ORKAnswerFormat *)impliedAnswerFormat {
...
+- (instancetype)initWithOriginalItemIdentifier:(NSString *)originalItemIdentifier {
+    self = [super init];
+    if (self) {
+        _originalItemIdentifier = [originalItemIdentifier copy];
+    }
+    return self;
+}
+
+- (BOOL)isAnswerValid:(id)answer {
+    return [(NSNumber *)answer isEqual:@YES] ? YES : NO;
```
YuanZhu-apple: I would check `isKindOfClass` before comparing. And do we need to check  `ORKIsAnswerEmpty()`? I am not sure.

ResearchKit/Common/ORKAnswerFormat.m
```diff
@@ -2062,6 +2071,66 @@ - (ORKAnswerFormat *)impliedAnswerFormat {
...
+@implementation ORKConfirmTextAnswerFormat
+
+- (Class)questionResultClass {
+    return [ORKBooleanQuestionResult class];
+}
+
+- (instancetype)initWithOriginalItemIdentifier:(NSString *)originalItemIdentifier {
+    self = [super init];
+    if (self) {
+        _originalItemIdentifier = [originalItemIdentifier copy];
```
YuanZhu-apple: If `_originalItemIdentifier` is only set through init and `readonly`, then we need to make sure it is not nil at here, instand of in `validateParameters`.

ResearchKit/Common/ORKAnswerFormat.m
```diff
@@ -2062,6 +2071,66 @@ - (ORKAnswerFormat *)impliedAnswerFormat {
...
+}
+
+- (BOOL)isAnswerValid:(id)answer {
+    return [(NSNumber *)answer isEqual:@YES] ? YES : NO;
+}
+
+- (void)validateParameters {
+    [super validateParameters];
+    
+    NSAssert(_originalItemIdentifier, @"Original item identifier cannot be nil.");
```
YuanZhu-apple: Move this assert to init, then this `validateParameters` can be removed.

ResearchKit/Common/ORKAnswerFormat.m
```diff
@@ -2062,6 +2071,66 @@ - (ORKAnswerFormat *)impliedAnswerFormat {
...
+}
+
+- (instancetype)copyWithZone:(NSZone *)zone {
+    ORKConfirmTextAnswerFormat *fmt = [super copyWithZone:zone];
+    fmt->_originalItemIdentifier = _originalItemIdentifier;
```
YuanZhu-apple: `[_originalItemIdentifier copy]` :)

ResearchKit/Common/ORKFormItemCell.m
```diff
@@ -493,6 +493,54 @@ - (BOOL)isAccessibilityElement {
...
+
+#pragma mark UITextFieldDelegate
+
+- (BOOL)textField:(UITextField *)textField shouldChangeCharactersInRange:(NSRange)range replacementString:(NSString *)string {
+    NSString *text = [textField.text stringByReplacingCharactersInRange:range withString:string];
+    if ([self isAnswerValidWithString:text]) {
+        [self ork_setAnswer:@YES];
+    } else {
+        [self ork_setAnswer:@NO];
+    }
```
YuanZhu-apple: Prefer the one liner:

```
[self ork_setAnswer:[self isAnswerValidWithString:text] ? @YES : @NO];
```

ResearchKit/Common/ORKFormItemCell.h
```diff
@@ -63,6 +63,7 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, copy, nullable) id defaultAnswer;
 @property (nonatomic) CGFloat maxLabelWidth;
 @property (nonatomic) CGFloat expectedLayoutWidth;
+@property (nonatomic) NSMutableDictionary   * __nullable __strong * __nullable savedAnswers;
```
YuanZhu-apple: Is `**` required? 
And if the cell is not going to modified the dictionary, then just declare  `NSDictionary`, which is more safe.

ResearchKit/Onboarding/ORKLoginStep.h
```diff
@@ -0,0 +1,86 @@
...
+#import <ResearchKit/ResearchKit.h>
+
+
+NS_ASSUME_NONNULL_BEGIN
+
+/**
+ Constants for the form items included in the login step.
+ These allow for convenient retrieval of user's inputted data from the result.
+ */
+ORK_EXTERN NSString *const ORKLoginFormItemEmail ORK_AVAILABLE_DECL;
```
YuanZhu-apple: Do you like `ORKLoginFormItemIdentiferEmail`?

ResearchKit/Onboarding/ORKRegistrationStep.m
```diff
@@ -0,0 +1,258 @@
...
+        item.optional = NO;
+        
+        [formItems addObject:item];
+    }
+    
+    if (_options & ORKRegistrationStepIncludeFamilyName) {
+        ORKTextAnswerFormat *answerFormat = [ORKAnswerFormat textAnswerFormat];
+        answerFormat.multipleLines = NO;
+        
+        ORKFormItem *item = [[ORKFormItem alloc] initWithIdentifier:ORKRegistrationFormItemFamilyName
```
YuanZhu-apple: `FamilyName` and `GivenName` is a little bit tricky:) 
Please refer to the consent `FamilyName` and `GivenName` section to make sure you list them in the right order.

ResearchKit/Onboarding/ORKVerificationStep.m
```diff
@@ -0,0 +1,98 @@
...
+    if (self) {
+        self.title = title;
+        self.text = text;
+        _verificationViewControllerString = NSStringFromClass(verificationViewControllerClass);
+    }
+    return self;
+}
+
+- (Class)verificationViewControllerClass {
+    return NSClassFromString(_verificationViewControllerString);
```
YuanZhu-apple: It would be more safe to throw exception when the conversion is failed.

ResearchKit/Common/ORKAnswerFormat.m
umerkhan-apple: Don't necessarily need to check if answer is empty because it is a non-optional form item. But for robustness, I'm going to add the check, just in case if someone wants to use it as an optional item.

ResearchKit/Common/ORKAnswerFormat.m
umerkhan-apple: Good call!

ResearchKit/Onboarding/ORKRegistrationStep.m
umerkhan-apple: Yep, will update the initializer to reflect the same name also.

ResearchKit/Onboarding/ORKRegistrationStep.m
umerkhan-apple: Would you recommend changing it to the same as the text?

ResearchKit/Onboarding/ORKRegistrationStep.m
umerkhan-apple: Not sure if the permissions page has been presented already at this point. I'm not certain if I want to grab anything from HK during the on boarding process. What do you think?

ResearchKit/Common/ORKAnswerFormat.m
umerkhan-apple: Good catch!

ResearchKit/Onboarding/ORKLoginStep.h
umerkhan-apple: YES!! Good call!

ResearchKit/Onboarding/ORKVerificationStep.m
umerkhan-apple: Yep, good call!

umerkhan-apple:
@md0u80c9 Thanks Andrew! Let me know if you find anything I might've overlooked.

Also, do we need to include an example in ORKTest?

ResearchKit/Common/ORKAnswerFormat.m
```diff
@@ -2062,6 +2074,68 @@ - (ORKAnswerFormat *)impliedAnswerFormat {
...
+
+@implementation ORKConfirmTextAnswerFormat
+
+- (Class)questionResultClass {
+    return [ORKTextQuestionResult class];
+}
+
+- (instancetype)initWithOriginalItemIdentifier:(NSString *)originalItemIdentifier {
+    
+    NSAssert(originalItemIdentifier, @"Original item identifier cannot be nil.");
```
YuanZhu-apple: I don't know why, but `NSParameterAssert` is recommended in this case.

ResearchKit/Common/ORKAnswerFormat.m
```diff
@@ -2062,6 +2074,68 @@ - (ORKAnswerFormat *)impliedAnswerFormat {
...
+        _originalItemIdentifier = [originalItemIdentifier copy];
+    }
+    return self;
+}
+
+- (BOOL)isAnswerValid:(id)answer {
+    BOOL isValid = NO;
+    if ([answer isKindOfClass:[NSString class]]) {
+        NSString *stringAnswer = (NSString *)answer;
+        isValid = (stringAnswer.length > 0) ? YES : NO;
```
YuanZhu-apple: How about just leave `isValid = (stringAnswer.length > 0);`

ResearchKit/Common/ORKQuestionStep.m
```diff
@@ -86,6 +87,10 @@ - (instancetype)init {
 
 - (void)validateParameters {
     [super validateParameters];
+    
+    NSAssert(![self.answerFormat isKindOfClass:[ORKConfirmTextAnswerFormat class]],
```
YuanZhu-apple: Throw an exception is prefered.

ResearchKit/Common/ORKFormStepViewController.m
```diff
@@ -825,6 +828,12 @@ - (UITableViewCell *)tableView:(UITableView *)tableView cellForRowAtIndexPath:(N
                         [formCell setExpectedLayoutWidth:self.tableView.bounds.size.width];
                         formCell.selectionStyle = UITableViewCellSelectionStyleNone;
                         formCell.defaultAnswer = _savedDefaults[formItem.identifier];
+                        if ([class isSubclassOfClass:[ORKFormItemConfirmTextCell class]]) {
```
YuanZhu-apple: `setSavedAnswers` is available on `ORKFormItemCell`, do we need this check here?

ResearchKit/Common/ORKAnswerFormat.m
```diff
@@ -2062,6 +2074,68 @@ - (ORKAnswerFormat *)impliedAnswerFormat {
 @end
 
 
+#pragma mark - ORKConfirmTextAnswerFormat
+
+@implementation ORKConfirmTextAnswerFormat
+
+- (Class)questionResultClass {
+    return [ORKTextQuestionResult class];
```
YuanZhu-apple: If `questionResultClass` is unchanged, we can drop this override.

ResearchKit/Common/ORKFormItemCell.m
```diff
@@ -493,6 +493,72 @@ - (BOOL)isAccessibilityElement {
 @end
 
 
+#pragma mark - ORKFormItemConfirmTextCell
+
+@implementation ORKFormItemConfirmTextCell
+
+- (void)setSavedAnswers:(NSDictionary *)savedAnswers {
+    [super setSavedAnswers:savedAnswers];
+
```
YuanZhu-apple: To be cautious, throw exception on  `savedAnswers` is nil, at here or in the super class?

ResearchKit/Common/ORKFormItemCell.m
```diff
@@ -493,6 +493,72 @@ - (BOOL)isAccessibilityElement {
...
+        textField.text = @"";
+        [self ork_setAnswer:textField.text];
+        [self showValidityAlertWithMessage:ORKLocalizedString(@"CONFIRM_PASSWORD_ERROR_MESSAGE", nil)];
+    }
+    return YES;
+}
+
+- (void)textFieldDidEndEditing:(UITextField *)textField {
+    self.editingHighlight = NO;
+    [self.delegate formItemCellDidResignFirstResponder:self];
```
YuanZhu-apple: If you don't call `[self ork_setAnswer:textField.text];` in `textFieldShouldEndEditing:`
Maybe you can just reuse super class's implementation of `textFieldDidEndEditing:`

ResearchKit/Onboarding/ORKLoginStep.m
```diff
@@ -0,0 +1,149 @@
...
+}
+
+- (Class)loginViewControllerClass {
+    return NSClassFromString(_loginViewControllerString);
+}
+
+- (void)validateParameters {
+    [super validateParameters];
+    
+    if (!_loginViewControllerString) {
```
YuanZhu-apple: Need to give a try to convert it to class, right?

ResearchKit/Onboarding/ORKLoginStep.m
```diff
@@ -0,0 +1,149 @@
...
+}
+
+- (void)encodeWithCoder:(NSCoder *)aCoder {
+    [super encodeWithCoder:aCoder];
+    ORK_ENCODE_OBJ(aCoder, loginViewControllerString);
+}
+
+- (instancetype)copyWithZone:(NSZone *)zone {
+    ORKLoginStep *step = [super copyWithZone:zone];
+    step->_loginViewControllerString = self.loginViewControllerString;
```
YuanZhu-apple: `copy`

ResearchKit/Onboarding/ORKLoginStep.m
```diff
@@ -0,0 +1,149 @@
...
+    if (self) {
+        _loginViewControllerString = NSStringFromClass(loginViewControllerClass);
+        self.formItems = [self loginFormItems];
+        
+        [self validateParameters];
+    }
+    return self;
+}
+
+- (NSArray *)loginFormItems {
```
YuanZhu-apple: Use generics in NSArray

ResearchKit/Common/ORKFormItemCell.m
umerkhan-apple: Yep good catch. I had to do this earlier when storing a boolean result, because otherwise it would try to store the textfield's text into a boolean result. But now that I switched back to storing the actual text, I can revert this as well!

ResearchKit/Onboarding/ORKRegistrationStep.m
```diff
@@ -0,0 +1,272 @@
...
+        ORKDateAnswerFormat *answerFormat = [ORKAnswerFormat dateAnswerFormatWithDefaultDate:defaultDate
+                                                                                 minimumDate:nil
+                                                                                 maximumDate:nil
+                                                                                    calendar:[NSCalendar currentCalendar]];
+        
+        ORKFormItem *item = [[ORKFormItem alloc] initWithIdentifier:ORKRegistrationFormItemIdentifierDOB
+                                                               text:ORKLocalizedString(@"DOB_FORM_ITEM_TITLE", nil)
+                                                       answerFormat:answerFormat];
+        item.placeholder = ORKLocalizedString(@"DOB_FORM_ITEM_PLACEHOLDER", nil);
+        item.optional = NO;
```
YuanZhu-apple: Use the new API? (init with optional?)

ResearchKit/Onboarding/ORKVerificationStep.m
```diff
@@ -0,0 +1,110 @@
...
+}
+
+- (void)validateParameters {
+    [super validateParameters];
+    
+    if (!_verificationViewControllerString) {
```
YuanZhu-apple: Same here, try to convert it first.

ResearchKit/Onboarding/ORKVerificationStep.m
```diff
@@ -0,0 +1,110 @@
...
+}
+
+- (void)encodeWithCoder:(NSCoder *)aCoder {
+    [super encodeWithCoder:aCoder];
+    ORK_ENCODE_OBJ(aCoder, verificationViewControllerString);
+}
+
+- (instancetype)copyWithZone:(NSZone *)zone {
+    ORKVerificationStep *step = [super copyWithZone:zone];
+    step->_verificationViewControllerString = self.verificationViewControllerString;
```
YuanZhu-apple: `copy`

ResearchKit/Onboarding/ORKVerificationStepViewController.m
```diff
@@ -0,0 +1,107 @@
...
+
+- (void)changeEmailButtonHandler:(id)sender {
+    [self changeEmailButtonTapped];
+}
+
+- (void)changeEmailButtonTapped {
+}
+
+- (NSString *)emailAddress {
+    return nil;
```
YuanZhu-apple: https://github.com/ResearchKit/ResearchKit/blob/7f106898e23dd00057dc08a77591913f851f6bd7/ResearchKit/Charts/ORKGraphChartView.m#L1012-L1016

Should enforce the developer to subclass the required abstract methods (all of them)?

ResearchKit/Common/ORKFormItemCell.m
umerkhan-apple: Decided to throw the exception in the super class, since savedAnswers is available in ORKFormItemCell.

umerkhan-apple:
@YuanZhu-apple Can you do another round of review when you get a chance?
Also, do we want to provide a passcode validation regex property on the Login Step as well?
I don't see that as a common pattern on login screens.

ResearchKit/Localized/en.lproj/Localizable.strings
```diff
@@ -101,6 +101,32 @@
...
+"GENDER_FORM_ITEM_TITLE" = "Gender";
+"GENDER_FORM_ITEM_PLACEHOLDER" = "Pick a gender";
+"DOB_FORM_ITEM_TITLE" = "Date of Birth";
+"DOB_FORM_ITEM_PLACEHOLDER" = "Pick a date";
+
+/* Verification step */
+"CHANGE_EMAIL_BUTTON_TITLE" = "Wrong email address?";
+"RESEND_EMAIL_LABEL_MESSAGE" = "I didn't recieve an email";
+"RESEND_EMAIL_BUTTON_TITLE" = "Resend Verification Email";
+"VERIFICATION_LABEL_MESSAGE" = "If verified, tap below.";
```
YuanZhu-apple: `tap the button below`? Ignore this, if you don't like it.

ResearchKit/Common/ORKAnswerFormat.h
```diff
@@ -1199,13 +1201,13 @@ ORK_CLASS_AVAILABLE
  
  This method is one of the designated initializers.
  
- @param expression                The regular expression used to validate the text.
+ @param validationRegex                The regular expression used to validate the text.
```
YuanZhu-apple: Alignment.

ResearchKit/Common/ORKAnswerFormat.m
```diff
@@ -1927,10 +1928,21 @@ - (ORKQuestionType)questionType {
...
 
+- (void)validateParameters {
+    [super validateParameters];
+    
+    if ( (!self.validationRegex && self.invalidMessage) || (self.validationRegex && !self.invalidMessage) ){
+        @throw [NSException exceptionWithName:NSInvalidArgumentException
+                                       reason:@"Both regex and invalid message properties must be set."
+                                     userInfo:nil];
+    }
+
```
YuanZhu-apple: Shall we also throw exception on an invalid `validationRegex`?

ResearchKit/Common/ORKAnswerFormat.m
```diff
@@ -1907,10 +1908,10 @@ - (instancetype)initWithMaximumLength:(NSInteger)maximumLength {
     return self;
 }
 
-- (instancetype)initWithValidationExpression:(NSString *)expression invalidMessage:(NSString *)invalidMessage {
+- (instancetype)initWithValidationRegex:(NSString *)validationRegex invalidMessage:(NSString *)invalidMessage {
     self = [super init];
     if (self) {
-        _regex = expression;
+        _validationRegex = validationRegex;
```
YuanZhu-apple: Copy it.

ResearchKit/Common/ORKAnswerFormat.m
```diff
@@ -1907,10 +1908,10 @@ - (instancetype)initWithMaximumLength:(NSInteger)maximumLength {
...
         _invalidMessage = invalidMessage;
```
YuanZhu-apple: Same here

ResearchKit/Common/ORKAnswerFormat.m
```diff
@@ -1927,10 +1928,21 @@ - (ORKQuestionType)questionType {
...
+                                     userInfo:nil];
+    }
+
+}
+
 - (instancetype)copyWithZone:(NSZone *)zone {
     ORKTextAnswerFormat *fmt = [[[self class] allocWithZone:zone] init];
     fmt->_maximumLength = _maximumLength;
-    fmt->_regex = _regex;
+    fmt->_validationRegex = _validationRegex;
```
YuanZhu-apple: copy

ResearchKit/Common/ORKAnswerFormat.m
```diff
@@ -1927,10 +1928,21 @@ - (ORKQuestionType)questionType {
...
     fmt->_invalidMessage = _invalidMessage;
```
YuanZhu-apple: And this one

ResearchKit/Common/ORKHelpers.h
```diff
@@ -162,6 +162,7 @@ UIColor *ORKRGB(uint32_t x);
 UIColor *ORKRGBA(uint32_t x, CGFloat alpha);
 
 id ORKFindInArrayByStudyId(NSArray *array, NSString *studyIdentifier);
+id ORKFindInArrayByFormItemId(NSArray *array, NSString *formItemIdentifier);
```
YuanZhu-apple: If this utility is only used within `ORKRegistrationStep`, move it into ORKRegistrationStep.m?
And `ORKFindInArrayByStudyId` is not being used in anywhere, consider to drop it too?

umerkhan-apple:
@YuanZhu-apple Comments addressed! :)

YuanZhu-apple:
r=me !

md0u80c9:
Hi Umer,

Finally had chance to play with this over the weekend (apologies - haven't contributed much to the ResearchKit codebase over the last week as have been busy getting my app ready - anyone interested I'm happy to send a TestFlight invite to in the next week or two as it is quite extensively testing ResearchKit).

I may be wrong, but is it actually possible to implement the 'verification email' system proposed without resorting to a server or a private API? I didn't think there was a public API on iOS to produce an email to send an email without showing the email view controller?  If you show the view controller then of course you give away the verification code.

umerkhan-apple:
@md0u80c9 It'd be nice to see your work whenever you're done!

Unfortunately, it is not possible to implement the verification email the way you have proposed. There are several reasons blocking us from doing that. In iOS, we cannot send an email without showing the email view controller.

Also an email account is required to be set up before the email view controller can even be presented. So lets say that the user did have an email account set up, then it would be sending an email from their account, which is not something that they would want.

Most vendors also have security checking on servers to ensure that verification emails are being sent by their domain name.

So all in all, the app can trigger sending the email, but the email has to be sent through the server.

Issue #545, opened 2015-10-26
Localization & Pods


huguesbr:
When using ResearchKit as a Pod, all the `Localizable.strings`'s files are included in the finale bundle which then for conflict with the current projet `Localizable.strings`'s file.

I propose to rename all `Localizable.strings`'s file to `ResearchKit.strings` (as most of Libraries do) and tweak the `ORKDefaultLocalizedValue` and `ORKLocalizedString` to use the right table.

See the linked PR (https://github.com/ResearchKit/ResearchKit/issues/546 )for detail.

huguesbr:
@YuanZhu-apple any chance you could review?

PR #546, opened 2015-10-26
Fix localization when using  ResearchKit as Pod (Localizable.strings > ResearchKit.strings)

huguesbr:
Rename Localizable.strings file to ResearchKit.strings and use table name in localization, avoiding conflict when including ResearchKit as a pod (the Localizable.strings files conflict with project one)

Fix: https://github.com/ResearchKit/ResearchKit/issues/545


YuanZhu-apple:
Thanks for bringing this up to us!

Overall this looks good to me.

One more change you want to include in this PR:
https://github.com/ResearchKit/ResearchKit/blame/master/CONTRIBUTING.md#L90

@rsanchezsaez Any feedback for this one?
Do you think https://github.com/ResearchKit/ResearchKit/blob/master/scripts/plist-to-strings.py need a update?

rsanchezsaez:
This PR looks good.

Regarding `plist-to-strings.py`, I think the code is ok but the usage examples in the heading comment need to be updated from `Localizable.strings` to `ResearchKit.strings`.

Additionally, I'd take the chance to rename `ORKDefaultLocaleBundle()` to `ORKBundleForDevelopmentLocale()`.

Also, I'd move all of `ORKDefines_Private.h` contents to `ORKHelpers.h` (since most of them are not defines, and their implementations are in in `ORKHelpers.m`).

And maybe extract the `ORKDefines.h` enums to a new file called `ORKTypes.h`.

YuanZhu-apple:
@rsanchezsaez Thanks for the review!

I agree with you to update the example code in `plist-to-strings.py`.

And maybe we can put `ORKDefines_Private.h` refactoring into a separate PR?

huguesbr:
@YuanZhu-apple @rsanchezsaez I've made the suggested update. let me know.

YuanZhu-apple:
@huguesbr Thanks!
This looks good to me.

rsanchezsaez:
I think that the `ResearchKit.strings` rename removed some blank spaces from the copyright headers. How come this happened?

It's not a big deal, but maybe it's better to leave them for coherence with the rest of the files.

umerkhan-apple:
Agree with @rsanchezsaez. Apart from minor nit, r=m3! :+1:

huguesbr:
@rsanchezsaez @umerkhan-apple youps... fixed...

YuanZhu-apple:
Thanks! Merging it.

huguesbr:
thanks

Issue #517, opened 2015-10-08
Image capture no permissions


umerkhan-apple:
# Image Capture bug
In the `Image Capture` task in `ORKTest`, if the user does not give permission to use the camera, the `Next` button does nothing. It seems as if this is a bug and not intended functionality. A suggested fix would be to show an alert if there is no camera permission when the user hits the `Next` button.

In the `Image Capture` task in `ORKCatalog`, if the user does not give permission to use the camera, the `Next` button just ends the task. It seems like a bug that is causing the task to crash. Same suggested fix as above,

#### Steps to reproduce on ORKTest
1. Open `ORKTest`.
2. Tap `Image Capture Step`.
3. Tap `Continue` until you cannot.

#### Steps to reproduce on ORKCatalog
1. Open `ORKCatalog`.
2. Tap `Image Capture Step`.
3. Tap `Continue` and the task just dismisses.

umerkhan-apple:
Investigating this issue.

PR #551, opened 2015-10-26
IOI # 517 - Image capture no permissions bug

umerkhan-apple:
The original implementation was implementing `requestedPermissions` on ORKImageCaptureStep which resulted in the delegate call `taskViewControllerDidFinish: withReason:` if camera permission was not granted. Removing that implementation and instead only requesting permission on the actual image step resolved part of the issue.

The second half of the bug was due to the fact that in the error handling of the camera permissions, the delegate callback `taskViewControllerDidFinish: withReason:` was being called also. This caused the image capture step to never appear on the screen and actually show the error message.

Removing both those factors allows the user to navigate through the steps and see the error messages so that they can resolve it by changing privacy in settings or by skipping the question.

Issue #517

### DIFF
--- a/ResearchKit/Common/ORKImageCaptureStep.m
+++ b/ResearchKit/Common/ORKImageCaptureStep.m
@@ -91,9 +91,4 @@
                         && ORKEqualObjects(self.accessibilityInstructions, castObject.accessibilityInstructions);
 }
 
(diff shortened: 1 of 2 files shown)


YuanZhu-apple:
Looks good to me!

PR #553, opened 2015-10-27
Switch `ORKLocationResult`'s answer from `ORKPlacemark` to `ORKLocation`

YuanZhu-apple:
Also created location questions task in ORKTest.

`ORKLocation` only includes `coordinate` and `addressString` which is strpping down version of  `MKPlacemark`.


ResearchKit/Common/ORKLocationSelectionView.m
```diff
@@ -232,27 +234,28 @@ - (void)reverseGeocodeAndDisplay:(CLLocation *)location {
...
 
-- (void)setAnswer:(id)answer {
+- (void)setAnswer:(ORKLocation *)answer {
     
     [_mapView removeAnnotations:_mapView.annotations];
     
-    if ([[answer class] isSubclassOfClass:[CLPlacemark class]]) {
-        _answer = [[ORKPlacemark alloc] initWithPlacemark:answer];
-    } else {
+    if (NO == [[answer class] isSubclassOfClass:[ORKLocation class]]) {
```
umerkhan-apple: Do we want to refactor to `!` notation? Or maybe at least change the `NO ==` to ` == NO` at the end?

ResearchKit/Common/ORKLocationSelectionView.m
```diff
@@ -232,27 +234,28 @@ - (void)reverseGeocodeAndDisplay:(CLLocation *)location {
...
         _answer = answer == ORKNullAnswerValue() ? ORKNullAnswerValue() : nil;
+    } else {
+        _answer = answer;
     }
```
umerkhan-apple: Or maybe change this whole if else into:
`_answer = ([[answer class] isSubclassOfClass:[ORKLocation class]]) ? answer : ORKNullAnswerValue()` ?

Do we need to store nil and ORKNullAnswerValue for some purpose?

ResearchKit/Common/ORKLocationSelectionView.m
```diff
@@ -266,8 +269,8 @@ - (void)setAnswer:(id)answer {
         }
         [_mapView setRegion:region animated:YES];
         
-        if (placemarkAnswer.addressDictionary) {
-            _textField.text = ABCreateStringWithAddressDictionary(placemarkAnswer.addressDictionary, NO);
+        if (location.address) {
```
umerkhan-apple: Should move the answer casting here since it doesn't get used until this if statement.

ResearchKit/Common/ORKResult.h
```diff
@@ -1254,6 +1254,26 @@ ORK_CLASS_AVAILABLE
 
 
 /**
+ The `ORKLocation` class represents the location addess obtained from a locaton question.
+ */
+@interface ORKLocation : NSObject <NSCopying, NSSecureCoding>
+
+- (instancetype)init NS_UNAVAILABLE;
+
```
umerkhan-apple: Do we want to add the other init methods here?

ResearchKit/Common/ORKResult.m
```diff
@@ -1920,6 +1918,62 @@ - (ORKStepResult *)stepResultForStepIdentifier:(NSString *)stepIdentifier {
...
+- (instancetype)initWithCoordinate:(CLLocationCoordinate2D)coordinate address:(NSString *)address {
+    self = [super init];
+    if (self) {
+        _coordinate = coordinate;
+        _address = [address copy];
+    }
+    return self;
+}
+
+- (instancetype)initWithPlaceMark:(CLPlacemark *)placeMark {
```
umerkhan-apple: placeMark to placemark

ResearchKit/Common/ORKLocationSelectionView.m
YuanZhu-apple: I like it!

ResearchKit/Common/ORKLocationSelectionView.m
YuanZhu-apple: `NO ==` prevent developer accidentally assign it to NO, like  `value == NO` =>  `value = NO`

ResearchKit/Common/ORKResult.m
```diff
@@ -1920,6 +1918,62 @@ - (ORKStepResult *)stepResultForStepIdentifier:(NSString *)stepIdentifier {
...
+    if (self) {
+        _coordinate = placeMark.location.coordinate;
+        _address = [ABCreateStringWithAddressDictionary(placeMark.addressDictionary, NO) copy] ;
+    }
+    return self;
+}
+
+- (id)copyWithZone:(NSZone *)zone {
+    // This object is not mutable
+    return self;
```
umerkhan-apple: You inherit NSCopying but this just returns self. Is there a way to avoid that?

YuanZhu-apple: If something is not mutable, then a `real copy` is not requried

ResearchKit/Common/ORKResult_Private.h
```diff
@@ -91,5 +92,14 @@ ORK_CLASS_AVAILABLE
 
 @end
 
+@interface ORKLocation ()
+
+- (instancetype)initWithCoordinate:(CLLocationCoordinate2D)coordinate address:(NSString *)address;
+
+- (instancetype)initWithPlaceMark:(CLPlacemark *)placeMark;
```
umerkhan-apple: refactor placeMark to placemark

Testing/ORKTest/ORKTest/MainViewController.m
```diff
@@ -3598,4 +3602,39 @@ - (void)updateProgress:(CGFloat)progress OfWaitTask:(ORKWaitStepViewController *
...
+    step3.answerFormat = locationAnswerFormat;
+    [steps addObject:step3];
+    
+    ORKCompletionStep *step4 = [[ORKCompletionStep alloc] initWithIdentifier:@"locationTask.step4"];
+    step4.title = @"Survey Complete";
+    [steps addObject:step4];
+    
+    ORKOrderedTask *locationTask = [[ORKOrderedTask alloc] initWithIdentifier:LocationTaskIdentifier steps:steps];
+    return locationTask;
+}
```
umerkhan-apple: Did you want to leave the location question at the end of Selection Survey still?

YuanZhu-apple: Remove it from selection survey.

ResearchKit/Common/ORKFormItemCell.m
```diff
@@ -47,7 +47,6 @@ OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 #import "ORKEligibilitySelectionView.h"
 #import "ORKSubheadlineLabel.h"
 #import "ORKLocationSelectionView.h"
-#import "ORKPlacemark.h"
 #import <AddressBookUI/AddressBookUI.h>
```
umerkhan-apple: Don't need this import anymore.

ResearchKit/Common/ORKLocationSelectionView.m
```diff
@@ -43,8 +43,8 @@ OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 #import "ORKAnswerTextField.h"
```
umerkhan-apple: Don't need AddressBook import anymore.

ResearchKit/Common/ORKSurveyAnswerCellForLocation.m
```diff
@@ -38,7 +38,6 @@ OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 #import "ORKAnswerFormat_Internal.h"
```
umerkhan-apple: Don't need AddressBook import anymore.

samples/ORKCatalog/ORKCatalog/Results/ResultTableViewProviders.swift
```diff
@@ -376,20 +376,15 @@ class LocationQuestionResultTableViewProvider: ResultTableViewProvider {
     
     override func resultRowsForSection(section: Int) -> [ResultRow] {
         let questionResult = result as! ORKLocationQuestionResult
-        let placemark = questionResult.locationAnswer
-        
-        var rows = super.resultRowsForSection(section) + [
+        let location = questionResult.locationAnswer
```
umerkhan-apple: Don't need AddressBook import in this file anymore.

umerkhan-apple:
Overall questions:
- Do we want to return an address string? I think we should only return coordinates, and you should convert those coordinates into an address string yourself. The main reason for this would be that our address string could become outdated as MapKit grows. Maybe a given lat/long right now queries a specific address, but with the next update to MapKit, it could possibly query a different address (mainly due to information error initially). By providing a string address as part of answer, we are almost taking guarantee that the lat/long will always be that address, which is not necessarily true.
- Currently the use can type in an address (or the textfield is populated using current location). This can be extremely limiting for questions like: What is your home address? Currently, I type in my home address and it points to the correct building, but I am not able to specify an apartment number. As soon as I specify an apartment number, it replaces it with the building address (minus my apartment number). We need to have a way to obtain the location (which is lat/long) and then save the text that is in the textfield as the inputtedAnswer versus saying the address (generated using the lat/long).

umerkhan-apple:
r=m3!

PR #556, opened 2015-11-01
ORKTest: improve some class and method names

rsanchezsaez:
This minor PR improves some class and method names by following naming conventions.

### DIFF
--- a/Testing/ORKTest/ORKTest/MainViewController.m
+++ b/Testing/ORKTest/ORKTest/MainViewController.m
@@ -3723,25 +3723,20 @@ stepViewControllerWillAppear:(ORKStepViewController *)stepViewController {
     return waitTask;
 }
 
-- (void)updateProgress:(CGFloat)progress waitStepViewController:(ORKWaitStepViewController *)viewController {
+- (void)updateProgress:(CGFloat)progress waitStepViewController:(ORKWaitStepViewController *)waitStepviewController {
     if (progress <= 1.0) {
-        [viewController setProgress:progress animated:true];
-        
+        [waitStepviewController setProgress:progress animated:true];
         double delayInSeconds = 0.1;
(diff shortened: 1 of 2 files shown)


Testing/ORKTest/ORKTest/MainViewController.m
```diff
@@ -3723,17 +3723,17 @@ - (ORKOrderedTask *)makeWaitingTask {
...
 
-- (void)updateProgress:(CGFloat)progress OfWaitTask:(ORKWaitStepViewController *)viewController {
+- (void)updateProgress:(CGFloat)progress waitStepViewController:(ORKWaitStepViewController *)viewController {
     if (progress <= 1.0) {
         [viewController setProgress:progress animated:true];
         
         double delayInSeconds = 0.1;
         dispatch_time_t popTime = dispatch_time(DISPATCH_TIME_NOW, (int64_t)(delayInSeconds * NSEC_PER_SEC));
-        __weak ORKWaitStepViewController *vc = viewController;
+        __weak ORKWaitStepViewController *weakViewController = viewController;
```
YuanZhu-apple: Not in your code, but is this `__weak` conversion necessary? 
I didn't see any possible retaining loop between the block and `waitStepViewController`.

rsanchezsaez: Done. I also removed the `__weak` conversion for `self` (since the block is temporary and not permanently retained locally) and improved the float comparison.

umerkhan-apple:
@rsanchezsaez Could you also update the login and verification class names in ORKCatalog?

rsanchezsaez:
Done.

Just curious, why did you use a *subclassing pattern* instead of a *delegation pattern* for these two view controllers? Wouldn't *delegation* be cleaner?

YuanZhu-apple:
@rsanchezsaez 
Delegation pattern require developer to monitor: `taskViewController:stepViewControllerWillAppear:` and assign its delegate to the new step view controller, which is not that obvious for the beginners.

Issue #552, opened 2015-10-26
Localization issue in ORKCatalog


umerkhan-apple:
#### Localization bug in ORKCatalog

If you change language and locale and open ORKCatalog, it remains in English.
The strings that are in ORKCatalog do not have translations, so that is expected behavior.
But the strings in RK Framework do have translations, so those should convert.

##### Steps to reproduce:
1. Open `Settings`
2. Tap `General`
3. Tap `Language & Region`
4. Change Language (Anything apart from English)
5. Open `ORKCatalog`
6. Tap `Consent-Obtaining Example`

*__Note:__ The content is still in English*

##### Sanity check:

1. Repeat steps 1-4 from above 
2. Open `ORKTest`
3. Tap `Consent`

*__Note:__ The content is in selected localized language*

PR #559, opened 2015-11-05
ORKCatalog Localization Files

umerkhan-apple:
Adding placeholder localization files for ORKCatalog.

Currently, ORKCatalog does not pull in the translations from RKFramework's strings files, because it doesn't support those languages. This fix updates the configuration for ORKCatalog to support all the languages supported in RK Framework and create placeholder files for future translation.

Issue #552


YuanZhu-apple:
r=me.

PR #564, opened 2015-11-17
CorrectDestionation

md0u80c9:
Changed all instances of Destionation to Destination


rsanchezsaez:
Thanks Andrew. Fix looks good!

YuanZhu-apple:
Hey @md0u80c9 can you resubmit this PR against `convergence` branch? 
So this fix can be included in next release. Thanks!

md0u80c9:
Is there a way to do this from GitHub or Xcode without re-doing it?

YuanZhu-apple:
Can you create a new PR and select against `convergence`?

md0u80c9:
I've just checked the convergence branch - the change is on code which is only on the Master branch at present, so I think the PR only needs to apply to master.

YuanZhu-apple:
OK, I will accept it and cherry pick it to `convergence`.

Issue #566, opened 2015-11-20
ORKRegistrationStep passcodeValidationRegex is case-insensitive


rliljest:
Currently the NSRegularExpression used with ORKAnswerFormat uses the NSRegularExpressionCaseInsensitive option. This makes it impossible to make a case-sensitive passcodeValidationRegex in ORKRegistrationStep (which could be used to enforce concepts such as "password must contain one lowercase and one uppercase letter").

Potential fix I have ready locally (creating this issue first per contribution guidelines):
- Create new NSRegularExpressionOptions property in ORKAnswerFormat.h for ORKTextAnswerFormat class
- Initialize to NSRegularExpressionCaseInsensitive in commonInit (for backwards compatibility)
- Update ORKTextAnswerFormat methods validateParameters, copyWithZone:, and isTextRegexValidWithString: to use new property

From there we could either
- Assume that the ORKRegistrationStep passcodeValidationRegex will always want to be case-insensitive (potentially breaking existing code that assume case insensitivity)
- Expose a new transparent NSRegularExpressionOptions property on ORKRegistrationStep that is backed by the ORKTextAnswerFormat NSRegularExpressionOptions, and leave it to the caller to set something other than NSRegularExpressionCaseInsensitive

YuanZhu-apple:
Thanks for the proposal!

I agree we should expose `NSRegularExpressionOptions`  in `ORKAnswerFormat.h` and  `ORKRegistrationStep.h` to replace current pattern string.
Also please take care of `ORKESerialization.h`.

umerkhan-apple:
Fixed in #897

PR #567, opened 2015-11-20
Expose NSRegularExpressionOptions for ORKTextAnswerFormat, ORKRegistrationStep

rliljest:
For Issue #566 

Currently text field validation can take a regex for validation, but the NSRegularExpression always uses the NSRegularExpressionCaseInsensitive option, making it impossible to have case sensitive password validation (e.g. password must contain one uppercase letter and one lowercase letter).

This change exposes a new NSRegularExpressionOptions property on ORKTextAnswerFormat (and transparently via ORKRegistrationStep) which could be used to change that option. By default the new validationRegexOptions is set to NSRegularExpressionCaseInsensitive for backwards compatibility.

### DIFF
--- a/ResearchKit/Common/ORKAnswerFormat.m
+++ b/ResearchKit/Common/ORKAnswerFormat.m
@@ -1896,6 +1896,7 @@ static NSArray *ork_processTextChoices(NSArray<ORKTextChoice *> *textChoices) {
     _spellCheckingType = UITextSpellCheckingTypeDefault;
     _keyboardType = UIKeyboardTypeDefault;
     _multipleLines = YES;
+    _validationRegexOptions = NSRegularExpressionCaseInsensitive;
 }
 
 - (instancetype)initWithMaximumLength:(NSInteger)maximumLength {
@@ -1938,7 +1939,7 @@ static NSArray *ork_processTextChoices(NSArray<ORKTextChoice *> *textChoices) {
     
     NSError *error;
     if (self.validationRegex && ![[NSRegularExpression alloc] initWithPattern:self.validationRegex
-                                                                      options:NSRegularExpressionCaseInsensitive
+                                                                      options:self.validationRegexOptions
                                                                         error:&error]) {
         @throw [NSException exceptionWithName:NSInvalidArgumentException
                                        reason:@"Validation regex is not valid."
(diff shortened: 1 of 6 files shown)


umerkhan-apple:
Great catch! Looks good to me! :+1:

YuanZhu-apple:
@rliljest You patch looks good. But I think it would be better to let the the answer format just takes a `NSRegularExpression` object, it would a simpler interface and the developer knows what exactly going to be evaluated. 

`NSRegularExpression` natively support `NSCoding`, but somehow we have to deal it in ORKESerialization.m a little bit more(pack the two values into a dictionary). 

Can you make a quick turn around to us and make it against convergence branch? If you are busy, then we can take it over and make sure it in our next release.

Thanks a lot!

rliljest:
Switching to use the NSRegularExpression object itself does sound like a good idea. My only concern is that we would need to change the -[ORKTextAnswerFormat initWithValidationRegex:invalidMessage:] initializer from a designated initializer to a failable convenience initializer. A new designated initializer that takes an NSRegularExpression -[ORKTextAnswerFormat initWithExpression:invalidMessage:] could take it's place.

If removing the designated status of that initializer is OK then I can go ahead and make that change on my convergence branch and send a pull request. I am traveling starting tomorrow so if I don't have the pull request up tonight the earliest it will be available is Sunday night

umerkhan-apple:
@rliljest I'm not sure if it would be good to have two initializers that essentially perform the same logic. We might need to find a consolidated solution. @YuanZhu-apple how do you feel about adopting `NSRegularExpression` object and just making it a convenience initializer.

umerkhan-apple:
@YuanZhu-apple Friendly ping! :)

umerkhan-apple:
@rliljest Could you resolve the merge conflicts? Thanks!

ResearchKit/Common/ORKAnswerFormat.m
```diff
@@ -1896,6 +1896,7 @@ - (void)commonInit {
     _spellCheckingType = UITextSpellCheckingTypeDefault;
     _keyboardType = UIKeyboardTypeDefault;
     _multipleLines = YES;
+    _validationRegexOptions = NSRegularExpressionCaseInsensitive;
```
umerkhan-apple: default should be 0 (case sensitive)

umerkhan-apple:
@rliljest Do you plan on updating this PR or should we close it for now?

rsanchezsaez:
I'll try to make the requested changes when I have time. :)

umerkhan-apple:
@rsanchezsaez Thanks Ricardo! 👍

rsanchezsaez:
Changes done on PR https://github.com/ResearchKit/ResearchKit/issues/952.

Issue #572, opened 2015-11-23
ORKLoginStepViewController is not publically visible


tomihisaw:
Hi,
In order to use the ORKLoginStep we must use the initializer:

```
- (instancetype)initWithIdentifier:(NSString *)identifier
                             title:(nullable NSString *)title
                              text:(nullable NSString *)text
          loginViewControllerClass:(Class)loginViewControllerClass;
```

Further, the "loginViewController" class is required to be a subclass of ORKLoginStepViewController.

However, in the current master branch ORKLoginStepViewController does not have "ORK_CLASS_AVAILABLE" enabled. Thus, if the project is compiled with "Symbols Hidden by Default" (the default project setting for Release builds), it is not possible to subclass due to a link error finding the symbols. I believe this might have been done accidentally when refactoring since the class previously had ORK_CLASS_AVAILABLE and this convention is used in other publicly available classes.

Would it be possible to add ORK_CLASS_AVAILABLE back or is there some other way to subclass? I could be happy to put in a pull request with this minor change too.

Thanks

YuanZhu-apple:
Good catch. Please put up a PR to add ORK_CLASS_AVAILABLE. Thanks!

celteeka:
Hi! Unrelated to ORKLoginStepViewController but somehow related to ORK_CLASS_AVAILABLE, I'm having sort of the same issue with a subclass of ORKQuestionStepCellHolderView which is defined in ORKCustomStepView.h 
The reason I'm subclassing ORKQuestionStepCellHolderView instead of ORKQuestionStepCustomView is because if I don't I then get an 'invalid selector sent to...' error in ORKQuestionStepViewController#viewDidAppear 

`ORKSurveyAnswerCellForPicker *cell = (ORKSurveyAnswerCellForPicker *)[(ORKQuestionStepCellHolderView *)_questionView.questionCustomView cell];`

The 'cell' property is defined in ORKQuestionStepCellHolderView instead of the super class ORKQuestionStepCustomView. 
Anyhow, I was getting link errors when attempting to archive my project. I was able to fix it if I add ORK_CLASS_AVAILABLE to ORKQuestionStepCellHolderView like this:
`ORK_CLASS_AVAILABLE
@interface ORKQuestionStepCellHolderView : ORKQuestionStepCustomView
`

I was just wondering if this is the correct approach (I mean, if this is a bug) or if I should use a different approach.

Thanks!!

YuanZhu-apple:
[...]

PR #573, opened 2015-11-23
Allow ORKLoginStepViewController to be subclassed by making the class…

tomihisaw:
… symbol visible

As discussed in issue: https://github.com/ResearchKit/ResearchKit/issues/572


YuanZhu-apple:
Thanks!

PR #575, opened 2015-11-25
Fixed build error in ResultTableViewProviders.swift caused by an outdated property.

UberJason:
Replaced the field "address" with "userInput" in the ORKCatalog project which was causing a build error.


YuanZhu-apple:
Thanks!

PR #581, opened 2015-12-15
Added method to create a copy of a step with a new identifier

syoung-smallwisdom:
This will allow a developer to copy a private subclass of ORKStep with a new identifier.

For an example usage, see:

https://github.com/Sage-Bionetworks/AppCore/blob/master/APCAppCore/APCAppCore/Library/Categories/ORKOrderedTask%2BAPCHelper.m

```
ORKInstructionStep *instructionStep = [tappingInstructionStep copyWithIdentifier:
    [NSString stringWithFormat:@"%@.%@", APCTapInstructionStepIdentifier, handIdentifier]];
ORKActiveStep *activeStep = [tappingStep copyWithIdentifier:
    [NSString stringWithFormat:@"%@.%@", APCTapTappingStepIdentifier, handIdentifier]];
```

In this example, 'ORKTappingIntervalStep' is a private class within the ResearchKit SDK so it cannot be instantiated directly with a new identifier used to differentiate tapping with the left hand from the right hand for the case where the research study requires a bilateral tapping test.

### DIFF
--- a/ResearchKit/Common/ORKStep.h
+++ b/ResearchKit/Common/ORKStep.h
@@ -85,7 +85,7 @@ ORK_CLASS_AVAILABLE
 /**
  Returns a copy of this step initialized with the specified identifier.
  
- @param identifier   The unique identifier of the step.
+ @param identifier   The unique identifier for the new step to be returned.
  
  @return A new step.
  */
--- a/ResearchKit/Common/ORKStep.h
+++ b/ResearchKit/Common/ORKStep.h
@@ -83,6 +83,15 @@ ORK_CLASS_AVAILABLE
 - (instancetype)initWithCoder:(NSCoder *)aDecoder NS_DESIGNATED_INITIALIZER;
 
 /**
+ Returns a copy of this step initialized with the specified identifier.
+ 
+ @param identifier   The unique identifier of the step.
+ 
+ @return A new step.
+ */
+- (instancetype)copyWithIdentifier:(NSString *)identifier;
+
+/**
  A short string that uniquely identifies the step within the task.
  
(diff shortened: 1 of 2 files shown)


ResearchKit/Common/ORKStep.m
```diff
@@ -59,6 +59,12 @@ - (Class)stepViewControllerClass {
     return [[self class] stepViewControllerClass];
 }
 
+- (instancetype)copyWithIdentifier:(NSString *)identifier {
+    ORKStep *step = [self copy];
+    step->_identifier = identifier;
```
YuanZhu-apple: Can you add `ORKThrowInvalidArgumentExceptionIfNil(identifier);` here to prevent a nil `identifier`?
And do `step->_identifier = [identifier copy];`?

syoung-smallwisdom:
@YuanZhu-apple 
Changes have been pushed. Please let me know if you would like for me to squish this into a single commit.

ResearchKit/Common/ORKStep.h
```diff
@@ -83,6 +83,15 @@ ORK_CLASS_AVAILABLE
 - (instancetype)initWithCoder:(NSCoder *)aDecoder NS_DESIGNATED_INITIALIZER;
 
 /**
+ Returns a copy of this step initialized with the specified identifier.
+ 
+ @param identifier   The unique identifier of the step.
```
YuanZhu-apple: How about `The unique identifier for the new step to be returned`?

syoung-smallwisdom:
@YuanZhu-apple 
Done: I went ahead and squished the change to the header.

YuanZhu-apple:
r=me.

PR #583, opened 2015-12-18
Improve Passcode View

p2:
This PR uses `instancetype` instead of `id`, helping when using from Swift, and adds the option to manually set a passcode.

The latter is useful when moving from AppCore's passcode entry (or a custom one). This PR simply moves the existing save-to-passcode method to a generic implementation, internally exposed, and adds a public method to `ORKPasscodeViewController` that uses the same method. Ideally all passcode storage methods would go to a custom subclass instead of being hosted by `ORKPasscodeStepViewController`.


ResearchKit/Common/ORKPasscodeViewController.h
```diff
@@ -129,6 +129,17 @@ ORK_CLASS_AVAILABLE
...
+/**
+ This method allows to force the passcode to be a certain value. You only want to use this capability when you're moving your existing users
+ from a custom passcode entry mechanism to the one provided by ResearchKit.
+ 
+ This method will raise an exception if `passcode` is nil.
+ 
+ @param passcode          The passcode to store
+ @param touchIdEnabled    Whether TouchId will be available during passcode entry
+ */
++ (void)forcePasscode:(NSString *)passcode withTouchIdEnabled:(BOOL)touchIdEnabled;
```
umerkhan-apple: Do we want to make this a public method? 

I really like the use case described here. I just think it might be better if the developer uses the keychain wrapper directly to migrate passcodes.

Let me know what you think?

umerkhan-apple:
PR looks good to me, with one minor concern.

ResearchKit/Common/ORKPasscodeViewController.h
p2: I chose to make this method public because if you use ORKKeychainWrapper directly, you need to know several internal keys. This is how the passcode gets stored, showing the three keys `KeychainDictionaryPasscodeKey `, `KeychainDictionaryTouchIdKey` and `forKey:PasscodeKey ` that one must use correctly:

```obj-c
NSDictionary *dictionary = @{
    KeychainDictionaryPasscodeKey : [passcode copy],
    KeychainDictionaryTouchIdKey : @(touchIdEnabled)
};
NSError *error;
[ORKKeychainWrapper setObject:dictionary forKey:PasscodeKey error:&error];
```

We could think of making a public method on the keychain wrapper, but since these keys live inside `ORKPasscodeStepViewController_Internal.h` we'd need to move those as well. Making it a class-level public method on the passcode (step) view controllers seemed like the least intrusive way to me, but I'm certainly open to other solutions.

umerkhan-apple: I see. My only concern was that the method might be misunderstood and used incorrectly.
That being said, it is well-documented.
Unless anyone else has any other solutions, I am okay to proceed with this implementation.

umerkhan-apple:
@YuanZhu-apple Can you review also?

ResearchKit/Common/ORKPasscodeViewController.h
YuanZhu-apple: @p2 I would suggest move this API to the private interface(need to create) of `ORKPasscodeViewController`, at least we can hide it from main interface.

p2: I need to call this method from outside of ResearchKit, at startup in the app delegate, where it detects the update that switched to RK's passcode entry. So there needs to be a public method somewhere, just like `removePasscodeFromKeychain` which already lives in this class. Where should it be moved to?

YuanZhu-apple:
r=me.

Issue #402, opened 2015-08-26
Can't subclass ORKTextChoiceAnswerFormat 


brendenwest:
I've looked at subclassing ORKTextChoiceAnswerFormat to support features not yet available in ResearchKit. For example, to add a property indicating if one choice in the list is exclusive:

    class testFormat :  ORKTextChoiceAnswerFormat {
        var exclusive : AnyObject?
        override init(style: ORKChoiceAnswerStyle, textChoices: [AnyObject]) {
            super.init(style: style, textChoices: textChoices)
        }

        required init(coder aDecoder: NSCoder) {
            super.init(coder: aDecoder)
        }
    }

This approach fails on rendering because ORKChoiceAnswerFormatHelper doesn't recognize the new class. See below. Can the helper class be updated to allow subclasses?

2015-08-26 10:20:53.417 RKtest[1720:39735] *** Terminating app due to uncaught exception 'NSGenericException', reason: 'Not a valid answerformat for this helper.'
*** First throw call stack:
(
	0   CoreFoundation                      0x00000001103fac65 __exceptionPreprocess + 165
	1   libobjc.A.dylib                     0x0000000110093bb7 objc_exception_throw + 45
	2   ResearchKit                         0x0000000111f84ab3 -[ORKChoiceAnswerFormatHelper initWithAnswerFormat:] + 435
	3   ResearchKit                         0x0000000111fab865 -[ORKTextChoiceCellGroup

YuanZhu-apple:
@brenbob `ORKTextChoice` already has `exclusive` variable.

brenbob:
Thanks. Any idea why that property is not listed in the RK docs? 

http://researchkit.org/docs/Classes/ORKTextChoice.html#

rsanchezsaez:
@brenbob I think that's because this enhancement was added after the last official  *ResearchKit* release. Presumably it'll be added to the docs along with the next release (which should be coming quite soon).

If you want to be on the cutting-edge without waiting for new releases, to see what's new you'll have to skim over the merged *pull requests* or the *commits* on the `master` branch. ;-)

brenbob:
I've run into a ResearchKit versioning problem with this. My local copy has the 'exclusive' property because I pulled from the master branch a few weeks ago. Now the master branch is broken and building fails with errors relating to generics and won't build. But the stable branch lacks the 'exclusive' property. 

What's the timeline for next release, or at least a build-able master branch?

[...]

PR #592, opened 2016-01-05
[ORKChoiceAnswerFormatHelper] Relax class checking so choice answer formats can be subclassed

rsanchezsaez:
This PR deals with [Issue #402](https://github.com/ResearchKit/ResearchKit/issues/402).


YuanZhu-apple:
r=me.

PR #597, opened 2016-01-07
Samples - ORKCatalog/Charts/ChartListViewController - set prop on dis…

evliu:
…creteGraphChartView instead of lineGraphChartView


YuanZhu-apple:
Looks right to me.
@rsanchezsaez Let me know if you agree.

rsanchezsaez:
Yeah, it was a mistake on my end. :+1:

Issue #600, opened 2016-01-12
ORKDiscreteGraphChartView crashes upon instantiation in playground


scott-ce:
ORKDiscreteGraphChartView crashes with an array index out of bounds error if layout is triggered before a data source is set. This condition should probably be handled more gracefully, especially because it prevents working with ORKDiscreteGraphChartView in a playground: as soon as the view is instantiated, the playground attempts to generate a quick look version before evaluating subsequent lines.

# Version

This is with the newly released 1.3, installed today via Cocoapods. Xcode 7.2 (7C68)

# Steps to Reproduce

1. Create a playground in a workspace that includes the ResearchKit framework.
2. Type the following line:
```
let graph = ORKDiscreteGraphChartView(frame: CGRect(x: 0, y: 0, width: 300, height: 300))
```

# Expected Results

The playground should compile, run, instantiate the chart view object, and display a reference to it.

# Actual Results

```
2016-01-12 14:21:35.322 APC Discrete Graph[92085:5477209] *** Terminating app due to uncaught exception 'NSRangeException', reason: '*** -[__NSArrayM objectAtIndex:]: index 0 beyond bounds for empty array'
*** First throw call stack:
(
	0   CoreFoundation                      0x000000010b98be65 __exceptionPreprocess + 165
	1   libobjc.A.dylib                     0x000000010d25bdeb objc_exception_throw + 48
	2   CoreFoundation                      0x000000010b86f404 -[__NSArrayM objectAtIndex:] + 212
	3   ResearchKit                         0x000000011a050cd8 -[ORKGraphChartView normalizedCanvasPointsForPlotIndex:canvasHeight:] + 88
	4   ResearchKit                         0x000000011a04c073 -[ORKGraphChartView updateYAxisPoints] + 259
	5   ResearchKit                         0x000000011a04bf37 -[ORKGraphChartView layoutSubviews] + 2999
	6   UIKit                               0x000000010bdef4a3 -[UIView(CALayerDelegate) layoutSublayersOfLayer:] + 703
	7   QuartzCore                          0x00000001124c559a -[CALayer layoutSublayers] + 146
	8   QuartzCore                          0x00000001124b9e70 _ZN2CA5Layer16layout_if_neededEPNS_11TransactionE + 366
	9   QuartzCore                          0x00000001124c54d3 -[CALayer layoutIfNeeded] + 159
	10  QuartzCore                          0x00000001124c2673 -[CALayer renderInContext:] + 139
[...]

PR #602, opened 2016-01-14
[Charts] Fix crash when displaying graph chart with no data source 

rsanchezsaez:
This PR fixes [Issue #600](https://github.com/ResearchKit/ResearchKit/issues/600), and also adds `reloadData` methods to both chart types.


YuanZhu-apple:
r=me.

PR #611, opened 2016-01-26
ORKGraphChartView - Remove the "No Data" label when there is data

scott-ce:
The "No Data" label continued to appear in ORKGraphChartView after data was loaded because of a reversed boolean check.

image


rsanchezsaez:
Good catch! Reviewed.

YuanZhu-apple:
@scott-ce Thanks!

Issue #610, opened 2016-01-26
Building master via Cocoapods fails


scott-ce:
I tried updating to the latest master in order to get the updates for PR #602 (we were running into the same layout issue displaying a chart in a table cell before the data source can fetch HealthKit data). Unfortunately, I'm unable to compile RK master because of a header issue: ORKWaitStepView.h fails to import ORKVerticalContainerView.h, because the latter isn't exposed as a public header.

In master, ORKWaitStepView.h appears in ResearchKit-umbrella.h. In release 1.3.0, it does not.

It appears this issue was introduced in commit 8c86b8d8, which updated the attributes of ORKWaitStepView.h in project.pbxproj and made it private. If I understand correctly, find_headers.rb determines the headers for the podspec by including the public & private headers. Now that ORKWaitStepView.h is private, it is included in the umbrella header, but ORKVerticalContainerView.h is not — so our build fails.

# Versions

Xcode 7.2 (7C68) building against commit 8c86b8d8 using Cocoapods 0.39.0.

# Steps to Reproduce

1. Create an application with a Podfile containing this dependency (54927408 is the latest master commit as I write this):
```
pod 'ResearchKit', :git => 'https://github.com/ResearchKit/ResearchKit.git', :commit => '54927408'
```
2. Run `pod install`.
3. Attempt to build the application.

# Expected Results

App builds successfully.

# Actual Results

```
<module-includes>:1:9: note: in file included from <module-includes>:1:
#import "/Users/USER/Library/Developer/Xcode/DerivedData/APPNAME-ewpkpsoawrrntgfcyxsmjufowntg/Build/Products/Debug-iphonesimulator/ResearchKit.framework/Headers/ResearchKit-umbrella.h"
        ^
/Users/USER/Library/Developer/Xcode/DerivedData/APPNAME-ewpkpsoawrrntgfcyxsmjufowntg/Build/Products/Debug-iphonesimulator/ResearchKit.framework/Headers/ResearchKit-umbrella.h:56:9: note: in file included from /Users/USER/Library/Developer/Xcode/DerivedData/APPNAME-ewpkpsoawrrntgfcyxsmjufowntg/Build/Products/Debug-iphonesimulator/ResearchKit.framework/Headers/ResearchKit-umbrella.h:56:
#import "ORKWaitStepView.h"
        ^
[...]

PR #612, opened 2016-01-26
Revert ORKWaitStepView.h to Project visibility

scott-ce:
This PR addresses #610. I confirmed that reverting ORKWaitStepView.h to Project visibility resolves the current build issue.

Note that along with ORKWaitStepView.h, commit 8c86b8d also switched these headers from Project to Private visibility:

* ORKReviewStep_Internal.h
* ORKReviewStepViewController_Internal.h
* ORKReviewStepViewController.h
* ORKWaitStepViewController.h

I wasn't sure if it any of these other headers should be exposed, and they don't import other private headers and cause build errors, so I didn't revert them. However, you may want to consider reverting them if they shouldn't be exposed.


YuanZhu-apple:
@scott-ce Thanks!

Can you also switch these two headers to `Project`? They shouldn't be exposed.
- ORKReviewStep_Internal.h
- ORKReviewStepViewController_Internal.h

I think it is OK to leave the other two headers as `Private`.

scott-ce:
Sure, done.

YuanZhu-apple:
Cool!

PR #626, opened 2016-02-26
About ORKQuestionType for ORKTextAnswerFormat?

weijentu:
At first, I'm bit surprised that question type for ``ORKEmailAnswerFormat`` is ``ORKQuestionTypeNone``. Owing to an ``ORKEmailAnswerFormat`` object produces an ``ORKTextQuestionResult`` object and it's implied with ``ORKTextAnswerFormat`` as well. Therefore, just little wondering if it could be better to set its question type to ``ORKQuestionTypeText``, but not ``ORKQuestionTypeNone``


rsanchezsaez:
Hey, this change sounded reasonable, why did you close the PR? :)

weijentu:
Hi there, I’m very glad you think this change sounded reasonable. Yesterday, I was suddenly little worried about that if it’s appropriate to have a suggestion like this as a pull request. That’s why I closed it >_<“.

YuanZhu-apple:
Thanks!

Issue #381, opened 2015-08-19
Vertical slider label no longer line-breaks


p2:
We have longer labels on the vertical slider. The bottom one no longer breaks into multiple lines, the top one still does.

This is a regression some time after c856b29933af99bcee91dd7af733a90cc09885ab.

![screen shot 2015-08-18 at 11 48 19](https://cloud.githubusercontent.com/assets/358485/9349073/a839aede-4640-11e5-9264-c1e7f9b6f9e0.png)

umerkhan-apple:
https://github.com/ResearchKit/ResearchKit/pull/407

PR #629, opened 2016-03-05
Allow 2 lines of text on vertical scale sliders

evands:
There's plenty of space for this to fit comfortably, and it allows much more flexibility in defining answer choices.

### DIFF
--- a/ResearchKit/Common/ORKScaleSliderView.m
+++ b/ResearchKit/Common/ORKScaleSliderView.m
@@ -88,7 +88,7 @@
                 ORKScaleRangeLabel *stepLabel = [[ORKScaleRangeLabel alloc] initWithFrame:CGRectZero];
                 stepLabel.text = textChoice.text;
                 stepLabel.textAlignment = NSTextAlignmentLeft;
-                stepLabel.numberOfLines = 2;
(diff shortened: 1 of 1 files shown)


p2:
Currently, the label should allow any number of lines, is this not working for you? This fix would limit the label to 2 lines max, which would be less that we'd like in our case, see #381.

(nice to hear from you here, Evan!)

evands:
Hey, Pascal! Nice to see you, too. Small world. :)
I'm not sure how it was working previously, but the default for numberOfLines on a UILabel is 1.  Based on your case, I've changed it to 0 in the branch, which is 'unlimited'... ideally the slider would size to allow the largest text w/ unlimited lines, but that's premature flexibility I'd say.
with HEAD
setting numberOfLines

p2:
You're right of course, I've missed that you're talking about the _continuous_ scale, I am talking about the min/max scale. See screenshot, those labels wrap to as many lines as one desires.

So I agree with your initial choice, 2 lines is definitely nice to have here. If longer text is needed one should probably choose a different answer format.

screen shot 2016-03-09 at 09 30 34

YuanZhu-apple:
@evands Is this issue resolved?

PR #620, opened 2016-02-08
[ORKGraphChartView] Fixes and small improvements

rsanchezsaez:
This PR includes some fixes and small improvements to the graph chart views:
- Fix error in accessibility string key (`AX_CHART_POINT_%@` -> `AX_GRAPH_POINT_%@`).
- Fix scrubber not being correctly positioned for last data point in discrete graphs.
- Make the discrete graph scrubber to snap on the actual chosen plot index for graphs with more than one plot (instead of scrubbing to the center of the horizontal axis).
- Rename `ORKRangedPoint` to `ORKFloatRange`, and update documentation accordingly. The new name is more suitable, as this class on its own doesn't have any notion of being point.
- Fix cosmetic typos in `ORKSample` graph data source names.

Issue #345, opened 2015-08-03
Implement stacked bar graphs


rsanchezsaez:
This is a follow-on issue on [pull request #178](https://github.com/ResearchKit/ResearchKit/pull/178), which adds pie charts and line graphs to *ResearchKit*.

*AppCore* had a [PR implementing stacked bar graphs](https://github.com/ResearchKit/AppCore/pull/120) submitted recently, but it cannot be trivially merged into RK because of code divergences. It'd be nice to port this change over to RK.

rsanchezsaez:
This feature has been implemented and merged. Closing issue.

PR #633, opened 2016-03-08
[Charts] Implement ORKBarGraphChart

rsanchezsaez:
This PR implements the new *bar graph chart* type.

It includes [PR #620](https://github.com/ResearchKit/ResearchKit/pull/620), since I developed it from the tip of that branch.

The new `ORKBarGraphChart` subclass is an (extensively rewritten) port of `APCBarGraphView` as detailed in [Issue #345](https://github.com/ResearchKit/ResearchKit/issues/345). I implemented and tested accessibility functionality for it as well.

Each data point in a single *bar graph chart* is represented by one or more float values, so the new ~~`ORKFloatStack`~~ `ORKValueStack` class has been introduced. Both ~~`ORKFloatStack` and `ORKFloatRange`~~ `ORKValueStack` and `ORKValueRange` (previously known as `ORKRangedPoint`) conform to the `ORKGraphChartType` protocol for code reuse.

To support this, `ORKGraphChartDataSource` has become and abstract protocol and now has two concrete subprotocols: ~~`ORKFloatRangeGraphChartDataSource` and `ORKFloatStackGraphChartDataSource`~~ `ORKValueRangeGraphChartDataSource` and `ORKValueStackGraphChartDataSource`.

The new class hierarchy for *graph charts* is as follows:
```
ORKGraphChartView (abstract)
 |
 |-- ORKValueRangeGraphChartView <ORKValueRangeGraphChartDataSource> (abstract)
 |     |    
 |     |-- ORKLineGraphChartView
 |     |-- ORKDiscreteGraphChartView
 |
 |-- ORKBarGraphChartView <ORKValueStackGraphChartDataSource>
```

See the expanded commit messages for more details on the implementation.

simulator screen shot 15 mar 2016 11 46 45


ResearchKit/Charts/ORKBarGraphChartView.m
```diff
@@ -0,0 +1,225 @@
...
+ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+
+#import "ORKBarGraphChartView.h"
+#import "ORKGraphChartView_Internal.h"
+#import "ORKChartTypes.h"
+
+
+const CGFloat BarWidth = 10.0;
```
YuanZhu-apple: Add static ?

rsanchezsaez: Oops. :)

ResearchKit/Charts/ORKChartTypes.h
```diff
@@ -0,0 +1,150 @@
...
+ */
+
+
+#import <UIKit/UIKit.h>
+#import "ORKDefines.h"
+
+
+NS_ASSUME_NONNULL_BEGIN
+
+@protocol ORKGraphChartType
```
YuanZhu-apple: Maybe `ORKGraphChartValue` is a better name?

ResearchKit/Charts/ORKGraphChartView.h
```diff
@@ -229,9 +218,66 @@ ORK_AVAILABLE_DECL
...
+ 
+ @param graphChartView      The graph chart view that is asking for the range point.
+ @param pointIndex          An index number identifying the range point in the graph chart view.
+ @param plotIndex           An index number identifying the plot in the graph chart view. This index
+ is 0 in a single-plot graph chart view.
+ 
+ @return The value stack specified by `pointIndex` in the plot specified by `plotIndex` for the
+ specified graph chart view`.
+ */
+- (ORKValueStack *)graphChartView:(ORKGraphChartView *)graphChartView pointForPointIndex:(NSInteger)pointIndex plotIndex:(NSInteger)plotIndex;
```
YuanZhu-apple: Do you think `axisIndex` is better than `pointIndex`?

Something like:
```
graphChartView: valueForAxisIndex: inPlotAtIndex:
```

ResearchKit/Charts/ORKChartTypes.h
rsanchezsaez: Hmm, given that we now have `ORKValueRange` (which can contain a range of two *values*) and `ORKValueStack` (which can contain one or more stacked *values*), I'm thinking `ORKGraphChartValue` may be misleading.

I came up with `ORKValueCollection`, but I'm not completely convinced with this name either.

Any other ideas?

YuanZhu-apple: How about `ORKValueRange ` -> `OCKRangeValue` and `ORKValueStack` -> `ORKStackValue `

ResearchKit/Charts/ORKChartTypes.h
```diff
@@ -0,0 +1,150 @@
...
+ @param value   A `nil` terminated list of `double` values.
+ 
+ @return A value stack object initialized with the passed double values.
+ */
+- (instancetype)initWithStackedValueArray:(NSArray<NSNumber *> *)stackedValues NS_DESIGNATED_INITIALIZER;
+
+/**
+ An array of `NSNumber` objects, each one representing a stacked double value.
+ */
+@property (nonatomic, strong, readonly) NSArray<NSNumber *> *stackedValues;
```
YuanZhu-apple: `copy`

ResearchKit/Charts/ORKChartTypes.h
```diff
@@ -0,0 +1,150 @@
...
+- (instancetype)init;
+
+/**
+ Returns an value stack initialized with the passed double values.
+ 
+ @param value   A `nil` terminated list of `double` values.
+ 
+ @return A value stack object initialized with the passed double values.
+ */
+- (instancetype)initWithStackedValueArray:(NSArray<NSNumber *> *)stackedValues NS_DESIGNATED_INITIALIZER;
```
YuanZhu-apple: How about `initWithStackedValues:` ?

ResearchKit/Charts/ORKChartTypes.h
rsanchezsaez: This sounds good, and in that case `ORKGraphChartValue` makes much more sense.

However, if we do that, how do we differentiate between a *value range*/*stack value* and their own contained `maximumValue`, `minimumValue`, and `stackedValues`? My worry if the the word *value* would be used to refer both the compound classes representing a graph data point, and the individual *values* (double variables) that are part of the compound classes.

ResearchKit/Charts/ORKChartTypes.h
rsanchezsaez: This is related to the discussion above of renaming `ORKValueRange` -> `ORKRangeValue` and `ORKValueStack` -> `ORKStackValue`, and problematic for the same reason I state above.

ResearchKit/Charts/ORKChartTypes.h
YuanZhu-apple: I modified my reply after realizing it could cause confusions. `ORKValueState` is my new idea:)

rsanchezsaez: Sorry to be contrarian, but `ORKValueState` seems confusing for the same reason: the protocol doesn't affect the state of a value, but the state of a value collection. :)

Another idea: `ORKValueContainingType`?

ResearchKit/Charts/ORKGraphChartView.h
rsanchezsaez: I agree that `pointForPointIndex` is suboptimal.

But we would have the same problem as above, it's not clear if value is a *value collection class* or an *individual double value*. Ah! Difficult naming conventions! :stuck_out_tongue_winking_eye:

ResearchKit/Charts/ORKChartTypes.h
YuanZhu-apple: I take back my renaming suggestions.

rsanchezsaez: Sorry about that! If you are convinced of any of your earlier suggestions I'm happy to obligue, but I only wanted to express my concerns. ;-)

Actually, if we don't change the names `ORKValueRange` and `ORKValueStack`, then it actually does make sense to rename this method to `-initWithStackedValues:` (the type is rarely part of the method name).

rsanchezsaez:
I have added a few new commits with general improvements:
- Improve pixel boundaries of `ORKBarGraphChartView` and `ORKLineGraphChartView` so bars and reference lines aren't draw at half-pixel positions.
- Trigger scrubbing by a static long press gesture (in addition to the previously method of a lateral pan gesture).
- Improve *pie chart legend view layout* so the cells are distributed evenly across all the rows, and each row is laid out compactly along the center (I feel this is more aesthetically pleasing).

rsanchezsaez:
I realized we also need to update the *[Charts and Graphs](http://researchkit.org/docs/docs/Charts/ChartsAndGraph.html)* document. By the way, I'd rename it to *Creating Charts* or just *Charts*, since *graphs* are a subtype of *charts* in *ResearchKit* nomenclature.

rsanchezsaez:
Resolved merge conflicts with latest `master`.

ResearchKit/Charts/ORKGraphChartView.h
```diff
@@ -106,20 +109,6 @@ ORK_AVAILABLE_DECL
 - (NSInteger)graphChartView:(ORKGraphChartView *)graphChartView numberOfPointsForPlotIndex:(NSInteger)plotIndex;
```
YuanZhu-apple: Do we want to take this opportunity to rename `numberOfPointsForPlotIndex` to `numberOfAxesForPlotIndex`? Because it makes more sense to three types of graph.

rsanchezsaez: I'm not sure if the proper term is `axis`. One could think that a graph has the *X* and *Y axes* as in `numberOfDivisionsInXAxisForGraphChartView:`.

How about `numberOfDataPointsForPlotIndex:`?

YuanZhu-apple: @rsanchezsaez Yeah, you are right. Sure, let's use `numberOfDataPointsForPlotIndex:`;
And are you going to rename `ORKRangedPoint` -> `ORKValueRangeDataPoint`
and
`ORKValueStack` -> `ORKValueStackDataPoint`

Maybe you can convince me we don't need `DataPoint` in the value names:)

rsanchezsaez: Ok, I'll go for `numberOfDataPointsForPlotIndex:`.

As for the types, they're currently named `ORKValueRange` and `ORKValueStack`. Perhaps we don't want to add the `DataPoint` suffix, as these types are internally chart-agnostic (they could be potentially reused outside charts). `DataPoint` refers more to its behavior inside a chart.

What I could do is rename the base `ORKGraphChartType` protocol to `ORKValueCollectionType`.

What do you reckon?

YuanZhu-apple: Agree, let's rename `ORKGraphChartType` to `ORKValueCollectionType`

rsanchezsaez:
Done!

YuanZhu-apple:
@rsanchezsaez Thanks! r=me.

Issue #308, opened 2015-07-02
ORKHealthKitCharacteristicTypeAnswerFormat does not have a member named 'minimumDate'


brendenwest:
I'm using the code shown in RK documentation to collect a date-of-birth:

        var dobFormat = ORKHealthKitCharacteristicTypeAnswerFormat(characteristicType: HKCharacteristicType.characteristicTypeForIdentifier(HKCharacteristicTypeIdentifierDateOfBirth))

It works fine for rendering a date picker, but doesn't support the minimumDate property available for ORKDateAnswerFormat.

YuanZhu-apple:
@brenbob `ORKHealthKitCharacteristicTypeAnswerFormat ` set a default minimumDate 150 years to current date. 
Any reason you would like to change this default setting?

brenbob:
I need to ensure respondents have a minimum age. Perhaps 'maximumDate' is the appropriate setting, but that's not supported either.

YuanZhu-apple:
That's a valid point, we can consider expose `minimumDate` and `maximumDate` in the API of `ORKHealthKitCharacteristicTypeAnswerFormat` as optional attributes for date of birth.

brucehappy:
I may not properly understand your purpose, but shouldn't determining whether the subject is at least of a certain age something that should be done in the onboarding process?  For instance, the AsthmaHealth app specifically asks if the subject is over 18 in question # 1:
https://github.com/ResearchKit/AsthmaHealth/blob/master/Asthma/Onboarding/APHInclusionCriteriaViewController.m

I am not suggesting the ability to restrict the minimum and maximum date shouldn't be exposed, however.  I'm just trying to understand if perhaps there is another way to accomplish what you want to do.

rsanchezsaez:
I have implemented this feature in the aforementioned PR.

rsanchezsaez:
This has been merged and the issue can be closed.

PR #639, opened 2016-03-21
[ORKHealthKitCharacteristicTypeAnswerFormat] Expose date picker attributes

rsanchezsaez:
This PR deals with [Issue #308](https://github.com/ResearchKit/ResearchKit/issues/308).

It adds the `defaultDate`, `minimumDate`, `maximumDate` and `calendar` properties to the `ORKHealthKitCharacteristicTypeAnswerFormat` which can be used to customize the presented date picker (only applicable to the 'HKCharacteristicTypeIdentifierDateOfBirth' characteristic type).

It also updates `ORKTest` *serialization tests* accordingly so they continue to pass.


ResearchKit/Common/ORKHealthAnswerFormat.m
```diff
@@ -154,26 +164,52 @@ - (ORKAnswerFormat *)impliedAnswerFormat {
...
+                                                                              calendar:_calendar];
             impliedAnswerFormat = format;
         }
     }
     return impliedAnswerFormat;
 }
 
+- (void)setDefaultDate:(NSDate *)defaultDate {
+    if (!defaultDate) {
+        defaultDate = [_calendar dateByAddingUnit:NSCalendarUnitYear value:-35 toDate:[NSDate date] options:0];
```
YuanZhu-apple: Do think default value should be generated during runtime?
In case, someone created one instance and archived, 10 years later, the system load this from archive:)

YuanZhu-apple: Same question for `minimumDate`/`maximumDate`/`calendar`

YuanZhu-apple: Maybe this answer format  should accept time intervals instead of a absolute time? So all the elements should be time intervals to survey present time?

rsanchezsaez: Regarding archiving and unarchiving you're probably right. I'll change the implementation so these properties are `nullable` instead of `null_resettable`.

---

Regarding intervals, what about if you want to target a date which is not around present time (e.g., people between 50 and 60)? Then just an interval is not enough. I modelled the arguments mirroring `ORKDateAnswerFormat`.

---

Regarding default value, you are right, I can make it so the default values are calculated on reading instead on writting.

Testing/ORKTest/ORKTestTests/ORKJSONSerializationTests.m
```diff
@@ -487,7 +485,7 @@ - (void)testORKSerialization {
         id instance2 = [ORKESerializer objectFromJSONObject:mockDictionary error:NULL];
        
         NSArray *unTouchedKeys = [mockDictionary unTouchedKeys];
-        
+
```
YuanZhu-apple: Remove extra spaces.

YuanZhu-apple:
@rsanchezsaez Thanks! R=me, can you resolve the conflict?

rsanchezsaez:
Conflict resolved!

YuanZhu-apple:
Thanks!

Issue #484, opened 2015-09-30
Add an answer format for Height input 


YuanZhu-apple:
2-column picker view for height (if current locale uses ft and inches). 

![14e7a6a4-6639-11e5-8a77-75f051a0bdaa](https://cloud.githubusercontent.com/assets/11466704/10186447/e6241000-6705-11e5-9f1e-57612c9eb1ed.jpeg)

Health App has a 2-column picker too.

PedroMaganinho:
@YuanZhu-apple I am a medical student and for my thesis I am trying to make a urological app using ResearchKit. How can I use this Height answer format? Can I put this in meters?  
I have another question. I am using ResearchKit and know I have all the questions I want. These issues do I get a result. Currently I'm trying to present this outcome in a dashboard as the applications that already exist. Unfortunately when I import APPCORE  for my project it's not working. Can you help me?

YuanZhu-apple:
@PedroMaganinho This is feature request we received, not yet implemented.

For the dashboard, you can check out the chart library we just introduce in ResearchKit, which very easy to use. Open `ORKCatalog` app, you will find some demo.

PedroMaganinho:
@YuanZhu-apple Thank you!!! I will try! If I have any doubt can I ask you?

YuanZhu-apple:
Sure, just create a issue. Many other developers can help you too.

EdAmorim:
@YuanZhu-apple @PedroMaganinho I'm having some trouble with the same subject. I have created an issue (#499). Can anyone help me with that?

rsanchezsaez:
I'm almost done implementing this, I'll probably finish it tomorrow.

rsanchezsaez:
Done.

rsanchezsaez:
This has been merged and the issue can be closed.

ninoguba:
@rsanchezsaez thanks for adding this! Only wondering why there isn't one for weight yet.   I can work on implementing ORKWeightAnswerFormat if no one else has started on it.  I will just use your work as a base and adapt it for weight.

ninoguba:
I'm almost done with my implementation of ORKWeightAnswerFormat...

PR #643, opened 2016-03-24
[ORKAnswerFormat] Implement picker-based 'ORKHeightAnswerFormat'

rsanchezsaez:
This PR implements the picker-based 'ORKHeightAnswerFormat' [Issue #484](https://github.com/ResearchKit/ResearchKit/issues/484).

Details:
- Supports the metric and ~~imperial~~USC measuring systems
- Supports standalone question steps and on form steps
- Serialization unit tests have been implemented

Notes:
- For input it uses the newly added 'ORKHeighPicker' class
- Height limits and default value are hardcoded into the picker, taking the same values as the *Health app*: [`0 ft, 0 in` - `9 ft, 11 in`] range, default: `5 ft, 4 in`; [`0 cm` - `299 cm`] range, default: `162 cm`
- Results are output as an `ORKNumericResult` object, with the unit canonically set to ~~`in` or~~ `cm` ~~depending on the chosen measuring system (on the imperial system, feet and inches always get converted to total inches in the output result)~~
- Visually, the picker and the chosen answer in form steps display the `cm` and decomposed `ft in` values, with their units localized if applicable (unlike the canonical result units which are never localized)

---
simulator screen shot 24 mar 2016 14 58 08
---
simulator screen shot 24 mar 2016 15 02 51
---
simulator screen shot 24 mar 2016 15 01 11


ResearchKit/Common/ORKAnswerFormat.h
```diff
@@ -1330,6 +1339,46 @@ ORK_CLASS_AVAILABLE
...
+ @return An initialized height answer format.
+ */
+- (instancetype)initWithMetricSystem:(BOOL)useMetricSystem NS_DESIGNATED_INITIALIZER;
+
+/**
+ Indicates the measuring system used by the answer format.
+ 
+ A value of `NO` indicates that the imperial system is used.
+ */
+@property (readonly) BOOL useMetricSystem;
```
YuanZhu-apple: Maybe we can make this configuration nullable, when it is not set, then in UI it respect system locale and convert to metric system in result.
Whether to use metric system can be determine from NSLocale. 
```
NSLocale *locale = [NSLocale currentLocale]; 
BOOL isMetric = [[locale objectForKey:NSLocaleUsesMetricSystem] boolValue];
```

The concept is like calendar attribute on  `ORKDateAnswerFormat`, developer can specify or leave it to framework to decide.
http://researchkit.org/docs/Classes/ORKDateAnswerFormat.html#//api/name/calendar

How do you think?:)

YuanZhu-apple: Maybe we can make this configuration nullable, when it is not set, then in UI it respect system locale and convert to metric system in result.
Whether to use metric system can be determine from NSLocale. 
```
NSLocale *locale = [NSLocale currentLocale]; 
BOOL isMetric = [[locale objectForKey:NSLocaleUsesMetricSystem] boolValue];
```

The concept is like calendar attribute on  `ORKDateAnswerFormat`, developer can specify or leave it to framework to decide.
http://researchkit.org/docs/Classes/ORKDateAnswerFormat.html#//api/name/calendar

How do you think?:)

YuanZhu-apple:
Also is it possible to make this become `ORKHealthKitQuantityTypeAnswerFormat` implied answerFormat if `type` is `HKQuantityTypeIdentifierHeight`?

YuanZhu-apple:
Thanks for the great work!

ResearchKit/Common/ORKAnswerFormat.h
rsanchezsaez: That's what I intended to do by adding the convenience [`init` method](https://github.com/rsanchezsaez/ResearchKit/blob/rsanchezsaez-ORKHeightAnswerFormat/ResearchKit/Common/ORKAnswerFormat.m#L2348-L2351) and making the property read-only. My idea is that it only gets calculated one when you instantiate the *answer format*. If we wanted for `useMetricSystem` to be `nullable`, then it would need to be an `NSNumber` instead of a `BOOL`, and that seemed a bit overboard for just a flag.

Or do you have something else in mind? :-)

rsanchezsaez: If you want that all results are reported in metric regardless of the `useMetricSystem` value, I can certainly do that.

rsanchezsaez:
You're welcome! I have added the `ORKHealthKitQuantityTypeAnswerFormat` implied `answerFormat` customization.

ResearchKit/Common/ORKAnswerFormat.h
YuanZhu-apple: Maybe create a enum [system/metric/american]?
Yes, nice to report value in metric only.

rsanchezsaez: Done!

Let me know if you don't like the nomenclature I chose for the enum values ([USC](https://en.wikipedia.org/wiki/United_States_customary_units)).

ResearchKit/Common/ORKAnswerFormat.m
```diff
@@ -1303,7 +1305,7 @@ - (NSString *)localizedInvalidValueStringWithAnswerString:(NSString *)text {
         return nil;
     }
     NSString *string = nil;
-    NSNumberFormatter *formatter = [self makeNumberFormatter];
+    NSNumberFormatter *formatter = ORKNumberFormatter();
```
YuanZhu-apple: Maybe `OCKDecimalNumberFormatter` a more accurate name?

ResearchKit/Common/ORKAnswerFormat.m
```diff
@@ -2320,6 +2322,102 @@ - (NSString *)stringForAnswer:(id)answer {
...
+}
+
+- (NSUInteger)hash {
+    return super.hash ^ _measurementSystem;
+}
+
+- (instancetype)initWithCoder:(NSCoder *)aDecoder {
+    self = [super initWithCoder:aDecoder];
+    if (self) {
+        ORK_DECODE_INTEGER(aDecoder, measurementSystem);
```
YuanZhu-apple: Can we use `ORK_DECODE_ENUM` and `ORK_ENCODE_ENUM`, even though they are quite the same.

ResearchKit/Common/ORKAnswerFormat.m
```diff
@@ -2320,6 +2322,102 @@ - (NSString *)stringForAnswer:(id)answer {
...
+    || (_measurementSystem == ORKMeasurementSystemLocal && ((NSNumber *)[[NSLocale currentLocale] objectForKey:NSLocaleUsesMetricSystem]).boolValue);
+}
+
+- (NSString *)stringForAnswer:(id)answer {
+    NSString *answerString = nil;
+    
+    if (!ORKIsAnswerEmpty(answer)) {
+        NSNumberFormatter *formatter = ORKNumberFormatter();
+        if (self.useMetricSystem) {
+            answerString = [NSString stringWithFormat:@"%@ %@", [formatter stringFromNumber:answer], ORKLocalizedString(@"MEASURING_UNIT_CM", nil)];
```
YuanZhu-apple: Can we use `NSLengthFormatter` to format the string? It will save tons of effort.
https://developer.apple.com/library/ios/documentation/Miscellaneous/Reference/NSLengthFormatter_Class/index.html#//apple_ref/occ/instm/NSLengthFormatter/stringFromMeters:

http://stackoverflow.com/questions/26073355/nslengthformatter-not-displaying-correct-value

ResearchKit/Common/ORKHeightPicker.m
```diff
@@ -0,0 +1,272 @@
...
+    }
+    return title;
+}
+
+- (void)pickerView:(UIPickerView *)pickerView didSelectRow:(NSInteger)row inComponent:(NSInteger)component {
+    [self valueDidChange:self];
+}
+
+- (NSArray *)centimeterValues {
+    static NSArray *inchesValues = nil;
```
YuanZhu-apple: centimeterValues

ResearchKit/Common/ORKHeightPicker.m
```diff
@@ -0,0 +1,272 @@
...
+    if (_answerFormat.useMetricSystem) {
+        NSUInteger index = [[self centimeterValues] indexOfObject:answer];
+        if (index == NSNotFound) {
+            [self setAnswer:[self defaultAnswerValue]];
+            return;
+        }
+        [_pickerView selectRow:index inComponent:0 animated:NO];
+    } else {
+        double feet, inches;
+        ORKInchesToFeetAndInches(((NSNumber *)answer).doubleValue, &feet, &inches);
```
YuanZhu-apple: Isn't the answer value in ORKNumericQuestionResult is always in centimeters? 
And define the answer value input/output for this picker always use centimeter value?

ResearchKit/Common/ORKHeightPicker.m
```diff
@@ -0,0 +1,272 @@
...
+    if (self) {
+        NSAssert([answerFormat isKindOfClass:[ORKHeightAnswerFormat class]], @"answerFormat should be ORKHeightAnswerFormat");
+        
+        // Take into account locale changes so unit string can be updated
+        [[NSNotificationCenter defaultCenter] addObserver:self
+                                                 selector:@selector(currentLocaleDidChange:)
+                                                     name:NSCurrentLocaleDidChangeNotification object:nil];
+
+        self.answer = answer;
+        _answerFormat = answerFormat;
```
YuanZhu-apple: Maybe pass in and cache `useMetricSystem`, then no need to pass in the answer format.

ResearchKit/Common/ORKHeightPicker.m
```diff
@@ -0,0 +1,272 @@
...
+    NSNumber *answer = nil;
+    if (_answerFormat.useMetricSystem) {
+        NSInteger row = [_pickerView selectedRowInComponent:0];
+        answer = [self centimeterValues][row];
+    } else {
+        NSInteger feetRow = [_pickerView selectedRowInComponent:0];
+        NSInteger inchesRow = [_pickerView selectedRowInComponent:1];
+        NSNumber *feet = [self feetValues][feetRow];
+        NSNumber *inches = [self inchesValues][inchesRow];
+        answer = @( ORKFeetAndInchesToInches(feet.doubleValue, inches.doubleValue) );
```
YuanZhu-apple: Save to centimeter value and regardless the unit system?

rsanchezsaez:
I'll look into this next week. :)

ResearchKit/Common/ORKHeightPicker.m
rsanchezsaez: The thing is that the answer value is used to look for the selected row in the picker (`selectedRowInComponent:`), so I'm afraid that converting back and forth to centimeters might introduce floating point rounding errors.

Because of this, I figured out that it's better to use the locale format in the picker (which when are in inches it should not cause rounding errors), and just convert to *cm* when then answer result is calculated.

ResearchKit/Common/ORKHeightPicker.m
rsanchezsaez: I'm passing the `answerFormat` because this implementation takes into account the case in which the user changes the locale while she is looking at the question midway through running the app ([this line](https://github.com/rsanchezsaez/ResearchKit/blob/rsanchezsaez-ORKHeightAnswerFormat/ResearchKit/Common/ORKAnswerFormat.m#L2394)), so it didn't want to implement that logic twice.

ResearchKit/Common/ORKHeightPicker.m
rsanchezsaez: Same reason as below.

The answer value is used to look for the selected row in the picker (selectedRowInComponent:), so I'm afraid that converting back and forth to centimeters might introduce floating point rounding errors.

Because of this, I figured out that it's better to use the locale format in the picker (which when are in inches it should not cause rounding errors), and just convert to cm when then answer result is calculated.

ResearchKit/Common/ORKAnswerFormat.m
rsanchezsaez: The length formatter doesn't seem to support output in feet and inches, which is what's nicest for a height value in USC.

Additionally, the formatter takes the user locale by default, but this implementation allows for the researcher to hardcode the desired unit system disregarding the user locale (`ORKMeasurementSystemMetric` or `ORKMeasurementSystemUSC`, as opposed to `ORKMeasurementSystemLocal`), and the formatter wouldn't support this.

rsanchezsaez:
I've corrected some of the noted issues, and replied to some other comments that I think should remain as they are currently implemented.

Please let me know if you disagree and I'll be happy to modify the implementation.

ResearchKit/Common/ORKAnswerFormat.h
```diff
@@ -1330,6 +1339,44 @@ ORK_CLASS_AVAILABLE
 
 
 /**
+ The `ORKHeightAnswerFormat` class represents the answer format for questions that require users
+ to enter a height.
+ 
+ A date answer format produces an `ORKNumericQuestionResult` object. The result is always reported
```
YuanZhu-apple: A `height` answer format:)

ResearchKit/Common/ORKHeightPicker.m
YuanZhu-apple: Please correct me if i am wrong, `cm = inches * 2.54` seems OK for me to convert back and forth.


IMO, `floating point rounding error` happens if a question has been presented in two metric systems. But most likely a question will be only presented in one system.

YuanZhu-apple:
@rsanchezsaez I tested and found the result value are restored if I hit `next` and come back (or hit `back` and re-enter the step).

rsanchezsaez:
I've fixed the issue about the answer not being correctly preserved, the documentation typo, and I have implemented your suggestion of using *cms* as the internal answer unit in `ORKHeightPicker`.

Latest master merged as well.

YuanZhu-apple:
@rsanchezsaez Thanks for the update.
In the `mini form`, the answer restoration is still incorrect to me.

rsanchezsaez:
@YuanZhu-apple Oops! Fixed now. :)

YuanZhu-apple:
@rsanchezsaez Thanks!
Approved!

Issue #642, opened 2016-03-24
Modify ResearchKit.podspec to include .swift files


jennadavis077:
I am coding in swift, and using Cocoa Pods to manage my dependencies. I am unable to use the argument resultPredicatesAndDestinationStepIdentifiers because it exists in  ORKStepNavigationRule.swift (located in the common directory), and the ResearchKit.podspec only imports .m and .h files, not .swift files. This podspec should be modified to support .swift files.

rsanchezsaez:
Thanks for the report, I've submitted a pull request fixing this.

PR #644, opened 2016-03-24
[Cocoapods] Fix podspec to include swift files

rsanchezsaez:
Fixes [Issue #642](https://github.com/ResearchKit/ResearchKit/issues/642).

### DIFF
--- a/ResearchKit.podspec
+++ b/ResearchKit.podspec
@@ -8,7 +8,7 @@ Pod::Spec.new do |s|
   s.author       = { 'researchkit.org' => 'http://researchkit.org' }
   s.source       = { :git => 'https://github.com/ResearchKit/ResearchKit.git', :tag => s.version.to_s }
   s.public_header_files = `./scripts/find_headers.rb --public --private`.split("\n")
-  s.source_files = 'ResearchKit/**/*.{h,m}'
+  s.source_files = 'ResearchKit/**/*.{h,m,swift}'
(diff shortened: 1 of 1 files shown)


vtourraine:
Thanks for addressing this issue, but I’m afraid this won’t really solve the problem.

Once there’s a Swift file inside a Pod, projects trying to import it need to use the `use_frameworks!` Podfile option. Problem is: ResearchKit fails to build when imported by CocoaPods as a framework.

screen shot 2016-03-25 at 2 38 58 pm

So right now, including the Swift files into the ResearchKit podspec just completely breaks the CocoaPods installation method, for both Objective-C and Swift projects. Not including them, as it is today, keep it partially CocoaPods compatible, but effectively excludes some ResearchKit classes from Swift projects, as reported in #642.

And unfortunately, I don’t have a solution for this. :sos: Xcode/CocoaPods expert?

rsanchezsaez:
Hey @vtourraine, thanks for spotting this.

I just finished an `#import` section refactor over at [PR](https://github.com/ResearchKit/ResearchKit/pull/648). I'm not completely sure, but I suspect this may fix the issue that you show on the screenshot. What is a good way to test this with a `podspec` file which is not on the main repo?

vtourraine:
I’ve tested configuring a project using CocoaPods with your branch, and yes, compilation is now fixed. Great work! 

Looking forward to #648 being merged in.

YuanZhu-apple:
Thanks @rsanchezsaez and @vtourraine !

rsanchezsaez:
Careful, not sure if this truly fixes the problem without merging [#648](https://github.com/ResearchKit/ResearchKit/pull/648) (which is a pretty big refactor) as well!

PR #637, opened 2016-03-15
[Misc] Code homogenization

rsanchezsaez:
This PR does a minor `ORKHelpers.h`/`ORKDefines.h`/`ORKDefines_Private.h` reorganization, and adds some minor and fairly harmless coding style conforming and homogenization (whitespace around unary and ternary operators; usage of some predefined macros; avoid using the `__` prefix, [Issue #510](https://github.com/ResearchKit/ResearchKit/issues/510)).

Things of note:
- `ORKDefines.h` was split into `ORKDefines.h` and `ORKTypes.h`, which favours encapsulation and reduces inter-dependencies.
- `ORKDefines_Private.h` (containing some functions needed in `ORKTest` unit tests) was renamed to `ORKHelpers_Private.h`, since those are helper functions rather than macros. At the same time, the localization function and macro declarations were moved to `ORKHelpers.h` (which has internal visibility), since the localization functions are not needed outside *ResearchKit*.

See the commit messages for details on the the rest of minor changes.

PR #648, opened 2016-03-28
[Misc] #import section refactor to reduce header interdependency

rsanchezsaez:
**Note:** This PR should only be merged after [PR #637](https://github.com/ResearchKit/ResearchKit/pull/637), as it was created from the tip of that branch.

Also, it's probably a good idea to merge all other ongoing PR first, as it will introduce minor conflicts with any PR that adds or modifies `#import` statements. I can happily resolve conflicts within this PR if they appear.

---

This PR performs an `#import` section refactor to reduce header interdependency.

Compilation time after you modify a header file has been greatly improved by avoiding including the umbrella headers (`ResearchKit.h` and `ResearchKit_Private.h`) in any of the RK framework files.

Also included:
- Restructure `.m` files `#import` section into a hierarchy of alphabetically sorted subsections separated by one newline. The `#import` order follows advice found [here](http://qualitycoding.org/import-order/). Import order as follows:

> 1. own class header (or headers if we include the `Private` or `Internal` ones)
> 2. local views and controls
> 3. view controllers and recorders
> 4. RK model files (`ORKAnswerFormat`, `ORKResult`, `ORKStep`, `ORKTask`, etc.)
> 5. general files and helpers (`ORKAccessibility`, `ORKHelpers`, `ORKSkin`, class extensions, etc.)
> 6. iOS frameworks
> 

- Remove needless and duplicate imports from `.m` files
- Make sure all headers include the minimum required headers by using `@class` and `@protocol` forward declarations. Generally, an RK header will only include (where appropriate): `Foundation` or `UIKit` (as appropriate); `ORKDefines.h`; `ORKTypes.h`;  the header  containing superclass; and the header containing any adopted protocol. There are some minor exceptions such as the `ORKAccessibility.h` helper header, or `ORKStepHeaderView_Internal.h`
- An exception to the previous point are `Private` or `Internal` headers which generally only include their own class header at the next public level
- Rename `ORKHelpers.h` to `ORKHelpers_Internal.h` (it is not public, and its name was misleading given that we also have `ORKHelpers_Private.h`)
- Add a few missing `NS_ASSUME_NONNULL_BEGIN`/`NS_ASSUME_NONNULL_END` annotation in headers
- Add a few missing double newline before and after import section


rsanchezsaez:
I have added a new commit using modules for framework imports, and moving headers around a little bit, that I believe allows adding *.swift* files to the Cocoapod spec file, and thus allows for the fixes for [Issue #642](https://github.com/ResearchKit/ResearchKit/issues/642).

---

I think there's a bug in the Xcode build chain because, even though all *unit tests* build and pass fine from a clean state, there remains an error that says *Unable to build module ResearchKit* when running *unit tests*. But, on the other hand, when running the `ORKTest` app, which also uses the *ResearchKit module* doesn't show this error.

rsanchezsaez:
I've resolved merged conflicts and this is now ready to merge.

@vtourraine Could you re-check that after the merge this branch still works through *Cocoapods*, and that it correctly imports the *swift* files?

YuanZhu-apple:
@rsanchezsaez Thanks!
R=me.

PR #653, opened 2016-04-05
Update ORKSignatureView to support pressure sensitive stroke widths.

samfalconer:
Add property for setting the stroke width variation (default to 3pts).

If the device supports 3D touch, use the true pressure to determine
how thick the line should be at a given point.

If the device does not support 3D touch, use the speed of the stroke
to determine how thick the line should be at a given point.

Here are some example screenshots using 3D touch:
force 1
force 2
force 3

... and here are some examples **not** using 3D touch:

velocity 1
velocity 2

### DIFF
--- a/ResearchKit/Consent/ORKSignatureView.m
+++ b/ResearchKit/Consent/ORKSignatureView.m
@@ -266,14 +266,23 @@ static const CGFloat LineWidthStepValue = 0.25f;
     dispatch_once(&onceToken, ^{
         
         isAvailable = NO;
-        if ([self.traitCollection respondsToSelector:@selector(forceTouchCapability)]) {
-            if (self.traitCollection.forceTouchCapability == UIForceTouchCapabilityAvailable) {
-                isAvailable = YES;
-            }
+        if ([self.traitCollection respondsToSelector:@selector(forceTouchCapability)] && 
+             self.traitCollection.forceTouchCapability == UIForceTouchCapabilityAvailable) {
+            isAvailable = YES;
         }
     });
     
     return isAvailable;
+}
+
+- (BOOL)_isTouchTypeStylus:(UITouch*)touch {
+    BOOL isStylus = NO;
+    
+    if ([touch respondsToSelector:@selector(type)] && touch.type == UITouchTypeStylus) {
+        isStylus = YES;
+    }
+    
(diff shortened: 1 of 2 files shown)


rsanchezsaez:
This looks like a very nice feature! Thanks for the contribution.

ResearchKit/Consent/ORKSignatureView.m
```diff
@@ -254,6 +259,23 @@ - (NSArray *)backgroundLines {
...
 
+- (BOOL)_isForceTouchAvailable {
+    static BOOL isAvailable;
+    static dispatch_once_t onceToken;
+    
+    dispatch_once(&onceToken, ^{
+        
+        isAvailable = NO;
+        if ([self.traitCollection respondsToSelector:@selector(forceTouchCapability)]) {
+            if (self.traitCollection.forceTouchCapability == UIForceTouchCapabilityAvailable) {
```
rsanchezsaez: You can fold both checks onto the same `if` line as:

    if ([self.traitCollection respondsToSelector:@selector(forceTouchCapability)] && self.traitCollection.forceTouchCapability == UIForceTouchCapabilityAvailable)`.

The second test won't be run unless the first one passes.

samfalconer:
Sure thing! I've merged the if statements, and added a check for Apple Pencil support.

samfalconer:
I don't own an Apple Pencil, so if someone could check that it is correctly detecting pressure and not falling back to using speed, that would be great.

umerkhan-apple:
@samfalconer Thanks for this upgrade to the feature. I will test it with an Apple Pencil and let you know of the results.

umerkhan-apple:
@samfalconer Tested on iPad Pro (9.7") with Apple Pencil. Works like a charm! 👍

samfalconer:
@umerkhan-apple Great to hear! Was a fun feature to add.

umerkhan-apple:
@samfalconer The PR looks good to me. Ill wait till EOD to see if @YuanZhu-apple has any updates before I merge it.

samfalconer:
It's been a pleasure working with you all!

umerkhan-apple:
@samfalconer Likewise! Thanks for the contribution! 👍

Issue #649, opened 2016-03-29
ORKESerialization runs on simulator but fails to archive


jefferythewind:
I included the ORKESerialization.m and .h files in my iOS project.  I did this but dragging the 2 files from the files downloaded from GitHub, and then doing the #import "ORKESerialization.h" in the bridging header. 

I was able to use the packages successfully, building no problem and running my app in the simulator for iOS versions 9.1 and 9.3.  However when I tried to archive the project for a generic iOS device, I get the following error:

```
Undefined symbols for architecture arm64:
"_OBJC_CLASS_$_ORKLocation", referenced from: objc-class-ref in ORKESerialization.o "_OBJC_CLASS_$_ORKConfirmTextAnswerFormat", referenced from: objc-class-ref in ORKESerialization.o ld: symbol(s) not found for architecture arm64 clang: error: linker command failed with exit code 1 (use -v to see invocation)

```

I asked a question on stackoverflow.com here: http://stackoverflow.com/questions/36266388/ios-build-succeeds-but-fails-to-archive

It was suggested that this may be a bug.

YuanZhu-apple:
I confirm the archive is failed:
1. Open ORKTest project
2. Check both target memberships for `ORKESerialization.m`
<img width="131" alt="screen shot 2016-03-29 at 11 27 54 am" src="https://cloud.githubusercontent.com/assets/11466704/14119092/61868cbc-f5a1-11e5-9fdb-ffd053e47982.png">
3. Run Archive

Error output:

```
Undefined symbols for architecture arm64:
  "_OBJC_CLASS_$_ORKLocation", referenced from:
      objc-class-ref in ORKESerialization.o
  "_OBJC_CLASS_$_ORKConfirmTextAnswerFormat", referenced from:
      objc-class-ref in ORKESerialization.o
```

jinseok7586:
I am having this issue, too

YuanZhu-apple:
ORKESerialization used `MKCoordinateValue` which is not a valid iOS symbol.

celteeka:
Was this fixed?

rsanchezsaez:
@YuanZhu-apple: I think this issue has been fixed and can be closed.

jefferythewind:
Sorry I haven't had a chance yet to try out the new build, will confirm asap.

umerkhan-apple:
@jefferythewind Could you confirm if the build has been fixed?

jefferythewind:
[...]

PR #663, opened 2016-04-20
Fix class visibility

YuanZhu-apple:
Mark class available for two Classes.
This is the fix for https://github.com/ResearchKit/ResearchKit/issues/649


YuanZhu-apple:
r=self.

Issue #705, opened 2016-06-07
ORKSample: Blank white screen after dismissing Notification Center


malhal:
Blank white screen after dismissing Notification Center or Control Center. Possibily similar to #694

Run the app, use Touch ID, on Activities tab, bring down NC, slide up NC, blank white screen.

ninoguba:
I think it is exactly the same root cause as #694.

malhal:
Yep this bug looks like it is caused by the application delegate method used for resuming, applicationWillEnterForeground won't get called when the application activates from NC.

Issue #694, opened 2016-05-28
Prompting for Permission in a task leads to a blank white screen when the task completes


ninoguba:
I have a step that set Local Notifications and when that step appears, I prompt the user for permission, regardless if the user grants or denies it, it continues with the rest of the steps of the task.  The issue is after the task completes, it leads to a blank white screen.  How can this be avoided?

ninoguba:
Ok I used the LockViewController provided in the ORKSample app and that hides the content when the passcode is already set.   This is the one causing the white screen whenever you ask for any permission.   To workaround this issue I just disabled the hiding of the content view.

PR #707, opened 2016-06-08
Fix for blank white screen when app activates back from viewing the N…

malhal:
…C or a privacy prompt

Fix for issue #705 and #694. It was using the wrong application delegate method for hiding the view from app switcher.


YuanZhu-apple:
R=me. Thanks!

PR #717, opened 2016-06-21
Modifications to ORKTappingContentView to support resizing the view if skippable.

syoung-smallwisdom:
In order to implement a two-hand tapping task (pull request TBD), the tapping view needs to be resizable to accommodate showing the skip button. The skip button is required so that the participant in a research study can skip one of their hands if needed.

### DIFF
--- a/ResearchKit/ActiveTasks/ORKTappingContentView.h
+++ b/ResearchKit/ActiveTasks/ORKTappingContentView.h
@@ -40,6 +40,8 @@ NS_ASSUME_NONNULL_BEGIN
 - (void)setTapCount:(NSUInteger)tapCount;
 - (void)setProgress:(CGFloat)progress animated:(BOOL)animated;
 
+@property (nonatomic, assign) BOOL hasSkipButton;
+
 @property (nonatomic, strong, readonly) ORKRoundTappingButton *tapButton1;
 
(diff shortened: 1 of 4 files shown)


YuanZhu-apple:
@syoung-smallwisdom This looks good, can you confirm that the tapping step by default is not optional? So the skip button won't show up if nobody tweak the `optional` attribute on the step.

syoung-smallwisdom:
@YuanZhu-apple 
Whoops, sorry about that. https://github.com/ResearchKit/ResearchKit/pull/723 has a unit test that checks this (should be optional == YES for two-hand case) but this got pulled into it's own PR. Fixed now.

YuanZhu-apple:
@syoung-smallwisdom Can you write a test case in `ORKTest` app to verify skipping a tapping test can discard the result file properly?

syoung-smallwisdom:
@YuanZhu-apple 

Sorry, but no, I can't. `ORKTappingIntervalStep`and `ORKTappingIntervalStepViewController` are both internal classes so there is no means of instantiating them directly from the unit test project. I can tell you that the result of running the task manually will include an `ORKTappingIntervalResult` with `nil` samples and no accelerometer data. Our servers are setup to handle this case and interpret an empty sample set as "skip".

syoung-smallwisdom:
Closing. Required for https://github.com/ResearchKit/ResearchKit/pull/743/files

PR #718, opened 2016-06-21
Fix bug where the defaultResultSource is using ivar rather than public property accessor

syoung-smallwisdom:
By using the ivar rather than the property method accessor, this will fail for implementations where the developer is overriding `-defaultResultSource` rather than setting it. Since override is an accepted pattern in both objc and swift, direct access of the ivar can result in unexpected behavior.


YuanZhu-apple:
OK with me.

PR #719, opened 2016-06-21
Allow steps to instantiate the view controller

syoung-smallwisdom:
Allow steps that are not a part of ResearchKit to have access to a method for instantiating the view controller since the `-stepViewControllerClass` is internal (and not designed for an implementation that uses swift).


YuanZhu-apple:
@syoung-smallwisdom I found this http://stackoverflow.com/questions/31119521/return-new-viewcontroller-in-swift

So you can override `stepViewControllerClass`.

Also `instantiateStepViewController` has the side effect of not accepting other inputs, like `result` and so on.

syoung-smallwisdom:
@YuanZhu-apple Then please expose stepViewControllerClass publicly.

PR #720, opened 2016-06-21
Fix a bug in task navigation that assumes first step is instruction

syoung-smallwisdom:
Sometimes the first step in a custom `ORKTask` is *not* an instruction step. The result was that this method would update the text on an internal instruction step to "Get Started" which is not accurate.

### DIFF
--- a/ResearchKit/Common/ORKTaskViewController.m
+++ b/ResearchKit/Common/ORKTaskViewController.m
@@ -873,7 +873,10 @@ static NSString *const _ChildNavigationControllerRestorationKey = @"childNavigat
     }
     
     ORKStep *step = viewController.step;
-    [self updateLastBeginningInstructionStepIdentifierForStep:step goForward:goForward];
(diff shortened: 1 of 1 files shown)


YuanZhu-apple:
Maybe the name of `_lastBeginningInstructionStepIdentifier` is not accurate, but  `isStepLastBeginningInstructionStep:` did check `  [step isKindOfClass:[ORKInstructionStep class]]`.

```
- (BOOL)isStepLastBeginningInstructionStep:(ORKStep *)step {
    if (!step) {
        return NO;
    }
    return (_lastBeginningInstructionStepIdentifier != nil &&
            [step isKindOfClass:[ORKInstructionStep class]]&&
            [step.identifier isEqualToString:_lastBeginningInstructionStepIdentifier]);
}
```

syoung-smallwisdom:
Hmmm... Not sure why we were getting issue then, but I will close. Thanks.

PR #724, opened 2016-06-21
Audio ActiveTask View - add ability to define threshold value

syoung-smallwisdom:
Allow for a custom threshold on `ORKAudioStepViewController`.

### DIFF
--- a/ResearchKit/ActiveTasks/ORKAudioStepViewController.h
+++ b/ResearchKit/ActiveTasks/ORKAudioStepViewController.h
@@ -37,6 +37,8 @@ NS_ASSUME_NONNULL_BEGIN
 ORK_CLASS_AVAILABLE
 @interface ORKAudioStepViewController : ORKActiveStepViewController
 
+@property (nonatomic, assign) CGFloat alertThreshold;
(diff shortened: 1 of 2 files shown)


ResearchKit/ActiveTasks/ORKAudioStepViewController.h
```diff
@@ -37,6 +37,8 @@ NS_ASSUME_NONNULL_BEGIN
 ORK_CLASS_AVAILABLE
 @interface ORKAudioStepViewController : ORKActiveStepViewController
 
+@property (nonatomic, assign) CGFloat alertThreshold;
```
YuanZhu-apple: Is a negative value a valid value? Should we use the negative value to restore the default value?

syoung-smallwisdom: Updated. This was originally written by Thread in advance of promoting mPower during the March Keynote. I am not certain why they set it initially to -1, but doing so is unnecessary. Thanks for catching.

YuanZhu-apple:
r=me.

PR #725, opened 2016-06-21
Add method to ORKOrderedTask to allow steps to be removed.

syoung-smallwisdom:
Allow developers to mutate the steps returned by an included module to remove a step that is unwanted.

### DIFF
--- a/ResearchKit/Common/ORKOrderedTask.m
+++ b/ResearchKit/Common/ORKOrderedTask.m
@@ -107,6 +107,15 @@ ORKTaskProgress ORKTaskProgressMake(NSUInteger current, NSUInteger total) {
     return [_identifier hash] ^ [_steps hash];
 }
 
+- (ORKStep *)removeStepAtIndex:(NSUInteger)index {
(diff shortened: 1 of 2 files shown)


ResearchKit/Common/ORKOrderedTask.m
```diff
@@ -107,6 +107,15 @@ - (NSUInteger)hash {
     return [_identifier hash] ^ [_steps hash];
 }
 
+- (ORKStep *)removeStepAtIndex:(NSUInteger)index {
+    NSParameterAssert(index < _steps.count);
+    NSMutableArray *mutableSteps = [_steps mutableCopy];
+    ORKStep *ret = [mutableSteps objectAtIndex:index];
```
rsanchezsaez: RK doesn't normally use the `ret` naming convention. Perhaps name it `removedStep`?

YuanZhu-apple:
@syoung-smallwisdom `ORKOrderedTask` is designed to be immutable. Because modifying a task during a task run can cause unpredictable result.

PR #726, opened 2016-06-21
Fix audio-prompts-coming-from-earpiece bug

syoung-smallwisdom:
save the current AVAudioSession category before starting voice recording and then restore afterward.

We discovered a bug where running a voice task followed by a walking task would cause the spoken instructions to use the incorrect speaker. This fix addresses that issue.

### DIFF
--- a/ResearchKit/ActiveTasks/ORKAudioRecorder.m
+++ b/ResearchKit/ActiveTasks/ORKAudioRecorder.m
@@ -81,9 +81,12 @@
     return self;
 }
 
-- (void)restoreSavedAudioSessionCategory:(NSError **)error {
+- (void)restoreSavedAudioSessionCategory {
     if (_savedSessionCategory) {
-        [[AVAudioSession sharedInstance] setCategory:_savedSessionCategory error:error];
+        NSError *error;
+        if (![[AVAudioSession sharedInstance] setCategory:_savedSessionCategory error:&error]) {
+            ORK_Log_Error(@"Failed to restore the audio session category: %@", [error localizedDescription]);
+        }
         _savedSessionCategory = nil;
     }
 }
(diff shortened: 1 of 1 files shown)


ResearchKit/ActiveTasks/ORKAudioRecorder.m
```diff
@@ -42,6 +42,8 @@ @interface ORKAudioRecorder ()
 
 @property (nonatomic, copy) NSDictionary *recorderSettings;
 
+@property (nonatomic, strong) NSString *savedSessionCategory;
```
YuanZhu-apple: mark it `copy`

syoung-smallwisdom: Done

ResearchKit/ActiveTasks/ORKAudioRecorder.m
```diff
@@ -79,6 +81,13 @@ - (instancetype)initWithIdentifier:(NSString *)identifier
     return self;
 }
 
+- (void)restoreSavedAudioSessionCategory {
+    if (_savedSessionCategory) {
+        [[AVAudioSession sharedInstance] setCategory:_savedSessionCategory error:nil];
```
YuanZhu-apple: Should we error out?

syoung-smallwisdom:
@YuanZhu-apple Updated to catch the error, if any, and forward it. Not sure why the original implementation did not include this, but it probably should.

YuanZhu-apple:
@syoung-smallwisdom Now I cannot see the `restoreSavedAudioSessionCategory` in the stop action. Shouldn't it be there?

syoung-smallwisdom:
@YuanZhu-apple Yup, thanks for catching. It had been implemented to be called from two places in the code. I moved it back into the `-doStopRecording` method and just log the error since an error here would not affect this task, but only tasks following it.

YuanZhu-apple:
@syoung-smallwisdom Thanks!
r=me.

PR #727, opened 2016-06-21
Added delegate method for discarding results without confirming cancel.

syoung-smallwisdom:
Adding this `optional` delegate method removes the necessity to jump through a number of hoops and implement a custom button item (text and action) for the cancel button in `-stepViewControllerWillAppear:` delegate method.

### DIFF
--- a/ResearchKit/Common/ORKTaskViewController.h
+++ b/ResearchKit/Common/ORKTaskViewController.h
@@ -118,6 +118,22 @@ task view controller and pass that data to `initWithTask:restorationData:` when 
 - (BOOL)taskViewControllerSupportsSaveAndRestore:(ORKTaskViewController *)taskViewController;
 
 /**
+ Asks the delegate if the cancel action should be confirmed
+ 
+ The task view controller calls this method to determine whether or not to confirm
+ result save or discard when user attempts to cancel a task that is in progress.
+ 
+ If this method is not implemented, the task view controller assumes that cancel should be confirmed.
+ If this method returns `YES`, then cancel action will be confirmed.
+ If this method returns `NO`, then the results will immediately be discarded.
+ 
+ @param taskViewController  The calling `ORKTaskViewController` instance.
+ 
+ @return `YES` to confirm cancel action; `NO` to immediately discard the results.
+ */
+- (BOOL)taskViewControllerShouldConfirmCancel:(ORKTaskViewController *)taskViewController;
+
+/**
  Asks the delegate if there is Learn More content for this step.
  
  The task view controller calls this method to determine whether a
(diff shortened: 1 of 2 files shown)


YuanZhu-apple:
@syoung-smallwisdom Without this change, you can modify the behavior of cancel button directly.

https://github.com/ResearchKit/ResearchKit/blob/b0f5e2cbb7fbea6c3905c87597d4e72a6709ba75/ResearchKit/Common/ORKStepViewController.h#L284

syoung-smallwisdom:
@YuanZhu-apple I am aware that this is possible without adding the optional delegate method. Again, this is a case of overloading `stepViewControllerWillAppear` in a way that is redundant (since it must be implemented for each delegate) and an ackward work-around.  There are certainly cases where changing the default behavior of the cancel/next buttons in `stepViewControllerWillAppear` is appropriate and it can be done, but it is not ideal.

Example:

With delegate method: The code is clean, simple and easy to explain and maintain
````
    func taskViewControllerShouldConfirmCancel(taskViewController: ORKTaskViewController) -> Bool {
        // Always override the default action and do not confirm cancellation of 
        // a task. The confirmation UI is confusing to participants who are unfamiliar with Apple
        // conventions. Since this is an app designed for use on a phone specific to a short-term
        // clinical trial, we cannot assume that the user is a typical iPhone user.
        return false
    }
````
OR:
````
    func taskViewControllerShouldConfirmCancel(taskViewController: ORKTaskViewController) -> Bool {
        if let task = taskViewController.task as? ORKOrderedTask,
            let step = taskViewController.currentStepViewController?.step
            where task.indexOfStep(step) == 0 {
            
            // If this is the first step then do not confirm discard of any results (even if results have been added). 
            // Just exit.
            return false
        }
        
        return true
    }
````

Verses overloading `stepViewControllerWillAppear`: Difficult to maintain and confusing
````
    // keep a pointer to the current task view controller
    var currentTaskViewController: ORKTaskViewController!
    
    override func taskViewController(taskViewController: ORKTaskViewController, stepViewControllerWillAppear stepViewController: ORKStepViewController) {
        
        if taskViewControllerShouldConfirmCancel(taskViewController) {
            // set the pointer to the current task view controller
            currentTaskViewController = taskViewController
            // replace default implementation of the cancel button with our own
            stepViewController.cancelButtonItem = UIBarButtonItem(barButtonSystemItem: .Cancel, target: self, action: #selector(self.cancelTaskWithoutConfirmation))
        }

        // call through to super
        super.taskViewController(taskViewController, stepViewControllerWillAppear: stepViewController)
    }
    
    override func taskViewController(taskViewController: ORKTaskViewController, didFinishWithReason reason: ORKTaskViewControllerFinishReason, error: NSError?) {
        
        // nil out the pointer to the current task view controller
        currentTaskViewController = nil
        
        // call through to super
        super.taskViewController(taskViewController, didFinishWithReason: reason, error: error)
    }
    func cancelTaskWithoutConfirmation() {
        self.taskViewController(currentTaskViewController, didFinishWithReason: .Discarded, error: nil)
    }
````

YuanZhu-apple:
@syoung-smallwisdom Thanks!
r=me.

PR #728, opened 2016-06-21
Expose method for getting the index of a step from ORKOrderedTask

syoung-smallwisdom:
Exposing this method allows subclasses of ORKOrderedTask to access this method. Additionally, allows the delegate to check the position of a step in the overall order for custom usage.


ResearchKit/Common/ORKOrderedTask_Internal.h
```diff
@@ -36,8 +36,6 @@ NS_ASSUME_NONNULL_BEGIN
 
 @interface ORKOrderedTask ()
```
YuanZhu-apple: Should this file be removed?

YuanZhu-apple:
@syoung-smallwisdom Consider add a unit test for it?

ResearchKit/Common/ORKOrderedTask_Internal.h
syoung-smallwisdom: Wasn't sure about that. It isn't being used but I didn't know if there might be other devs who had added something to that file. I can remove it. :)

syoung-smallwisdom:
@YuanZhu-apple Added unit test and removed empty implementation file.

syoung-smallwisdom:
@YuanZhu-apple I rebased this PR on top of master so that it will auto merge.

ResearchKitTests/ORKTaskTests.m
```diff
@@ -1338,4 +1338,24 @@ - (void)testResultPredicates {
...
+    // get the second step
+    ORKStep *step1 = [task stepWithIdentifier:ORKInstruction1StepIdentifier];
+    XCTAssertNotNil(step1);
+    XCTAssertEqual([task indexOfStep:step1], 1);
+    
+    // get the last step
+    ORKStep *stepLast = [task.steps lastObject];
+    XCTAssertNotNil(stepLast);
+    XCTAssertEqual([task indexOfStep:stepLast], task.steps.count - 1);
+
```
YuanZhu-apple: Add a test for NSNotFound

YuanZhu-apple:
@syoung-smallwisdom Out of my curiosity, shouldn't just let the delegate check the step identifier instead?

ResearchKitTests/ORKTaskTests.m
syoung-smallwisdom: Test added

syoung-smallwisdom:
@YuanZhu-apple 

Example Usage:
````
    public func shouldHideCancelForStep(step: ORKStep, taskViewController: ORKTaskViewController) -> Bool {
        
        // Return false if cancel is *not* disabled
        guard let schedule = scheduledActivityForTaskViewController(taskViewController),
            let taskRef = bridgeInfo.taskReferenceForSchedule(schedule) where taskRef.cancelDisabled
        else {
            return false
        }
        
        // If the task does not respond then assume that cancel should be hidden for all steps
        guard let task = taskViewController.task as? SBATaskExtension
        else {
            return true
        }

        // Otherwise, do not disable the first step IF and ONLY IF there are more than 1 steps.
        return task.stepCount() == 1 || task.indexOfStep(step) > 0;
    }
````

In this case, I am using a protocol to extend ORKTask. If it responds to the protocol, even if it does *not* inherit from `ORKOrderedTask` then it will return a valid response.

Another example is a subclass of ORKOrderedTask that can use this function and does not have to define a *different* method signature to avoid override conflict.

syoung-smallwisdom:
@YuanZhu-apple Rebased against master to resolve merge conflicts.

PR #729, opened 2016-06-21
ORKResult start/end dates should not be nullable

syoung-smallwisdom:
By default, these values are set on instantiation to `now`. Not allowing null values for these simplifies archiving of results for upload to a server that is expecting non-nil values.


YuanZhu-apple:
r=me.

PR #730, opened 2016-06-21
Add optional property to ORKQuestionResult to record the question text.

syoung-smallwisdom:
Adding a nullable `questionText` property on an `ORKQuestionResult` allows for simpler schema handling when serializing a result for saving it to a server.


YuanZhu-apple:
@syoung-smallwisdom I am not sure this is the right approach. Since the result object and step object shares the same identifier. The server should be able to recover all the information from that identifier.
  
And if the `text` is added, then how about the `title` and even answer format?

syoung-smallwisdom:
@YuanZhu-apple It was a request by our researchers to facilitate readability of the json, but we have other ways of adding this.

syoung-smallwisdom:
closing. Will handle in our API.

PR #731, opened 2016-06-22
Added a optional ambient noise check to the voice task.

syoung-smallwisdom:
During user feedback and via researcher analysis, we were finding that the ambient noise in a room where the participant was attempting to record their voice was too loud. We added a navigation rule to the task to check for ambient noise level and direct the participant to move to a quieter location if necessary.

### DIFF
--- a/ResearchKit/ActiveTasks/ORKAudioLevelNavigationRule.h
+++ b/ResearchKit/ActiveTasks/ORKAudioLevelNavigationRule.h
@@ -1,5 +1,5 @@
 /*
- Copyright (c) 2015, Apple Inc. All rights reserved.
+ Copyright (c) 2016, Apple Inc. All rights reserved.
  
  Redistribution and use in source and binary forms, with or without modification,
  are permitted provided that the following conditions are met:
--- a/ResearchKit/ActiveTasks/ORKAudioLevelNavigationRule.h
+++ b/ResearchKit/ActiveTasks/ORKAudioLevelNavigationRule.h
@@ -1,0 +1,64 @@
+/*
+ Copyright (c) 2015, Apple Inc. All rights reserved.
+ 
+ Redistribution and use in source and binary forms, with or without modification,
+ are permitted provided that the following conditions are met:
+ 
+ 1.  Redistributions of source code must retain the above copyright notice, this
+ list of conditions and the following disclaimer.
+ 
+ 2.  Redistributions in binary form must reproduce the above copyright notice,
+ this list of conditions and the following disclaimer in the documentation and/or
+ other materials provided with the distribution.
+ 
+ 3.  Neither the name of the copyright holder(s) nor the names of any contributors
+ may be used to endorse or promote products derived from this software without
+ specific prior written permission. No license is granted to the trademarks of
+ the copyright holders even if such marks are included in this software.
+ 
+ THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE
+ FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#import "ORKStepNavigationRule.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface ORKAudioLevelNavigationRule : ORKStepNavigationRule
+
+/**
+ Returns an initialized direct-step navigation rule using the specified destination step identifier.
+ 
+ @param audioLevelStepIdentifier   The identifier of the step with the audio file to check.
+ @param defaultStepIdentifier   The identifier of the destination step if audio test passes.
+ 
+ @return A audio level step navigation rule.
+ */
+- (instancetype)initWithAudioLevelStepIdentifier:(NSString *)audioLevelStepIdentifier
+                           defaultStepIdentifier:(NSString *)defaultStepIdentifier
+                               recordingSettings:(NSDictionary *)recordingSettings NS_DESIGNATED_INITIALIZER;
+
+/**
+ Returns a new direct-step navigation rule initialized from data in a given unarchiver.
+ 
+ @param aDecoder    The coder from which to initialize the step navigation rule.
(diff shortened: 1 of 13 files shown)


ResearchKit/Common/ORKOrderedTask.h
```diff
@@ -236,6 +236,41 @@ typedef NS_OPTIONS(NSUInteger, ORKPredefinedTaskOption) {
...
+ @param shortSpeechInstruction  Instructional content shown during audio recording. If the value of
+ this parameter is `nil`, default localized text is used.
+ @param duration                The length of the count down timer that runs while audio data is
+ collected.
+ @param recordingSettings       See "AV Foundation Audio Settings Constants" for possible values.
+ @param options                 Options that affect the features of the predefined task.
+ 
+ @return An active audio task that can be presented with an `ORKTaskViewController` object.
+ */
++ (ORKOrderedTask *)audioLevelNavigableTaskWithIdentifier:(NSString *)identifier
```
YuanZhu-apple: Do we need a new task here or just add a parameter to the existing audio task?

syoung-smallwisdom: Existing apps already use this task with the assumption that it is an `ORKOrderedTask` and *not* an `ORKNavigableOrderedTask`. Because the task steps in an `ORKOrderedTask` are immutable (and the publicly exposed accessor is *not* used internally), any app that needs to mutate the task has to do so by creating this task and then mutating the steps and instantiating a *new* `ORKOrderedTask`. Adding the branching step will break those implementations. That said, I realize that the method signature should explicitly return an `ORKNavigableOrderedTask` and will update the method signature.

syoung-smallwisdom:
@YuanZhu-apple Updated method signature and copyrights.

ResearchKit/ActiveTasks/ORKAudioLevelNavigationRule.h
```diff
@@ -0,0 +1,65 @@
...
+/**
+ Returns an initialized direct-step navigation rule using the specified destination step identifier.
+ 
+ @param audioLevelStepIdentifier   The identifier of the step with the audio file to check.
+ @param defaultStepIdentifier   The identifier of the destination step if audio test passes.
+ 
+ @return A audio level step navigation rule.
+ */
+- (instancetype)initWithAudioLevelStepIdentifier:(NSString *)audioLevelStepIdentifier
+                           defaultStepIdentifier:(NSString *)defaultStepIdentifier
```
YuanZhu-apple: Maybe `defaultDestinationStepIdentifier` is better? Or just `destinationStepIdentifier`?

YuanZhu-apple:
@syoung-smallwisdom Overall, it pretty cool! Just one nit picking.

syoung-smallwisdom:
@YuanZhu-apple Changed name to "destination" which I agree makes more sense. Thanks.

YuanZhu-apple:
@syoung-smallwisdom Not sure if you agree, what if we reuse the existing one by adding boolean to control the testing behavior?
1. Old behavior is still supported
2. NavigableOrderedTask is still an OrderedTask, developers don't need to know the difference.

syoung-smallwisdom:
@YuanZhu-apple 

My concern is that the "old" way of modifying a task to insert/delete steps is to do the following:
1. Make a mutable array of the steps
2. Modify the step array
3. Instantiate a new `ORKOrderedTask` with the mutated step array

Because I am changing the class instance that is returned (in order to be add an optional step), I am concerned that just adding an optional parameter might break someone else's existing app logic.

syoung-smallwisdom:
@YuanZhu-apple 

Also, I'm not sure that I understand the reasoning behind not using deprecation to give developers a warning that a method will be deleted in future versions. This gives the user an opportunity to change their method signatures and includes a hint as to how to do so.

ResearchKit/Common/ORKOrderedTask.h
```diff
@@ -213,27 +217,39 @@ typedef NS_OPTIONS(NSUInteger, ORKPredefinedTaskOption) {
...
  
  @return An active audio task that can be presented with an `ORKTaskViewController` object.
  */
++ (ORKNavigableOrderedTask *)audioTaskWithIdentifier:(NSString *)identifier
+                              intendedUseDescription:(nullable NSString *)intendedUseDescription
+                                   speechInstruction:(nullable NSString *)speechInstruction
+                              shortSpeechInstruction:(nullable NSString *)shortSpeechInstruction
+                                            duration:(NSTimeInterval)duration
+                                   recordingSettings:(nullable NSDictionary *)recordingSettings
+                                     checkAudioLevel:(BOOL)checkAudioLevel
```
YuanZhu-apple: Can you document it?

syoung-smallwisdom:
@YuanZhu-apple 
Documentation is updated and I also added a unit test to confirm that the optional step to check the noise level is really and truly optional (and reverse-compatible). Note: adding the test required rebasing against master to allow merge.

ResearchKit/Common/ORKOrderedTask.h
syoung-smallwisdom: Done

ResearchKit/Common/ORKOrderedTask.h
```diff
@@ -226,40 +228,59 @@ typedef NS_OPTIONS(NSUInteger, ORKPredefinedTaskOption) {
...
+ recording begins. If the value of this parameter is `nil`,
+ default localized text is used.
  @param shortSpeechInstruction  Instructional content shown during audio recording. If the value of
-                                    this parameter is `nil`, default localized text is used.
+ this parameter is `nil`, default localized text is used.
  @param duration                The length of the count down timer that runs while audio data is
-                                    collected.
+ collected.
  @param recordingSettings       See "AV Foundation Audio Settings Constants" for possible values.
+ @param checkAudioLevel         If `YES` the add navigational rules to check the background noise level.
```
YuanZhu-apple: If `YES` ~~the~~ then add navigational rules to check the background noise level.

syoung-smallwisdom: @YuanZhu-apple Fixed typo

YuanZhu-apple:
```
Test Case '-[ORKJSONSerializationTests testORKSerialization]' started.
/Users/yuanzhu/resource/github/ResearchKit/Testing/ORKTest/ORKTestTests/ORKJSONSerializationTests.m:325: error: -[ORKJSONSerializationTests testORKSerialization] : ((unregisteredList.count) equal to (0)) failed: ("1") is not equal to ("0") - Classes didn't implement ORKSerialization (
    ORKAudioLevelNavigationRule
)
Test Case '-[ORKJSONSerializationTests testORKSerialization]' failed (0.054 seconds).
```

Can you update ORKSerialization for `ORKAudioLevelNavigationRule`? Thanks!

syoung-smallwisdom:
@YuanZhu-apple Updated

YuanZhu-apple:
@syoung-smallwisdom Thanks!

ResearchKit/ActiveTasks/ORKAudioLevelNavigationRule.h
```diff
@@ -0,0 +1,65 @@
...
+ CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#import <ResearchKit/ORKStepNavigationRule.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface ORKAudioLevelNavigationRule : ORKStepNavigationRule
```
YuanZhu-apple: Add `ORK_CLASS_AVAILABLE` above this line.
Can you please also do the same for `ORKBarGraphChartView`, it also missed a declaration. Thanks!

syoung-smallwisdom: @YuanZhu-apple Done, Sorry, missed that.

YuanZhu-apple:
Sorry, there is one more `ORKPredicateSkipStepNavigationRule`.
I promise this is the last one:) 
Error emits when archiving the ORKTest app.

YuanZhu-apple:
r=me. Thanks!

PR #732, opened 2016-06-22
Added a general purpose step table view controller

syoung-smallwisdom:
`ORKTableStepViewController` is designed to take advantage of the internal class methods used by `ORKFormStepViewController`, `ORKQuestionStepViewController` and `ORKReviewStepViewController` to allow for a consistent UI in a custom implementation.

### DIFF
--- a/ResearchKit/Common/ORKTableStep.h
+++ b/ResearchKit/Common/ORKTableStep.h
@@ -40,18 +40,12 @@ NS_ASSUME_NONNULL_BEGIN
  To use `ORKTableStep`, instantiate the object, fill in its properties, and include it
  in a task. Next, create a task view controller for the task and present it.
  
- Each question in the form is represented by an `ORKFormItem` object. The form items have an
- `optional` property that defaults to `YES`. All required questions need to be answered for the
- Continue button to be enabled. If all the form items are optional, at least one question needs to
- be answered for the Continue button to be enabled. You can allow the user to completely skip a
- form step using the Skip button, even if it has required form items, by setting the form step
- `optional` property to yes.
+ The base class implementation will instatiate a read-only `ORKTableStepViewController` to display 
+ the list of items using `UITableViewCell` with the text set to the `-description` for each item in 
+ the `items` array.
  
- The form can be broken into sections by using an `ORKFormItem` object that includes only a section
- title.
- 
- The result of a form step is an `ORKStepResult` object that includes a child `ORKQuestionResult`
- object for each form item.
+ Customization can be handled by overriding the base class implementations in either `ORKTableStep`
+ or `ORKTableStepViewController`.
  */
 
 FOUNDATION_EXPORT NSString *const ORKBasicCellReuseIdentifier;
--- a/ResearchKit/Common/ORKTableStep.h
+++ b/ResearchKit/Common/ORKTableStep.h
@@ -1,0 +1,111 @@
+/*
+ Copyright (c) 2016, Sage Bionetworks
+ 
+ Redistribution and use in source and binary forms, with or without modification,
+ are permitted provided that the following conditions are met:
+ 
+ 1.  Redistributions of source code must retain the above copyright notice, this
+ list of conditions and the following disclaimer.
+ 
+ 2.  Redistributions in binary form must reproduce the above copyright notice,
+ this list of conditions and the following disclaimer in the documentation and/or
+ other materials provided with the distribution.
+ 
+ 3.  Neither the name of the copyright holder(s) nor the names of any contributors
+ may be used to endorse or promote products derived from this software without
+ specific prior written permission. No license is granted to the trademarks of
+ the copyright holders even if such marks are included in this software.
+ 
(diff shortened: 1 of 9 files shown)


YuanZhu-apple:
@syoung-smallwisdom Can you give us a use case?

syoung-smallwisdom:
@YuanZhu-apple 

Example implementation:
https://github.com/Sage-Bionetworks/BridgeAppSDK/blob/master/BridgeAppSDK/SBAProgressStep.swift

Note: This example has not been updated to include the newer method for instantiating a ORKStepViewController for a given step.

YuanZhu-apple:
@syoung-smallwisdom It would be nice to see it in a screen shot:)

syoung-smallwisdom:
simulator screen shot jun 23 2016 12 00 19 pm

syoung-smallwisdom:
@YuanZhu-apple Here you go!

ResearchKit/Common/ORKTableStepViewController.h
```diff
@@ -0,0 +1,75 @@
...
+ OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#import <ResearchKit/ResearchKit.h>
+
+ORK_CLASS_AVAILABLE
+@interface ORKTableStepViewController : ORKStepViewController <UITableViewDataSource, UITableViewDelegate>
+
+@property (nonatomic, readonly) UITableView *tableView;
```
YuanZhu-apple: Please document this attribute.

ResearchKit/Common/ORKTableStepViewController.h
```diff
@@ -0,0 +1,75 @@
...
+ */
+- (BOOL)continueButtonEnabled;
+
+/**
+ Required implementations for a data source. By default there is 1 section in the table.
+ 
+ @param  tableView      The tableview used to display rows
+ @return                The number of sections in the tableview
+ */
+- (NSInteger)numberOfSectionsInTableView:(UITableView *)tableView;
```
YuanZhu-apple: Since this method is not required to be override like other delegate/datasource functions, do we need to list it here?

YuanZhu-apple:
@syoung-smallwisdom Thanks!

YuanZhu-apple:
Is `ORKTableStepViewController` an abstract class and always require a subclassing? Do you think we should prevent direct initialization?

syoung-smallwisdom:
@YuanZhu-apple 

Yes, this is an abstract class. It is similar to `ORKStepViewController` or `UITableViewController` which, if instantiated directly, will show nothing.

No, I would prefer *not* to prevent direct initialization. Throwing an exception on init if this class `isMemberOfClass:` does not add any value and just makes for more code that would need to be maintained.

YuanZhu-apple:
@syoung-smallwisdom IMO, to make it more developer friendly, nice to add a ORKTableInfomationStep to carry strings for each cell, and a concrete implementation of ORKTableStepViewController to present those strings in table. What do you think?

syoung-smallwisdom:
@YuanZhu-apple I can take a look at adding that next week.

syoung-smallwisdom:
@YuanZhu-apple Added `ORKTableStep` which can implement either a base class implementation where it just lists strings *or* supports overriding to customize the cells.

YuanZhu-apple:
@syoung-smallwisdom Do you think we should make table cell un-selectable by default?

syoung-smallwisdom:
@YuanZhu-apple Good point. Updated, thanks.

YuanZhu-apple:
@syoung-smallwisdom Can you resolve the merge conflict from the recent master branch change?

syoung-smallwisdom:
@YuanZhu-apple Rebased against master

ResearchKit/Common/ORKTableStep.h
```diff
@@ -0,0 +1,111 @@
...
+ in a task. Next, create a task view controller for the task and present it.
+ 
+ Each question in the form is represented by an `ORKFormItem` object. The form items have an
+ `optional` property that defaults to `YES`. All required questions need to be answered for the
+ Continue button to be enabled. If all the form items are optional, at least one question needs to
+ be answered for the Continue button to be enabled. You can allow the user to completely skip a
+ form step using the Skip button, even if it has required form items, by setting the form step
+ `optional` property to yes.
+ 
+ The form can be broken into sections by using an `ORKFormItem` object that includes only a section
```
YuanZhu-apple: Wrong documentation...

ResearchKit/Common/ORKTableStep.h
```diff
@@ -0,0 +1,111 @@
...
+ title.
+ 
+ The result of a form step is an `ORKStepResult` object that includes a child `ORKQuestionResult`
+ object for each form item.
```
YuanZhu-apple: And here.

syoung-smallwisdom:
@YuanZhu-apple Comments updated

YuanZhu-apple:
@syoung-smallwisdom Thanks!
R=me.

PR #733, opened 2016-06-22
Changes to allow for more flexible onboarding UI

syoung-smallwisdom:
In some cases, given name and family name are not appropriate information for a consent document. Our consent process sometimes takes a full name. In other countries there are legal constraints requiring that the user be de-indentified.  In many cases, the user can only consent if they are over 18. Making this step inherit from `ORKFormStep` allows greater flexibility within the consent process without having to hack in work-arounds.

### DIFF
--- a/ResearchKit/Consent/ORKConsentReviewStep.h
+++ b/ResearchKit/Consent/ORKConsentReviewStep.h
@@ -31,7 +31,6 @@
 
 #import <ResearchKit/ResearchKit.h>
 
-
 NS_ASSUME_NONNULL_BEGIN
 
 @class ORKConsentDocument;
@@ -59,7 +58,7 @@ NS_ASSUME_NONNULL_BEGIN
  present the task in a task view controller.
  */
 ORK_CLASS_AVAILABLE
-@interface ORKConsentReviewStep : ORKStep
+@interface ORKConsentReviewStep : ORKFormStep
 
 /// @name Initialization.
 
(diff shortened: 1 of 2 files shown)


YuanZhu-apple:
@syoung-smallwisdom A few question
1. Why `stepViewControllerDidFinish:` is required? Can you use `stepViewControllerWillAppear:` instead?

2. Have the fixed entries on the consent step ensures the result can be written into the consent document. Not sure the consequences of making it flexible.

syoung-smallwisdom:
@YuanZhu-apple

1. `stepViewControllerDidFinish`

Providing an implementation in `stepViewControllerWillAppear` requires that the delegate knows the order of the steps and that the step about to be displayed is the one following the step that was previously displayed. This is a brittle assumption and it's not really telling you what you need to know which is that a step has finished. 

In this case, our governance team required a change to the design for mPower to be made available in Luxembourg. We had to rearrange the steps and add our own custom implementation to meet their consent requirements. The app was failing the onboarding process because we were looking for the wrong step (we were using `stepViewControllerWillAppear` in the original `AppCore` implementation). The `stepViewControllerDidFinish` method allows us to know when the ORKConsentReviewStep has finished which is really what we are interested in knowing.

2. The consent document is shown to the user *before* the user adds their name and signature. The purpose of the ORKConsentDocument including these signatures is to be able to create a printable version that includes all signers of the document. This change does *not* affect the default behavior. That said, I have changed it to be additive to the given/family name rather than replacing them.

YuanZhu-apple:
@syoung-smallwisdom `stepViewControllerDidFinish` sounds the step is completed to me, how about rename it to `stepViewControllerWillDisappear`.

And actually two different things are in this same PR, can you separate them to two? Thanks!

syoung-smallwisdom:
@YuanZhu-apple 

Pulled out the code related to adding a delegate method for when a step view controller will disappear.

YuanZhu-apple:
@syoung-smallwisdom After thinking about this, I feels that:
1) Full name can be an option
2) Age restriction can be applied before consent step.

Overall I am opposing to change ORKConsentReviewStep to a ORFormStep.

syoung-smallwisdom:
closing.

PR #735, opened 2016-06-22
Allow steps to instantiate the view controller

syoung-smallwisdom:
`- (Class)stepViewControllerClass` is defined as an internal method and is not exposed in the public API. This implementation allows for subclasses of ORKStep to return a view controller that is appropriate to their design.

In response to the explanation given in https://github.com/ResearchKit/ResearchKit/pull/719 for closing that pull request: Overriding a method that is marked as internal is brittle (internal methods cannot be assumed to remain and are not marked as deprecated when they might go away with a future release). Additionally, doing so with a Swift implementation will fail when the app is archived.

### DIFF
--- a/ResearchKit/Common/ORKStep.h
+++ b/ResearchKit/Common/ORKStep.h
@@ -37,6 +37,9 @@ NS_ASSUME_NONNULL_BEGIN
 
 ORK_EXTERN NSString *const ORKNullStepIdentifier ORK_AVAILABLE_DECL;
 
+@class ORKStepViewController;
+@class ORKResult;
+
 @protocol ORKTask;
 
 /**
@@ -178,6 +181,18 @@ ORK_CLASS_AVAILABLE
  properties, and must call super.
  */
 - (void)validateParameters;
+
+/**
+ Instantiates a step view controller for this class.
+ 
+ This method is called when a step is about to be presented. The default implementation returns
+ a view controller that is appropriate to this step. 
+ 
+ @param result    The result associated with this step
+ 
+ @return A newly initialized step view controller.
+ */
+- (ORKStepViewController *)instantiateStepViewControllerWithResult:(ORKResult *)result;
 
 @end
 
--- a/ResearchKit/Common/ORKStep.h
+++ b/ResearchKit/Common/ORKStep.h
@@ -183,6 +183,12 @@ ORK_CLASS_AVAILABLE
 - (void)validateParameters;
 
 /**
+ Returns the class that the task view controller should instantiate to display
+ this step.
+ */
+- (Class)stepViewControllerClass;
+
+/**
  Instantiates a step view controller for this class.
  
  This method is called when a step is about to be presented. The default implementation returns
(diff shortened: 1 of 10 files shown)


YuanZhu-apple:
@syoung-smallwisdom Do you think would it be easier to just expose `stepViewControllerClass()`?

syoung-smallwisdom:
@YuanZhu-apple Exposing `stepViewControllerClass()` would only allow for creating a view controller using the default `ORKStepViewController` initializer (which is also defined as internal and cannot be overridden by inheriting classes). It does not allow for instantiating using a nib or storyboard. Additionally, it does not allow the step implementation to mutate based on the results which in some cases makes more sense than performing that mutation in the stepViewController.

YuanZhu-apple:
@syoung-smallwisdom It does make sense. Can you send an email to researchkit@apple.com, and I will send you a contribution term which is required for me to accept your PR. Thanks!

syoung-smallwisdom:
@YuanZhu-apple I'm now added as a contributor.

YuanZhu-apple:
>Exposing stepViewControllerClass() would only allow for creating a view controller using the default ORKStepViewController initializer (which is also defined as internal and cannot be overridden by inheriting classes). 
This can be done by making `initWithStep:result` public.

> It does not allow for instantiating using a nib or storyboard. 
If a VC is from a nib or storyboard, how you do you set the step and result on it? 

> Additionally, it does not allow the step implementation to mutate based on the results which in some cases makes more sense than performing that mutation in the stepViewController.

Do you mean the stepVC should be able to mutate based on the results? If so, does it have to be done in the `init`?

syoung-smallwisdom:
@YuanZhu-apple 

We are frequently required to add a lot of circuitous work-arounds that are brittle and hard to maintain across different apps because of things like `stepViewControllerClass` being private. While exposing this method will work *for some cases* it will not work for the case where (A) the view controller is instantiated from a nib or storyboard or (B) the step has a model that is different depending upon the previous result. AppCore has several examples of work-arounds (that are circuitous and difficult to maintain) because of case A. I don't have an example of case B in our code - I was merely intending to suggest that there might be use cases for allowing the step to instantiate the view controller in addition to the use cases for which I've already had to find work-arounds.

(We do have a case where the results from a previous step influence what is shown in the next step, but that is handled internally to our API using a specialized subclass of ORKOrderedTask.)

YuanZhu-apple:
@syoung-smallwisdom Thanks for explaining.
Can you write a subclass use case in `ORKTest` App? 
Because I am not sure how a subclass should override `instantiateStepViewControllerWithResult:`, for a few reasons:
1. Calling `[super instantiateStepViewControllerWithResult:]` leads to be not able provide a vc class.
2. If super call is by passed, and developer has no access to `initWithStep:result`, then how to set the result?

syoung-smallwisdom:
@YuanZhu-apple Added unit test and publicly expose method for `stepViewControllerClass` and `initWithStep:Result:`

YuanZhu-apple:
@syoung-smallwisdom Can you write a test case in ORKTest App as well?

syoung-smallwisdom:
@YuanZhu-apple Moved the test case into `ORKTest`

YuanZhu-apple:
@syoung-smallwisdom Thanks!
r=me.

Issue #713, opened 2016-06-19
Update to support iOS 10 SDK HealthKit API changes


king7532:
ResearchKit framework fails to build using Xcode 8 and the iOS 10 SDK in [`ORKHealthQuantityTypeRecorder.m:131`](https://github.com/ResearchKit/ResearchKit/blob/9f5caad71d474af0014ff4885ee3292fe7434e24/ResearchKit/ActiveTasks/ORKHealthQuantityTypeRecorder.m#L131) because the following HealthKit API has been removed:

`-[HKAnchoredObjectQuery initWithType:predicate:anchor:limit:completionHandler:]`

See the [HealthKit Changes for Objective-C in iOS 10.0 API Diffs](https://developer.apple.com/library/prerelease/content/releasenotes/General/iOS10APIDiffs/Objective-C/HealthKit.html)

Instead it has been replaced with:

``` objc
- (instancetype)initWithType:(HKSampleType *)type
                   predicate:(nullable NSPredicate *)predicate
                      anchor:(nullable HKQueryAnchor *)anchor
                       limit:(NSUInteger)limit
              resultsHandler:(void(^)(HKAnchoredObjectQuery *query, NSArray<__kindof HKSample *> * _Nullable sampleObjects, NSArray<HKDeletedObject *> * _Nullable deletedObjects, HKQueryAnchor * _Nullable newAnchor, NSError * _Nullable error))handler HK_AVAILABLE_IOS_WATCHOS(9_0, 2_0);
```

king7532:
The above pull request 710 aleady implements the changes.

YuanZhu-apple:
@king7532 I'd like to keep `master` as it is until we full embrace XCode8 GM. For now, the xcode8 branch is availble to use.

umerkhan-apple:
Merged!

PR #736, opened 2016-06-22
Fix ORKHealthQuantityTypeRecorder to allow for iOS 10 and iOS 8 support

syoung-smallwisdom:
This is a fix for https://github.com/ResearchKit/ResearchKit/issues/713

It can be safely pulled into ResearchKit *now* while still targeting iOS 8.

With Xcode 8 and iOS 10, the `init` method used in iOS 8 has been deleted. Future versions of the iOS 10 beta may include this or they may not. This implementation allows for reverse-compatibility to iOS 8 (for apps that require supporting that version) while using the newer `init` method introduced in iOS 9 for both iOS 9 and iOS 10.

### DIFF
--- a/ResearchKit/ActiveTasks/ORKHealthQuantityTypeRecorder.m
+++ b/ResearchKit/ActiveTasks/ORKHealthQuantityTypeRecorder.m
@@ -43,6 +43,8 @@
     HKHealthStore *_healthStore;
     NSPredicate *_samplePredicate;
     HKObserverQuery *_observerQuery;
+    /// Either the HKQueryAnchor object *or* NSUInteger value are tracked since the initializer for
+    /// iOS 8 and iOS 9 use different objects. Only one will actually be referenced in the initalizer.
     HKQueryAnchor *_anchor;
     NSUInteger _anchorValue;
     HKQuantitySample *_lastSample;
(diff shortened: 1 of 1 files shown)


ResearchKit/ActiveTasks/ORKHealthQuantityTypeRecorder.m
```diff
@@ -43,12 +43,27 @@ @interface ORKHealthQuantityTypeRecorder () {
     HKHealthStore *_healthStore;
     NSPredicate *_samplePredicate;
     HKObserverQuery *_observerQuery;
-    NSInteger _anchor;
+    HKQueryAnchor *_anchor;
+    NSUInteger _anchorValue;
```
YuanZhu-apple: Can you add some comment for these two anchor values?

YuanZhu-apple:
@syoung-smallwisdom Overall this looks good. I need to reconsider whether to drop iOS8 support in the next release. 
It seems you want to build apps for iOS 8-10 with Xcode 8.0?

syoung-smallwisdom:
@YuanZhu-apple 
Not quite. I want to be able to run using Xcode 8 (testing, whatnot) but still *build* and release using Xcode 7 and a target of iOS 8.4.  Since it looks like there are other folks interested in doing this as well, I figured I would post the work-around that we are using on our branch of ResearchKit.

YuanZhu-apple:
Cool.

PR #738, opened 2016-06-23
Expose method for instantiating an answer format for confirmation of another field.

syoung-smallwisdom:
This answer format is intended to be used with an `ORKFormStep` in order to confirm a previous formItem input. Example usage includes a password or participant identifier that is used to anonymously identify a study participant.

Note: I intentionally decided *against* exposing `ORKConfirmTextAnswerFormat` publicly. Because it requires referencing another form item, it is requires special-case handling in `ORKFormStepViewController`.

### DIFF
--- a/ResearchKit/Common/ORKFormStep.h
+++ b/ResearchKit/Common/ORKFormStep.h
@@ -185,6 +185,19 @@ ORK_CLASS_AVAILABLE
  */
 @property (nonatomic, copy, readonly, nullable) ORKAnswerFormat *answerFormat;
 
+/**
+ Returns an form item that can be used for confirming a text entry.
+ 
+ This form item is intended to be used with an `ORKFormStep` in order to confirm a previous
+ formItem input. Example usage includes a password or participant identifier that is used to
+ anonymously identify a study participant.
+ 
+ The answer format for this item produces an `ORKBooleanQuestionResult` object.
+ */
+- (ORKFormItem *)confirmationAnswerFormItemWithIdentifier:(NSString *)identifier
+                                                     text:(nullable NSString *)text
+                                             errorMessage:(NSString *)errorMessage;
+
 @end
 
 NS_ASSUME_NONNULL_END
--- a/ResearchKit/Common/ORKFormStep.h
+++ b/ResearchKit/Common/ORKFormStep.h
@@ -192,7 +192,18 @@ ORK_CLASS_AVAILABLE
  formItem input. Example usage includes a password or participant identifier that is used to
  anonymously identify a study participant.
  
+ Currently, only `ORKTextAnswerFormat` is supported. Unsupported `ORKAnswerFormat` types will
+ throw an exception.
+ 
  The answer format for this item produces an `ORKBooleanQuestionResult` object.
+ 
+ @param identifier      The identifier for the `ORKFormItem` that is returned.
+ @param text            The text for the `ORKFormItem` that is returned.
+ @param errorMessage    The error message to display if the fields do not match
+ 
+ @return                An `ORKFormItem` with the indicated identifier and text and an ORKAnswerFormat 
+                        that is appropriate for confirming the input form item.
+ 
  */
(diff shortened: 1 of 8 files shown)


syoung-smallwisdom:
@YuanZhu-apple 

WRT a method on `ORKFormItem`: I thought about that, but wasn't sure what the appropriate confirmation answer format would be for cases where the original ORKAnswerFormat was *not* an instance of ORKTextAnswerFormat. Return `nil` if not applicable?

WRT checking if the answer format is a text answer format, not sure that I follow your question. The unit test checks for that and drops out early if it fails because the unit test checks that the applicable properties from `ORKTextAnswerFormat` are copied. If there wasn't an early dropout then the test would fail in confusing ways because of an assumed cast.

YuanZhu-apple:
>WRT a method on ORKFormItem: I thought about that, but wasn't sure what the appropriate confirmation answer format would be for cases where the original ORKAnswerFormat was not an instance of ORKTextAnswerFormat. Return nil if not applicable?

I would just throw an exception if the answer format is not a instance of `ORKTextAnswerFormat` .

syoung-smallwisdom:
@YuanZhu-apple Changed to a method off of `ORKFormItem`.  Currently, behavior is only defined for `ORKTextAnswerFormat`.

ResearchKit/Common/ORKFormStep.h
```diff
@@ -185,6 +185,19 @@ ORK_CLASS_AVAILABLE
...
+/**
+ Returns an form item that can be used for confirming a text entry.
+ 
+ This form item is intended to be used with an `ORKFormStep` in order to confirm a previous
+ formItem input. Example usage includes a password or participant identifier that is used to
+ anonymously identify a study participant.
+ 
+ The answer format for this item produces an `ORKBooleanQuestionResult` object.
+ */
+- (ORKFormItem *)confirmationAnswerFormItemWithIdentifier:(NSString *)identifier
```
YuanZhu-apple: Please document each of the parameters individually.

YuanZhu-apple:
Also can you provide a test case in `ORKTest` App to verify it works as expected? Some test cases in my mind, 
itemA is optional + itemB is optional
itemA is optional + itemB is non-optional
itemA is non-optional + itemB is optional
itemA is non-optional + itemB is non-optional

ResearchKit/Common/ORKFormStep.h
syoung-smallwisdom: Done

syoung-smallwisdom:
@YuanZhu-apple 
I have added to the existing unit test to check for the case where the input form item is optional (and thus the confirm answer field should also be optional) and to check the original identifier and error message.

syoung-smallwisdom:
@YuanZhu-apple I've added an example to ORKTest

Testing/ORKTest/ORKTest/MainViewController.m
```diff
@@ -3732,4 +3736,75 @@ - (ORKOrderedTask *)makeLocationTask {
...
+    step1.title = @"Confirmation Form Items Survey";
+    [steps addObject:step1];
+    
+    // Create a step for entering password with confirmation
+    ORKFormStep *step2 = [[ORKFormStep alloc] initWithIdentifier:@"confirmationForm.step2" title:@"Password" text:nil];
+    [steps addObject:step2];
+    
+    {
+        ORKTextAnswerFormat *answerFormat = [ORKAnswerFormat textAnswerFormat];
+        answerFormat.multipleLines = NO;
```
YuanZhu-apple: What if the answerFormat is `multipleLines`?

syoung-smallwisdom:
@YuanZhu-apple Added assert to throw if the answer format is multiple lines.

YuanZhu-apple:
r=me. @syoung-smallwisdom Can you resolve the conflict?

syoung-smallwisdom:
@YuanZhu-apple 
Rebased against master and ran unit tests again. Ready to merge.

YuanZhu-apple:
Merging.

PR #739, opened 2016-06-23
Add delegate method to task view controller for step will disappear

syoung-smallwisdom:
See PR https://github.com/ResearchKit/ResearchKit/pull/733 for conversation.

In this case, our governance team required a change to the design for mPower to be made available in Luxembourg. We had to rearrange the onboarding steps and add our own custom implementation to meet their consent requirements. The app was failing the onboarding process because we were looking for the wrong step (we were using stepViewControllerWillAppear in the original AppCore implementation). The `stepViewControllerWillDisappear` method allows us to know when the `ORKConsentReviewStep` has finished which is really what we are interested in knowing.

### DIFF
--- a/ResearchKit/Common/ORKTaskViewController.h
+++ b/ResearchKit/Common/ORKTaskViewController.h
@@ -212,6 +212,16 @@ task view controller and pass that data to `initWithTask:restorationData:` when 
 - (void)taskViewController:(ORKTaskViewController *)taskViewController stepViewControllerWillAppear:(ORKStepViewController *)stepViewController;
 
 /**
+ Tells the delegate that a step has finished.
+ 
+ The task view controller calls this method after a stepViewController has finished.
+ 
+ @param taskViewController  The calling `ORKTaskViewController` instance.
+ @param step                The `ORKStepViewController` that has just finished.
+ */
+- (void)taskViewController:(ORKTaskViewController *)taskViewController stepViewControllerWillDisappear:(ORKStepViewController *)stepViewController;
+
+/**
  Tells the delegate that the result has substantively changed.
  
  The task view controller calls this method when steps start or finish, or if an answer has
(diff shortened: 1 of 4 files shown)


ResearchKit/Common/ORKTaskViewController.h
```diff
@@ -212,6 +212,16 @@ task view controller and pass that data to `initWithTask:restorationData:` when
 - (void)taskViewController:(ORKTaskViewController *)taskViewController stepViewControllerWillAppear:(ORKStepViewController *)stepViewController;
 
 /**
+ Tells the delegate that a step has finished and will disappear.
+ 
+ The task view controller calls this method after a stepViewController has finished.
```
YuanZhu-apple: Update the comment, the step is not always finished.

ResearchKit/Common/ORKTaskViewController.h
```diff
@@ -212,6 +212,16 @@ task view controller and pass that data to `initWithTask:restorationData:` when
...
+ 
+ @param taskViewController  The calling `ORKTaskViewController` instance.
+ @param step                The `ORKStepViewController` that has just finished.
```
YuanZhu-apple: Update the comment, the step is not always finished.

ResearchKit/Common/ORKTaskViewController.h
```diff
@@ -212,6 +212,16 @@ task view controller and pass that data to `initWithTask:restorationData:` when
 - (void)taskViewController:(ORKTaskViewController *)taskViewController stepViewControllerWillAppear:(ORKStepViewController *)stepViewController;
 
 /**
+ Tells the delegate that a step has finished and will disappear.
```
YuanZhu-apple: Not sure if `finished` the right term here.

syoung-smallwisdom:
@YuanZhu-apple I've update the method signature to include direction and updated the comment to describe at what point in the flow this method will be called.

YuanZhu-apple:
@syoung-smallwisdom This looks great! 
Can you build a test case in ORKTest app to implement this delegate method and throw alert on step disappear?

syoung-smallwisdom:
@YuanZhu-apple Unit test added.

syoung-smallwisdom:
@YuanZhu-apple Added example to ORKTest

YuanZhu-apple:
@syoung-smallwisdom Thanks!
r=me.

PR #740, opened 2016-06-24
Add finishedSpokenInstruction to active step

syoung-smallwisdom:
The finished spoken instruction property allows a step to announce that it is completed. This is useful in the case where the phone screen is not visible such as a walking activity (where the phone is in the participant's phone) or a hand tremor measurement (where the screen is not facing the participant).

I have pulled it out into it's own PR in an effort to make the addition of our tremor, two-handed tapping and walking modules able to be submitted independently of one another.


syoung-smallwisdom:
This commit is now included in PR https://github.com/ResearchKit/ResearchKit/pull/747
Closing.

PR #746, opened 2016-06-28
[ORKTimedWalkStepViewController] Add Turn Around step

vtourraine:
_This PR is built upon #606_

The current Timed Walk active task consists of 2 active steps: walking up, and walking back to the starting point. We’ve started testing our app for multiple sclerosis patients including this task, and it seems like we should tweak that task to include a dedicated “turn around” step between the 2 existing steps.

The act of “turning around” is easy to overlook for people with good health, but for multiple sclerosis (and other diseases affecting motion capabilities) patients, this can be challenging to perform.

So this pull-requests adds this new step, as an instance of `ORKTimedWalkStep`. It also adds a new `turnAroundTimeLimit` parameter to create this task (with default value, similar to the existing `timeLimit` parameter for walking up and back).

This addition requires a new localized label, and changes the existing label for the “walking back” step.


ResearchKit/Common/ORKOrderedTask.h
```diff
@@ -475,6 +475,7 @@ typedef NS_OPTIONS(NSUInteger, ORKPredefinedTaskOption) {
                                       localized text is displayed.
  @param distanceInMeters            The timed walk distance in meters.
  @param timeLimit                   The time limit to complete the trials.
+ @param turnAroundTimeLimit         The time limit to complete the turn around step.
```
syoung-smallwisdom: Please *add* a method rather than replacing an existing one. This task module is used by other apps and this is a breaking change.

YuanZhu-apple: @syoung-smallwisdom Or there should be a way to exclude the turnAroundStep?

syoung-smallwisdom: @vtourraine , @YuanZhu-apple 
Yes, should also include allowing the turn-around step to be skipped. I hadn't realized that there wasn't a check for `turnAroundTimeLimit == 0`. Additionally, original signature should be included. Deprecated, maybe, but apps that use this module should not break.

syoung-smallwisdom:
@vtourraine , @YuanZhu-apple 
Because this *changes* the steps of an existing module, there needs to be reverse-compatibility for apps that already implement this task by using the results from the current steps. Existing research apps would be corrupted and lose data because of the insertion of this step (turning around) as a separate result.

Note: the apps that Sage Bionetworks currently supports do not use this module (we use a variation on the short walking test) and I am not sure who does, but it is worth considering that this is a breaking change to any apps that *do* use this module.

vtourraine:
So we need to choose between:
- Option A: Add a new method, and keep the original method
- Option B: Add a new parameter "include turn around step" to the existing method

For the original `timeLimit` parameter, passing `0` means no time limit. I kept the same behavior for the `turnAroundTimeLimit`, therefore choosing to include the turn around step or not needs to be yet another parameter, IMO.

YuanZhu-apple:
@syoung-smallwisdom I vote for B.

syoung-smallwisdom:
@vtourraine @YuanZhu-apple 
I think B makes more sense though I'd vote for C (smile)
Option C: deprecate the old method for release 1.4 and delete in release 1.5
Either way, having the conversation and having a simple means for existing apps to support the new method without having to change their result schemas is important. Whether or not to deprecate the old method is a nice-to-have.

umerkhan-apple:
@vtourraine I think we should proceed with including a parameter for `include turn around step`. If in the future, we get more feedback that this is ONLY how this task is used, then we'll deprecate the older method and migrate to this new one.

vtourraine:
I apologize for this super-late reply, but here are the requested changes:

- Merged changes from master to resolve pending conflicts
- Added distinct method to create the activity with a turn around step, while keeping the original method
- Added the new activity in the Catalog app

umerkhan-apple:
@vtourraine Could you resolve the merge conflicts? This PR looks ready to be merged.

vtourraine:
@umerkhan-apple Conflicts resolved. 🕊

rsanchezsaez:
This, as merged, has quite bad code duplication.

IMO, this function
```
+ (ORKOrderedTask *)timedWalkTaskWithIdentifier:(NSString *)identifier
                         intendedUseDescription:(nullable NSString *)intendedUseDescription
                               distanceInMeters:(double)distanceInMeters
                                      timeLimit:(NSTimeInterval)timeLimit
                            turnAroundTimeLimit:(NSTimeInterval)turnAroundTimeLimit
                     includeAssistiveDeviceForm:(BOOL)includeAssistiveDeviceForm
                                        options:(ORKPredefinedTaskOption)options;
```

should have the extra `includeTurnAroundStep` parameter, and then the old method should just call the extended method setting `includeTurnAroundStep` to `NO`.

I think that's what @umerkhan-apple was suggesting on his comment.

PR #753, opened 2016-07-07
Fix cast warning introduced with ORKSignatureResult

syoung-smallwisdom:
This was casting to ORKSignatureResult without first checking the class. Produced a warning but not an error because *currently* the assumption is valid.


YuanZhu-apple:
r=me.

PR #757, opened 2016-07-11
Use generic placeholders for email, first name, and last name in ORKRegistration and ORKLogin

ninoguba:
Currently the following placeholders are used for the email, first name, and last name fields in ORKRegistration and ORKLogin:

Email - "jappleseed@apple.com"
First Name - "John"
Last Name - "Appleseed"

On actual apps, these will often have to modified.  I propose to change them to the following generic placeholders consistent with the rest of the fields.

Email - "Enter email"
First Name - "Enter first name"
Last Name - "Enter last name"


YuanZhu-apple:
@ninoguba 
1. Definitely want to avoid @apple.com 
2. Not sure how difficult to localized this to other languages...

ninoguba:
Yeah it may be difficult now since there are several localizations already.

PR #758, opened 2016-07-13
Add ORKPageStep and ORKPageStepViewController

syoung-smallwisdom:
The ORKPageStep can be used to group a set of steps together and return a single flattened ORKStepResult for a group of steps. It uses a page step view controller to present the steps.

We will be using this to create a custom consent review step that is appropriate to our consent flow. Additionally, I will be using it to create step results that require a consolidated step result in response to multiple input steps.


md0u80c9:
Hi Shannon,

This looks really useful. Not sure if you saw my proposal for a way to implement tasks which can produce HL7 CCD documents but this fits in really well to that.

Have you considered calling it an ORKSectionStep rather than a Page step? I'm just thinking it probably more accurately describes what it's doing - dividing ORKTask into Sections. It would also fit if we could eventually go down the producing printing and CCD documents more directly from the task object because you would cluster each group of ORKSteps into sections.

syoung-smallwisdom:
@md0u80c9

I didn't see that. In our own code we call it a "subtask step" but for submission to ResearchKit, I simplified it so that it would be easier to grok. I went with "page step" for ResearchKit b/c that's what it is - a data source for a UIPageViewController - and I'm an iOS dev (smile). That said, having now googled "HL7 CCD", the term section does make sense though this implementation is more generic. My concern would be that the term "section" is already used for consent and there, a "section" is a single view (with a learn more) whereas this is a subgrouping of views.

It's also worth keeping in mind that I kept this version intentionally simplified and it does not implement all the features of `ORKTaskViewController`. I felt it was flexible enough to allow (via override) subclass implementations that include more functionality without risking obfuscation b/c of trying to do too much (lighting candles with a blow torch).

md0u80c9:
Yeah that makes sense.

Does ORKPageStep still support the use of predicates? Or more importantly is it possible for this to be added later from the design by extension of either ORKPageStep or the ORKResultPredicate functions? I'm asking because by the sounds of it your use case wouldn't necessarily require it, but other use cases (complex consents, or use of PageStep as a 'non-information grouping tool as per my original reply) might require predicates to be used and clearly it would be preferable if that were achievable without an API redesign later.

If you set up a predicate, let's say you set one to step from a page within ORKPageStep to one outside ORKPageStep - would that work safely? I was looking at it over lunch (not managed to test with any real code); but I wondered whether the predicate a) would work at all because you normally specify stepIdentifier, which is now nested, and b) if you set a predicate which went from inside an ORKPageStep to one which went outside ORKPageStep whether you'd break the restoreState function and end up with a load of problems. I may be completely wrong though - as I said I've not managed to check this with a piece of test code.

Best wishes,

Andrew

syoung-smallwisdom:
@md0u80c9 

I've changed the implementation to use the one that we are using in our projects. It's a little more obtuse (which is why I had simplified it initially) and it does *not* support all the same functionality of `ORKTaskViewController` but it can take an `ORKNavigableOrderedTask` as it's navigation control and thus handle predicates.

WRT jumping from a substep inside to a step outside, you would need to set the predicate to recognize that the substep result was flattened into it's parent step result. If you run the example in ORKTest and look at the debug output, you will see what I mean.

syoung-smallwisdom:
Note: If you want a subtask step to exit early (then just set the "jump to" step to "exit" or "Null" or whatnot. This will return a `nil` step which is what triggers the page step view controller to `goForward`.

md0u80c9:
Brilliant - thanks very much for that! The predicate stuff is really helpful for building quite rich interfaces with (hence my work this week at trying to a) simplify the interfaces for Swift and b) build comprehensive testing for them into ORKTest.

ResearchKit/Common/ORKPageStep.h
```diff
@@ -0,0 +1,132 @@
...
+ The page view controller calls this method to determine the step to display after the specified step. 
+ The page view controller can also call this method every time the result updates, to determine if the 
+ new result changes which steps are available.
+ 
+ @param identifier      The reference step identifier. Pass `nil` to specify the first step.
+ @param result          A snapshot of the current set of results.
+ 
+ @return The step that comes after the specified step, or `nil` if there isn't one.
+ */
+- (nullable ORKStep *)stepAfterStepWithIdentifier:(nullable NSString *)identifier withResult:(ORKTaskResult *)result;
```
YuanZhu-apple: Why not let `page view controller` call pageTask.stepAfterStepWithIdentifier, since these functions are already in ORKTask. And ask developer to override at here can lead to confusion.

ResearchKit/Common/ORKPageStep.h
```diff
@@ -0,0 +1,132 @@
...
+ @param aDecoder    The coder from which to initialize the ordered task.
+ 
+ @return An initialized ordered task.
+ */
+- (instancetype)initWithCoder:(NSCoder *)aDecoder NS_DESIGNATED_INITIALIZER;
+
+/**
+ The subtask used to determine the next/previous steps that are in this grouping
+ */
+@property (nonatomic, copy, readonly) ORKOrderedTask *pageTask;
```
YuanZhu-apple: `id<ORKTask>` to be more flexible?

syoung-smallwisdom: To simplify the encoding so that it would work for ORKTest serialization.  Initially, I was using `id <ORKTask, NSCopying, NSSecureCoding>` but I couldn't figure out how to get the ORKTest serialization tests to work correctly.

ResearchKit/Common/ORKPageStep.h
syoung-smallwisdom: @YuanZhu-apple 
Because for our use-cases we aren't always using the pageTask directly. The original implementation didn't use a task at all and just used a set of steps. However, one of the other users of ResearchKit was asking about how to implement this for navigable surveys and so I changed the implementation to work for that case directly instead of hiding that implementation detail.

YuanZhu-apple: But I see now the steps are always converted to a task, so I don't think it is necessary to have these here.

syoung-smallwisdom: @YuanZhu-apple 
This is an implementation that we are currently using in our code. The original base class implementation did not use an `ORKOrderedTask` and relied upon the methods overriding the `stepBeforeStep:result:` and `stepAfterStep:result` methods instead. Because overriding and implementing custom navigation logic was more confusing for the purposes for which @md0u80c9 wanted to use the class, I changed it to allow for initializing with an `ORKOrderedTask`.

I've gone back to using the prior implementation's public methods and instead have a subclass called `ORKPageTaskStep` that publicly exposes the pageTask property and initializer.

syoung-smallwisdom:
@umerkhan-apple 
Ping. This PR has been out since July and we are using it in our apps by pointing to our own fork. It would be nice to bring our fork a little closer to following ResearchKit/master.

umerkhan-apple:
@syoung-smallwisdom Thanks for this contribution! 👍

PR #759, opened 2016-07-13
ork_answer_format_property

dephillipsmichael:
I am one of the primary developers for the Football Player's Health Study App.  We had a bug filed that user's could enter any blank string, like "    ", as their name in the consent review flow, and it would allow them to continue.  

To fix it, I added a new property to ORKAnswerFormat that allows the consent review step vc to make any "blank" string be considered invalid input.


YuanZhu-apple:
@dephillipsmichael Do you think this can be done via `validationRegex`?

dephillipsmichael:
@YuanZhu-apple I was able to accomplish it with validationRegex.  Thanks for the tip!  Closing this PR...

PR #762, opened 2016-07-15
bad_access_crash_fix

dephillipsmichael:
Our QA team found this crash in Football Player's Health Study when a user would cancel a balance or walking activity right after it shows step 2 of 7.  In our case, it is a 5 second countdown timer that triggers the showViewController method in pageViewController.  However the pageViewController had been deallocated, and the strongSelf was trying to access the iVar in a block, which was causing a exc_bad_access crash.

I changed the strongSelf to weakSelf in the block, and switch that iVar to a property so that it could properly access from weakSelf.

### DIFF
--- a/ResearchKit/Common/ORKTaskViewController.m
+++ b/ResearchKit/Common/ORKTaskViewController.m
@@ -166,7 +166,7 @@ static void *_ORKViewControllerToolbarObserverContext = &_ORKViewControllerToolb
     NSMutableArray *_managedStepIdentifiers;
     ORKViewControllerToolbarObserver *_stepViewControllerObserver;
     ORKScrollViewObserver *_scrollViewObserver;
-    BOOL _hasSetProgressLabel;
+    
     BOOL _hasBeenPresented;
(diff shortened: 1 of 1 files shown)


ResearchKit/Common/ORKTaskViewController.m
```diff
@@ -946,19 +949,18 @@ - (void)showViewController:(ORKStepViewController *)viewController goForward:(BO
     
     ORKWeakTypeOf(self) weakSelf = self;
     [self.pageViewController setViewControllers:@[viewController] direction:direction animated:animated completion:^(BOOL finished) {
-        ORKStrongTypeOf(weakSelf) strongSelf = weakSelf;
```
syoung-smallwisdom: Did you test leaving this in and instead adding a check with an early exit if `weakSelf == nil`. By strongifying self, the code shouldn't dealloc *during* this block.

dephillipsmichael: That's a good thought. I'll try that solution and see if the crash still exists.  It would be a much smaller change.

PR #764, opened 2016-07-18
bad_access_fix

dephillipsmichael:
Our QA team found an EXC_BAD_ACCESS crash on strongSelf of this block callback.  It is caused by quickly canceling a task while it is transitioning.

To fix it, I added a check if weakSelf is nil so that we can avoid the strong self bad access crash.


YuanZhu-apple:
@dephillipsmichael Thanks!

Issue #766, opened 2016-07-26
ORKGraphChartView crashes without datasource


p2:
It may happen that an `ORKGraphChartView` sees its first layout before a data source has been assigned. The [`numberOfPlots`](https://github.com/ResearchKit/ResearchKit/blob/master/ResearchKit/Charts/ORKGraphChartView.m#L582) method returns **1** by default, when it's called in the `layoutSubviews` » `updateYAxisPoints` methods this leads to a crash.

The default `numberOfPlots` should be **0** to avoid this crash.

This is somewhat related to #691 .

p2:
Following up, documentation of `- (NSInteger)numberOfPlotsInGraphChartView:(ORKGraphChartView *)graphChartView` says that **1 plot** is assumed if that method is not implemented. So to prevent this crash, the internal methods could do index range checking before accessing array members.

umerkhan-apple:
Fix merged!

Issue #691, opened 2016-05-27
Pie chart layout layers crash


sukh-cohaesus:
Hello

I'm having a weird crash with the Pie chart data source. The error is

`*** Terminating app due to uncaught exception 'NSRangeException', reason: '*** -[__NSArrayM objectAtIndex:]: index 0 beyond bounds for empty array'`

Which is being thrown within at

`CAShapeLayer *segmentLayer = _segmentLayers[idx];`

within this method of `ORKPieChartPieView` class

```
- (void)layoutPieChartLayers {
    NSInteger numberOfSegments = [_parentPieChartView.dataSource numberOfSegmentsInPieChartView:_parentPieChartView];
    for (NSInteger idx = 0; idx < numberOfSegments; idx++) {
        CAShapeLayer *segmentLayer = _segmentLayers[idx];
        segmentLayer.frame = _circleLayer.bounds;
        segmentLayer.path = _circleLayer.path;
        segmentLayer.lineWidth = _circleLayer.lineWidth;
    }
}
```

Here's my code

```
    class PieChartDataSource: NSObject, ORKPieChartViewDataSource {

        private let pieColors = [
            UIColor.greenColor(),
            UIColor.grayColor(),
        ]
        private var pieValues = [Double]()

        override init() {
            super.init()
            
            HealthKitManager.sharedInstance.getHealthAuthorization() { succeeded, error in
                guard succeeded else {
                    print(error!.description)
                    return
                }
                
                self.getData()
            }
            
        }
        
        func getData() {
            
            HealthKitManager.sharedInstance.queryDailySteps() { result, error in
                
                let max: Double = 100 // 100%
                let steps: Double = (max / 10000) * result!
                let remaining = steps < max ? max - steps : 0
                if steps != 0 {
                    self.pieValues = [steps, remaining]
                } else {
                    self.pieValues = [remaining]
                }
            }
        }
        
        func numberOfSegmentsInPieChartView(pieChartView: ORKPieChartView ) -> Int {
            return pieValues.count
        }
        
        func pieChartView(pieChartView: ORKPieChartView, valueForSegmentAtIndex index: Int) -> CGFloat {
            if pieValues.count > index {
                return CGFloat(pieValues[index])
            } else {
                return 0
            }
        }
        
        func pieChartView(pieChartView: ORKPieChartView, colorForSegmentAtIndex index: Int) -> UIColor {
            return pieColors[index]
        }
    }
```

Any thoughts?

[...]

PR #770, opened 2016-07-29
[Charts] Fix crashes

rsanchezsaez:
This fixes crashes https://github.com/ResearchKit/ResearchKit/issues/691 and https://github.com/ResearchKit/ResearchKit/issues/766:
- `ORKGraphChartView` was crashing if laid out before the `dataSource` was set. The fix makes `-numberOfPlotsInGraphChartView` non-optional, and makes it to return 0 plots by default.
- `ORKPieChartView` was crashing if its `dataSource` returned a `0` value for any segment. Now segments with a `0` value are supported. The corresponding item will be added to the *legend*, and the percentage label will read `0 %`.

Big thanks to @p2 for discovering and/or investigating both issues.


YuanZhu-apple:
r=me!

PR #775, opened 2016-08-01
Piechart add scale factor support 

YuanZhu-apple:
With `scaleFactor` = `0.4`
screen shot 2016-08-01 at 2 37 04 pm 

With `scaleFactor` = `0.6`
screen shot 2016-08-01 at 2 36 27 pm


YuanZhu-apple:
r=me

PR #778, opened 2016-08-01
Displaying a color gradient above the scale slider

YuanZhu-apple:
For example, this can be used to show a scale from bad to good by setting.

screen_shot_2016color2
screen_shot_2016color1

### DIFF
--- a/ResearchKit/Common/ORKAnswerFormat.h
+++ b/ResearchKit/Common/ORKAnswerFormat.h
@@ -432,6 +432,26 @@ ORK_CLASS_AVAILABLE
  */
 @property (strong, nullable) UIImage *minimumImage;
 
+/**
+ The colors to use when drawing a color gradient above the slider. Colors are drawn such that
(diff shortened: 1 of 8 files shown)


md0u80c9:
This looks very nice! Haven't had chance to look at the code yet, but I can see this being really helpful.

md0u80c9:
A nice real-world example use case of this would be disability scales, such as the modified Rankin Scale (mRS); which is a measure of functional disability. The scale goes from 0 (unaffected); through to 2 (a disability which has an effect but the individual can maintain all activities of daily living) through to 5 (severely disabled and fully dependent), or 6 (dead). Anaesthetic scales are similarly 'non-linear' numeric scales.

Most clinical research trials consider 0-2 a 'good outcome' of treatment e.g. for stroke care (ie. makes a near-full recovery); others consider 0-3 a 'good outcome' (ie. a good recovery to independent living, give or take some support). Using the colour bar supports not just what the mRS is, but what the trial is considering a 'good outcome'.

YuanZhu-apple:
r=self

Issue #785, opened 2016-08-11
"Timer" and "Distance Travelled" Labels in ORKFitnessStep not Appearing


phantomkirby:
I opened ORKCatalog and looked at the Fitness Active Task and the timer and pedometer labels are not showing up.

Yet, when you look at the documentation: [here](http://researchkit.org/docs/docs/ActiveTasks/ActiveTasks.html), you can see that step 4 has the timer and distance travelling showing.

To be sure, I actually created my own ORKFitnessStep as well, using this code:

```
func IfitnessActiveStep() -> ORKStep {
        let step: ORKFitnessStep = ORKFitnessStep(identifier: ORKFitnessWalkStepIdentifier)
        step.stepDuration = 10 * 60
        step.title = "Title"
        step.text = "text text"
        step.image = UIImage(named: "imageExample")
        step.spokenInstruction = "Instruction. text to speech"
        step.optional = false;
        
        step.shouldShowDefaultTimer = true
        
        step.shouldUseNextAsSkipButton = true
        step.shouldStartTimerAutomatically = true
        step.shouldSpeakRemainingTimeAtHalfway = true
        
        step.shouldPlaySoundOnStart = true
        step.shouldVibrateOnStart = true
        
        step.shouldSpeakCountDown = true
        
        step.shouldVibrateOnFinish = true
        step.shouldPlaySoundOnFinish = true
        step.shouldContinueOnFinish = true
        
        var recorderConfigurations: [ORKRecorderConfiguration] = []
        recorderConfigurations.append(ORKPedometerRecorderConfiguration(identifier: ORKPedometerRecorderIdentifier))
        recorderConfigurations.append(ORKHealthQuantityTypeRecorderConfiguration(identifier: ORKHeartRateRecorderIdentifier, healthQuantityType: HKQuantityType.quantityTypeForIdentifier(HKQuantityTypeIdentifierHeartRate)!, unit: HKUnit.countUnit().unitDividedByUnit(HKUnit.minuteUnit())))
        recorderConfigurations.append(ORKAccelerometerRecorderConfiguration(identifier: ORKAccelerometerRecorderIdentifier, frequency: 100))
        recorderConfigurations.append(ORKDeviceMotionRecorderConfiguration(identifier: ORKDeviceMotionRecorderIdentifier, frequency: 100))
        step.recorderConfigurations = recorderConfigurations;
        return step
    }
```

Neither of the two labels show up still.

YuanZhu-apple:
merged.

PR #788, opened 2016-08-12
Fixing bug where timerLabel did not display in ORKFitnessStep

phantomkirby:
in the "updateTimerLabel" function, the timerLabel was always hidden because the labelString was null. The labelString was null every time because the "formatter" object (an NSDateComponentsFormatter object) was null the whole time. This object was null because inside the "dispatch_once", we did not reference/edit the existing static "formatter" object outside of the dispatch_once, but rather, we instantiated a new NSDateComponentsFormatter object.

This seems to work for me. I tested on iOS 6 simulator. I am on iOS 9/Xcode 7.

Edit: This pull request is related to the issue I made a day ago: #785


YuanZhu-apple:
@phantomkirby Thanks!

PR #803, opened 2016-08-25
Make ORKInstructionStepViewController and ORKCompletionStepViewController public

syoung-smallwisdom:
We use a custom subclass of ORKInstructionStep which uses a custom implementation for learn more actions. By publicly exposing these view controllers, we can return the appropriate view controller (or subclass).


YuanZhu-apple:
OK, r=me.

YuanZhu-apple:
@syoung-smallwisdom Is it necessary to move them from private header?

syoung-smallwisdom:
@YuanZhu-apple 
The rules change from time to time, but last time I checked, if a class is subclassed that isn't available in the public SDK then the build cannot archive.

PR #804, opened 2016-08-26
There is a copyright that should be included with the memory game.

syoung-smallwisdom:
We are including this copyright in mPower elsewhere in the app, but since this is a part of ResearchKit, the copyright should be shown as a part of the activity.


YuanZhu-apple:
I will discuss this with my colleague.

syoung-smallwisdom:
thanks!

Issue #774, opened 2016-08-01
Layout issue for inputs in ORKTaskView


Geosef:
My team and I noticed issues with the layout of ORKTasks within ORKTaskViews in both sliders and text inputs. We have not tried other inputs to verify if they are affected or not.
<hr />
This is text input prior to being tapped in the field:
![screen shot 2016-08-01 at 3 02 27 pm](https://cloud.githubusercontent.com/assets/5271942/17307128/27b84fec-57f9-11e6-8134-13b2d8ae75cb.png)
<hr />
Once tapped for input, the field and title both shrink: 
![screen shot 2016-08-01 at 3 02 37 pm](https://cloud.githubusercontent.com/assets/5271942/17307127/27b77c0c-57f9-11e6-8910-44523bc42f95.png)
The placeholder text is cut off already.
<hr />
This is after "12" has been entered:
![screen shot 2016-08-01 at 3 02 49 pm](https://cloud.githubusercontent.com/assets/5271942/17307126/27b4df1a-57f9-11e6-8dab-e51d348035b9.png)
The field is so small you cannot see any input and the title is also cut off.
<hr />
And after a medium length string of numbers:
![screen shot 2016-08-01 at 3 02 55 pm](https://cloud.githubusercontent.com/assets/5271942/17307129/27b8edb2-57f9-11e6-9368-81bed5a8cf8a.png)
The title has reappeared, but the first few characters of the input are still cut off.
<hr />
This is a slider with very short descriptions for min/max:
![screen shot 2016-08-01 at 3 08 56 pm](https://cloud.githubusercontent.com/assets/5271942/17307306/09a608cc-57fa-11e6-82ed-74c3ab777f22.png)
<hr />
This is the same slider after being tapped for input:
![screen shot 2016-08-01 at 3 09 10 pm](https://cloud.githubusercontent.com/assets/5271942/17307327/213ce744-57fa-11e6-94a2-ce6e94a61bd8.png)
The slider is so small you can't change the value.
<hr />
Finally, a slider with really long descriptions:
![screen shot 2016-08-01 at 3 09 24 pm](https://cloud.githubusercontent.com/assets/5271942/17307356/48d534fa-57fa-11e6-89ca-7632cebb0538.png)
<hr />

We believe a minimum width should be in place to handle text input and short descriptions on sliders. It may also be worthwhile looking into other inputs and seeing how they are affected.

YuanZhu-apple:
[...]

PR #806, opened 2016-08-30
Fixed width constraint priority

Geosef:
Added fix for issue #774


YuanZhu-apple:
@Geosef Thanks!
r=me!

PR #818, opened 2016-09-08
Use NSPredicate instead of Predicate

asynchrony-ringo:
NSPredicate was a thing. Then in Swift 3.0 beta something it became Predicate instead. Then in beta 6 it went back to NSPredicate. Predicate thus is no longer a thing.


rsanchezsaez:
Thanks for the change @asynchrony-ringo. However, this has already been done on a different PR, so this PR can be closed. ;-)

akshay-yadav-apple:
@asynchrony-ringo  This issue has been fixed in the latest master, so closing this PR.

PR #825, opened 2016-09-16
Always set a date for non-nil ORKResult startDate

syoung-smallwisdom:
This is a non nil property and should not be set to a possibly nil value. Doing so can cause a crash in Swift 3.0 override of `-(ORKStepResult)result` because of a violated assumption.


kelseydedoshka-apple:
@syoung-smallwisdom thanks for this fix.

@umerkhan-apple take a look.

PR #828, opened 2016-09-21
Fix navigation looping to return unique results for each step.

syoung-smallwisdom:
Previous to this commit, `ORKTaskViewController` only tracked the _last_ result, inserting that result into the returned array of results for each instance of the step result with that identifier. This commit fixes that bug so that an `ORKNavigableOrderedTask` with a looping navigation will return the unique result for each loop.

Repro Steps:
1. Run the "Navigation Loop Task" in ORKTest
2. Loop through the answers 

Expected:
The results will show a different value for steps with the same identifier that are in the loop.

Actual:
All steps are replaced with the last result.

### DIFF
--- a/ResearchKit/Common/ORKTaskViewController.m
+++ b/ResearchKit/Common/ORKTaskViewController.m
@@ -695,30 +695,49 @@ static NSString *const _ChildNavigationControllerRestorationKey = @"childNavigat
 - (NSArray *)managedResults {
     NSMutableArray *results = [NSMutableArray new];
     
-    [_managedStepIdentifiers enumerateObjectsUsingBlock:^(id obj, NSUInteger idx, BOOL *stop) {
-        NSString *identifier = obj;
-        ORKResult *result = _managedResults[identifier];
-        NSAssert(result, @"Result should not be nil for identifier %@", identifier);
+    [_managedStepIdentifiers enumerateObjectsUsingBlock:^(NSString *identifier, NSUInteger idx, BOOL *stop) {
+        id <NSCopying> key = [self uniqueManagedKey:identifier index:idx];
+        ORKResult *result = _managedResults[key];
+        NSAssert2(result, @"Result should not be nil for identifier %@ with key %@", identifier, key);
         [results addObject:result];
     }];
     
     return [results copy];
 }
 
-- (void)setManagedResult:(id)result forKey:(id <NSCopying>)aKey {
+- (void)setManagedResult:(ORKStepResult *)result forKey:(NSString *)aKey {
     if (aKey == nil) {
         return;
     }
     
-    if (result == nil || NO == [result isKindOfClass:[ORKResult class]]) {
+    if (result == nil || NO == [result isKindOfClass:[ORKStepResult class]]) {
         @throw [NSException exceptionWithName:NSGenericException reason:[NSString stringWithFormat: @"Expect result object to be ORKResult type and not nil: {%@ : %@}", aKey, result] userInfo:nil];
         return;
     }
+    
+    // Manage last result tracking (used in predicate navigation)
+    // If the previous result and the replacement result result are the same result then `isPreviousResult`
+    // will be set to `NO` otherwise it will be marked with `YES`.
+    ORKStepResult *previousResult = _managedResults[aKey];
+    previousResult.isPreviousResult = YES;
+    result.isPreviousResult = NO;
     
     if (_managedResults == nil) {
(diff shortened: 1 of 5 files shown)


zigdanis:
I would also like to have looped survey. Can you also make it to reset default Answer to nil while iterating on each new survey loop?

zigdanis:
Also I getting: 

> [ResearchKit][Warning] -[ORKNavigableOrderedTask stepAfterStep:withResult:] Index of next step ("loopAStep") is equal or lower than index of current step ("loopBStep") in ordered task. Make sure this is intentional as you could loop idefinitely without appropriate navigation rules. Also please note that you'll get duplicate result entries each time you loop over the same step.

While I'm testing this PR in ORKTest app

syoung-smallwisdom:
@zigdanis 
The warning is not new to this PR. It is there as a hint to developers to expect multiple instances of `ORKStepResult` with the same `identifier`. This PR is intended to address an issue where the instances of `ORKStepResult` were all pointing to the same instance rather than creating a unique instance for the previous results.  I will look into reseting the default answer.

md0u80c9:
Hi,

Yes the warning is deliberate because if you have multiple steps with the same identifier in your ORKTask, your ORKTaskResult will have multiple results with the same identifier - which can lead to a number of issues if you then try to use methods to refer to a result by identifier (the methods all expect a single result).

Any looped questions must ensure they return unique identifiers if you expect to be able to use the results from that survey question meaningfully.

syoung-smallwisdom:
@zigdanis 
I just checked and showing the previous result (if available) has *not* been changed by this pull request. All that this pull request does is to fix a bug and return *all* results uniquely. Additionally, it fixes the predicate used in navigation looping and fixes `resultForIdentifier` to use reverse enumeration.

https://github.com/ResearchKit/ResearchKit/pull/854 would address the issue you have requested. Since that is a change request (rather than a bug fix), I would like to keep it separate from this pull request.

@md0u80c9 
Correct. This does not change the previously implemented behavior of having duplicate pointers to results with the same identifier. In other words, it cannot be assumed that a valid result set will not include multiple results with the same identifier. (Hence, the warning)

ResearchKit/Common/ORKTaskViewController.m
```diff
@@ -695,30 +695,49 @@ - (UIImageView *)findHairlineViewUnder:(UIView *)view {
...
 
-- (void)setManagedResult:(id)result forKey:(id <NSCopying>)aKey {
+- (void)setManagedResult:(ORKStepResult *)result forKey:(NSString *)aKey {
     if (aKey == nil) {
         return;
     }
     
-    if (result == nil || NO == [result isKindOfClass:[ORKResult class]]) {
+    if (result == nil || NO == [result isKindOfClass:[ORKStepResult class]]) {
         @throw [NSException exceptionWithName:NSGenericException reason:[NSString stringWithFormat: @"Expect result object to be ORKResult type and not nil: {%@ : %@}", aKey, result] userInfo:nil];
```
rsanchezsaez: Perhaps change the exception message to something like `'result' shouldn't be nil and should be of type 'ORKStepResult': {%@ : %@}`? :-)

ResearchKit/Common/ORKTaskViewController.m
```diff
@@ -695,30 +695,49 @@ - (UIImageView *)findHairlineViewUnder:(UIView *)view {
...
         return;
     }
     
+    // Manage last result tracking (used in predicate navigation)
+    // If the previous result and the replacement result result are the same result then `isPreviousResult`
```
rsanchezsaez: There's a typo in the comment: `result result`.

Testing/ORKTest/ORKTest/MainViewController.m
```diff
@@ -3768,6 +3769,14 @@ - (void)taskViewControllerDidComplete:(ORKTaskViewController *)taskViewControlle
     
     NSLog(@"[ORKTest] task results: %@", taskViewController.result);
     
+    // Validate the results
+    NSArray *results = taskViewController.result.results;
+    if (results) {
+        NSSet *uniqueResults = [NSSet setWithArray:results];
+        BOOL allResultsUnique = (results.count == uniqueResults.count);
+        NSAssert(allResultsUnique, @"The returns results have duplicates of the same object.");
```
rsanchezsaez: `The returned results...`. :)

umerkhan-apple:
Looks great! Thanks for this great contribution! 👍

Issue #820, opened 2016-09-12
Commit 74326bae Breaks the sizing of the completion checkmark


syoung-smallwisdom:
This commit has broken the checkmark that is used as the completion step for every active task.

![simulator screen shot sep 12 2016 1 52 35 pm](https://cloud.githubusercontent.com/assets/5649217/18452598/68fa7aa8-78f0-11e6-8a20-486cd6f50c70.png)

syoung-smallwisdom:
This commit was made to fix issue https://github.com/ResearchKit/ResearchKit/issues/774

kelseydedoshka-apple:
@syoung-smallwisdom this issue has been resolved!

PR #837, opened 2016-10-12
Add a flag to allow displaying the continue button in the standard position for the compltion step.

syoung-smallwisdom:
We are finding that many of our users are confused as to how to dismiss a task because the continue button is moved to the upper right corner for the last step.  This commit allows overriding the default behavoir of moving that button and places it in the standard position for a continue button.

Note: This commit does not address https://github.com/ResearchKit/ResearchKit/issues/820 and therefore the checkmark in the included screenshot is not sized correctly.

simulator screen shot oct 12 2016 11 27 42 am


akshay-yadav-apple:
@syoung-smallwisdom 
Thanks for the PR.
We will discuss this issue with other colleagues, and get back to you.

umerkhan-apple:
@syoung-smallwisdom Looks great! Thanks! 👍

PR #848, opened 2016-10-25
Add ORKStepModifier to navigable task.

syoung-smallwisdom:
`ORKStepModifier` was designed to allow optional modifications to an existing step in an  `ORKNavigableOrderedTask`. This allows changing the step based on navigation results. This class is used to fix a bug with the spoken instruction in the tremor task if the participant skips performing the task with one of their hands.

Repro steps:
1. Perform the tremor task. Set a breakpoint to check which hand is randomly selected as the first hand.
2. Select the second hand as the hand to skip.
3. Go through the tasks.

Expected:
The spoken instruction after the queen wave should say. "This concludes this activity." 

Actual:
The spoken instruction after the queen wave says. "Switch to X hand."


ResearchKit/Common/ORKStepNavigationRule.m
```diff
@@ -450,3 +451,110 @@ - (NSUInteger)hash {
...
+        _resultPredicate = resultPredicate;
+        _keyValueMap = keyValueMap;
+    }
+    return self;
+}
+
+- (void)modifyStep:(ORKStep *)step withTaskResult:(ORKTaskResult *)taskResult {
+    
+    // The predicate can either have:
+    // - an ORKResultPredicateTaskIdentifierVariableName variable which will be substituted by the ongoign task identifier;
```
rsanchezsaez: Typo in `ongoing`.

rsanchezsaez: Sorry, another typo in `substituionVariables` (-> `substitutionVariables`), my fault again!

rsanchezsaez:
This is very clever, but I wonder if it's over-complicating things a little bit, or making code a little unsafe? The `ORKStepModifier` could broke a step very badly?

The same could be achieved in a "simpler" way by just using two regular alternative steps and a navigation rule, no?

Anyway, if @YuanZhu-apple thinks this is worthwhile, I'm fine with it. The code looks good. The only thing: adding some unit tests would be great. Also, let's make sure that the serialization tests work as well (I'm not sure if they where even working before this).

syoung-smallwisdom:
@rsanchezsaez 
I thought about copying the step and adding a skip rule, but that seemed obfuscated when what I *really* want is just to have the spoken instruction changed to match the navigation. 

While yes, this will crash if the developer attempts to set a key for a property that isn't a KVO property, I would hope that devs who might use this feature would test their code. ;)

syoung-smallwisdom:
@rsanchezsaez 
Took a look at changing to use navigation to skip one or the other step and ran into an issue with the expected results. Because the two steps have to have unique identifiers, this would change the result set. The solution I am proposing would be reverse-compatible.

rsanchezsaez:
I was thinking further about this and it occurred to me that, perhaps, to make this slightly safer we could declare an `ORKModifiableStep` protocol which has a `- (NSArray<NSString *> *)modifiableKeys` method.

Then, `ORKStepModifier` could check that the step adopts the protocol and that the key to be changed is returned as modifiable by the step before trying to do the key-value setting. If any of those checks fail we could either: log an error (less crash-happy); or raise a more descriptive exception (instead of the more generic `NSUndefinedKeyException` which will raise if the modifier steps tries to change a non-existing key).

@syoung-smallwisdom Do you think this makes sense? Also, should we add some small unit tests to `ORKTaskTests.m`? :-)

syoung-smallwisdom:
@rsanchezsaez 
I've added a unit test for the new class.

`ORKStepModifier` is an abstract base class. `ORKKeyValueStepModifier` states in the comments that the `keyValueMap` is used to modify the step (if the predicate matches) using key-value coding. If it crashes, then it is because the implementation violates expectations of the `NSKeyValueCoding` protocol and it *should* crash in order to alert the developer that their code is failing to meet the expected assumptions. That said, I have added a try/catch so that it will assert if there an exception but will swallow the exception in production code.

rsanchezsaez:
What you say makes sense, your argument is convincing. I'm now torn between throwing the original `NSUndefinedKeyException` exception or leaving the assert as you implemented. :-)

ResearchKit/Common/ORKNavigableOrderedTask.m
```diff
@@ -118,6 +119,34 @@ - (void)removeSkipNavigationRuleForStepIdentifier:(NSString *)stepIdentifier {
...
+    ORKThrowInvalidArgumentExceptionIfNil(stepModifier);
+    ORKThrowInvalidArgumentExceptionIfNil(stepIdentifier);
+    
+    if (!_stepModifiers) {
+        _stepModifiers = [NSMutableDictionary new];
+    }
+    _stepModifiers[stepIdentifier] = stepModifier;
+}
+
+- (nullable ORKStepModifier *)stepModifierForStepIdentifier:(NSString *)stepIdentifier {
```
umerkhan-apple: remove nullable

ResearchKit/Common/ORKStepNavigationRule.h
```diff
@@ -320,4 +321,83 @@ ORK_CLASS_AVAILABLE
...
+/**
+ Modify the steps for a task.
+ 
+ @param step            The step that is associated with this modifier
+ @param taskResult      The current task result
+ */
+- (void)modifyStep:(ORKStep *)step withTaskResult:(ORKTaskResult *)taskResult;
+
+@end
+
```
umerkhan-apple: Add extra new line.

ResearchKit/Common/ORKStepNavigationRule.h
```diff
@@ -320,4 +321,83 @@ ORK_CLASS_AVAILABLE
...
+/**
+ The `ORKKeyValueStepModifier` class is an class for an object that can be used to modify a step
+ if a given navigation rule is matched.
+ 
+ Step modifiers can be used within an `ORKNavigableOrderedTask` object. You assign step modifiers
+ to be triggered after a task step is shown. Each step can have one step modifier at most.
+ 
+ */
```
umerkhan-apple: Remove extra line in comment.

syoung-smallwisdom:
@umerkhan-apple 
Code style changes implemented

umerkhan-apple:
@syoung-smallwisdom Thank you! 👍

PR #853, opened 2016-10-27
Fix typo in podspec version quoting characters.

ryanschneider:
Looks like when the pod spec version was bumped backticks were used instead of single quotes.


ryanschneider:
FWIW, the backticks were introduced in this commit: 160bb111f03a8d663ae5f3b0a75d01a78c3dbec4

rsanchezsaez:
Wrong username, but oh well. ;-)

umerkhan-apple:
@ryanschneider Could you also do a similar PR for CareKit podspec?

ryanschneider:
Yup will do.

PR #854, opened 2016-10-27
Allow optional check of default result source before using previous.

syoung-smallwisdom:
By request from another dev, this allows the developer to force checking the default source _before_ using the previous result. If an empty `ORKStepResult` is returned, then the previous result will be ignored.

### DIFF
--- a/ResearchKit/Common/ORKTaskViewController.m
+++ b/ResearchKit/Common/ORKTaskViewController.m
@@ -1074,13 +1074,9 @@ static NSString *const _ChildNavigationControllerRestorationKey = @"childNavigat
                 result = [self.defaultResultSource stepResultForStepIdentifier:step.identifier];
             }
             
-            // If nil, assign to the previous result
+            // If nil, assign to the previous result (if available) otherwise create new instance
             if (!result) {
-                result = previousResult;
-            }
-            
-            if (!result) {
-                result = [[ORKStepResult alloc] initWithIdentifier:step.identifier];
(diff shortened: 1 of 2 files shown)


ResearchKit/Common/ORKTaskViewController.m
```diff
@@ -1065,9 +1065,18 @@ - (ORKStepViewController *)viewControllerForStep:(ORKStep *)step {
...
+            if ((previousResult == nil) || alwaysCheckForDefaultResult) {
+                result = [self.defaultResultSource stepResultForStepIdentifier:step.identifier];
+            }
+            
+            // If nil, assign to the previous result
+            if (!result) {
+                result = previousResult;
             }
             
             if (!result) {
```
kelseydedoshka-apple: Please condense down the two (!result) sequential if statements to a more recognized paradigm. 

One possibility is something like this:

if (!result) {
	result = previousResult ? previousResult : [[ORKStepResult alloc] initWithIdentifier:step.identifier];
}

kelseydedoshka-apple:
@syoung-smallwisdom - this is looking great so far and thanks for the contribution. Please checkout the quick code change in the Files Changed comments.

syoung-smallwisdom:
@kelseydedoshka-apple 
Change has been made as requested.

kelseydedoshka-apple:
@syoung-smallwisdom thanks for the quick update!

@umerkhan-apple take a look.

Issue #839, opened 2016-10-13
ResearchKit.h framework header imports gone wrong


malhal:
This commit made a mistake in the umbrella header ResearchKit.h:

https://github.com/ResearchKit/ResearchKit/commit/f7d1591c19ad0c12a969f1dceeb8d41716079073

The imports in should be of the framework format:

```
#import <ResearchKit/ORKStep.h>
```

not

```
#import "ORKStep.h"
```

The commit states cocoa pods as the reason for the change, I'd be happy to advise on how to correct the cocoa pod configuration rather than break the header.

rsanchezsaez:
Hi @malhal 

I think you're right, by looking at the rest of the Apple's Frameworks, they all use the *framework format*. I'm ok with changing them to this format.

However, the current direct format seems to work fine too: I built ResearchKit as a standalone framework, added the framework to a new project (not including ResearchKit source code), and it compiled and ran fine. Just to get to the bottom of this issue, in which way do you mean that using this header format is broken?

Thanks!
Ricardo

malhal:
Hi @rsanchezsaez, yes the umbrella header ResearchKit.h should contain the framework format, as all public headers should contain the framework format, which it looks to me like most if not all already do. I'm not sure why you aren't encountering an error but right now if I include a pre-built version of the framework in a test project, or embed the ResearchKit project (with an appropriate include header dir build setting), either way when I do:

    #import <ResearchKit/ResearchKit.h>

I get the compiler error:
        `/Users/mh/Development/Test/ResearchTest/dependencies/ResearchKit/ResearchKit/ResearchKit.h:32:9: 'ORKTypes.h' file not found`

Which is accurate because it is importing "ORKTypes.h" and not <ResearchKit/ORKTypes.h>. 

How are you including Research Kit in your example project?

rsanchezsaez:
Hey @malhal,

Sorry for my again late reply, I'm a bit busy. Still trying to get to the bottom of this.

For me it works with the (not completely correct) current headers. By manually adding a precompiled *ResearchKit* library to my project, leaving all the project settings by default, it compiles and runs:

<img width="1519" alt="screen shot 2016-10-29 at 9 01 19 pm" src="https://cloud.githubusercontent.com/assets/444313/19834411/ec1c5358-9e1a-11e6-98ae-6108a68fe423.png">

[...]

PR #859, opened 2016-11-08
[Umbrella headers] Update #imports to framework format

rsanchezsaez:
Update *umbrella header* `#import` statements to the more correct *Framework format*. Thanks to @malhal for suggesting this on [Issue #839](https://github.com/ResearchKit/ResearchKit/issues/839).

This format mirrors the rest of *Apple's iOS Framework* headers.

---

@malhal Could you confirm that a precompiled framework continues to work for you after this change?
@vtourraine Could you re-check that after the merge this branch still works through Cocoapods, and that it correctly imports the swift files?

---

This change also happily resolves the (harmless) problem explained in my second comment in [PR #648](https://github.com/ResearchKit/ResearchKit/pull/648):

> I think there's a bug in the Xcode build chain because, even though all Unit Tests build and pass fine from a clean state, an *Unable to build module ResearchKit* error appears after the tests successfully finish. On the other hand, when running the `ORKTest` app, which also uses `@import ResearchKit` (...) doesn't show this error.

It changed the problem to this (also harmless) error:

screen shot 2016-11-07 at 10 29 17 pm

But then, additionally importing the non-private *ResearchKit* module from `ORKESerialization.m` silenced it completely. (I still think there's some minor Xcode bug, as both of these errors didn't affect successful compilation and run of tests).


malhal:
Hi @rsanchezsaez I can confirm this is working for my case. (That is referencing a built version of the framework from a folder other than the project folder and adding it to framework search paths.)

However when I checked out this pull request it had a build error in ORKOrderedTask.m line 1921 etc. The NSArray isn't typed so firstObject.identifier was ambiguous. I can PR a fix for this if you like.

rsanchezsaez:
Thanks for checking!

Do you get an error or warning? What's the exact wording of it? Which version of Xcode are you running? (I'm on *Version 8.1 (8B62)*). Perhaps my Xcode version is inferring the type of the array from somewhere else?

---

Anyway, what you suggest make sense, templatized arrays are always better. Feel free to submit the PR, but let's omit the space between the type and the `<`, and also let's use `__kindof` so `ORKStep` subclasses are also valid. Like this:

     NSArray<__kindof ORKStep *> *leftSteps

Cheers!

malhal:
Xcode Version 8.1 (8B62)

```
/Users/mh/Development/Github/ResearchKit/ResearchKit/Common/ORKOrderedTask.m:1921:96: Multiple methods named 'identifier' found with mismatched result, parameter type or attributes

/Users/mh/Development/Github/ResearchKit/ResearchKit/Common/ORKOrderedTask.m:1928:91: Multiple methods named 'identifier' found with mismatched result, parameter type or attributes

/Users/mh/Development/Github/ResearchKit/ResearchKit/Common/ORKOrderedTask.m:1929:42: Multiple methods named 'identifier' found with mismatched result, parameter type or attributes
```

rsanchezsaez:
Weird. I guess it's *Xcode* and their caches getting funny. ¯\_(ツ)_/¯

umerkhan-apple:
@rsanchezsaez Any updates on this one?

rsanchezsaez:
@umerkhan-apple: Sure! I merged master and resolved the merge conflicts.

umerkhan-apple:
Thanks @rsanchezsaez !

Issue #857, opened 2016-10-30
ORKTest unit tests fail


rsanchezsaez:
`ORKTest` unit tests seem to be failing.

~~~First, all `ORKDataCollectionTests` fail~~~. (The first three tests were failing due to lack of HealthKit access permissions). First, the `testDelegateRejectDelivery` test fails. Then, the serialization test for `ORKLocation` causes a wrong memory access crash.

<img width="1552" alt="screen shot 2016-10-29 at 9 32 23 pm" src="https://cloud.githubusercontent.com/assets/444313/19834501/35bc6ca6-9e1f-11e6-805e-04ff8ed90635.png">

kelseydedoshka-apple:
@rsanchezsaez thanks for your fix for this. It was merged, therefore closing this issue.

PR #862, opened 2016-11-08
[ORKESerialization] Fix ORKLocation unit tests

rsanchezsaez:
Fixes `ORKLocation` serialization unit test which was causing a *bad memory access* crash due  to `CLCircularRegion` being initialized using `init` instead of `initWithCenter:radius:identifier:` ([Issue #857](https://github.com/ResearchKit/ResearchKit/issues/857)). I've also submitted this to *bug report* as *Radar #29156129*, because I feel that trying to initialize `CLCircularRegion` through `init` should assert, rather than creating an invalid object.

Also, fix some `ORKDataCollectionTests` typos. The `testDelegateRejectDelivery` data collection test is still failing, I didn't have the time to familiarize with the new `ORKDataCollection` classes and fix the test.

---

Perhaps you guys should add a *GitHub* PR hook so the serialization tests are run for each PR, that way we'd find these kind of errors earlier. Also, I find `ORKESerialization` code quite tricky to debug. Maybe *ResearchKit* could benefit by rewriting this file without using *C preprocessor macros*: the code would be more verbose but easier to understand, debug, and maintain.

### DIFF
--- a/ResearchKit/Common/ORKHealthSampleQueryOperation.m
+++ b/ResearchKit/Common/ORKHealthSampleQueryOperation.m
@@ -229,7 +229,7 @@ static NSUInteger const QueryLimitSize = 1000;
         
         if (!handoutSuccess) {
             doContinue = NO;
-            self.error = [NSError errorWithDomain:ORKErrorDomain code:ORKErrorException userInfo:@{NSLocalizedFailureReasonErrorKey: @"Results are not devlivered to delegate properly."}];
+            self.error = [NSError errorWithDomain:ORKErrorDomain code:ORKErrorException userInfo:@{NSLocalizedFailureReasonErrorKey: @"Results were not properly delivered to the data collection manager delegate."}];
         }
     }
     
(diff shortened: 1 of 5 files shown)


rsanchezsaez:
I've added a few commits fixing `ORKDataCollectionTests` unit tests. This completely fixes [Issue #857](https://github.com/ResearchKit/ResearchKit/issues/857)

The problem was that the collectors in the *sample rejection* test were not only receiving the locally added samples, but also the ones from the previously ran `-testDataCollection` test. I fixed that by passing different `startDate`s to each of the collectors, and also parametrizing the start and end dates on the added samples to match the ones on the collectors.

syoung-smallwisdom:
Thank you!

kelseydedoshka-apple:
@rsanchezsaez: Thanks for your contribution to ResearchKit and resolving the unit test failures!

@umerkhan-apple please take a look.

umerkhan-apple:
Looks good! 👍

PR #865, opened 2016-11-15
Add missing HealthKit characteristic types to answer format.

syoung-smallwisdom:
This commit does a few changes:

1. Add the missing characteristics of Fitzpatrick Skin Type and Wheelchair Use to the `ORKHealthKitCharacteristicTypeAnswerFormat`

2. Add `shouldRequestAuthorization` flag to both defined HealthKit answer format subclasses to allow *not* requesting authorization for access to HealthKit if this does not meet the UI/UX requirements for the app.

3. Publicly expose the `impliedAnswerFormat` method to allow for subclasses to replace the default implied answer format with their own representation.


umerkhan-apple:
@syoung-smallwisdom This PR looks good. The only thing I don't completely understand is the use-case for shouldRequestAuthorization. 

In my discussion with the team, we think if you're using a HK answer format, it should always request authorization (if the permission sheet hasn't already been granted). In the scenario that you don't want to request authorization, you should opt for using a different answer format than a HK answer format.

Let me know if we're overlooking a use-case. Thanks!

syoung-smallwisdom:
@umerkhan-apple 

The use-case for `shouldRequestAuthorization` is how we are using the model objects. We use the same model objects for both onboarding and the profile (settings). Our onboarding flow has sub-sections for eligibility, consent, registration and *then* setting up your profile. Because of the way `ORKTaskViewController` manages the timing for when to ask about permissions, it does a recursive interaction through the steps and asks for permissions before the first step that is *not* a subclass of `ORKInstructionStep`. In our case, that means that before the user is shown *anything* including eligibility and consent, they see the HealthKit permissions popover. This is incredibly jarring and doesn't meet our IRB requirements. The `shouldRequestAuthorization` flag allows for customized permission handling that happens later, when we actually want to show those permission modals.

This PR is not the full story for how we are implementing this in our apps; at this point, it is 4 months old. In the interest of *not* expanding the scope of this branch to keep including stuff that is only tangentially related, I'm using a flag that defaults to `NO` but allows us more flexibility in how we handle UI/UX around permissions.

Make sense?

umerkhan-apple:
@syoung-smallwisdom Sounds good! Thanks for the PR! 👍

PR #872, opened 2016-12-02
Fix audio step duplicate file creation bug

Erin-Mounts:
Start was getting called in viewDidAppear:, creating an audio file, and meanwhile the superclass viewDidAppear: had dispatched a block to main that then saw the recorder was started and called resume, creating another audio file, only one of which was included in the results, leaving the other which the caller didn't know about and wasn't aware needed to be deleted, using up storage space over time. The fix is to use the shouldStartTimerAutomatically flag to let the superclass viewDidAppear: start the recorder, rather than override that method in the audio step view controller to start it explicitly.

### DIFF
--- a/ResearchKit/ActiveTasks/ORKAudioStep.m
+++ b/ResearchKit/ActiveTasks/ORKAudioStep.m
@@ -48,6 +48,7 @@
     self = [super initWithIdentifier:identifier];
     if (self) {
         self.shouldShowDefaultTimer = NO;
+        self.shouldStartTimerAutomatically = YES;
     }
     return self;
 }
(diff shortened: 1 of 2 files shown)


akshay-yadav-apple:
@Erin-Mounts Thank you for fixing this issue.

ResearchKit/ActiveTasks/ORKAudioStep.m
```diff
@@ -48,6 +48,7 @@ - (instancetype)initWithIdentifier:(NSString *)identifier {
     self = [super initWithIdentifier:identifier];
     if (self) {
         self.shouldShowDefaultTimer = NO;
+        self.shouldStartTimerAutomatically = YES;
```
majeedyaseen, line 51: @umerkhan-apple will this make shouldStartTimerAutomatically to YES as default ?

PR #877, opened 2016-12-08
Expose passcode flow types

sroymdsol:
Unless I'm mistaken, currently the only way to ask the user for a passcode is to use the static `ORKPasscodeViewController.passcodeAuthenticationViewController()` method, which produces a view controller to be modally presented with little in the way of customization or integration with other steps.

One use case we have is the need to show passcode authentication as a step along with an `ORKLoginStep` so that the user can switch between the two as desired.

This PR publicly exposes the `passcodeFlow` property and associated enums at the ORKPasscodeStep level so that its value can be set to `.authenticate` when constructing the step. The constructor that takes only an identifier remains untouched so existing code is not affected.

### DIFF
--- a/ResearchKit/Common/ORKPasscodeStep.h
+++ b/ResearchKit/Common/ORKPasscodeStep.h
@@ -36,6 +36,16 @@
 NS_ASSUME_NONNULL_BEGIN
 
 /**
+ An enumeration of values used in `ORKPasscodeStepViewController` to indicate the type of flow used
+ by the view controller.
+ */
+typedef NS_ENUM(NSUInteger, ORKPasscodeFlow) {
+    ORKPasscodeFlowCreate,
+    ORKPasscodeFlowAuthenticate,
+    ORKPasscodeFlowEdit
+};
+
+/**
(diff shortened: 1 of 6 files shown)


sroymdsol:
Sorry for the closing and reopening, realized I had to make a couple changes before this was a viable PR.

umerkhan-apple:
@sroymdsol It might make more sense to create ORKPasscodeStep and providing the enum as a parameter. What do you think?

sroymdsol:
@umerkhan-apple I wasn't sure if there was a good reason why the passcodeFlow was a property of the view controller instead of the step and I thought it was best to keep the change small. Let me see what it looks like moving the passcodeFlow property to the step.

sroymdsol:
I think this is ready for another look.

kelseydedoshka-apple:
@sroymdsol looks good!

@umerkhan-apple take a look.

umerkhan-apple:
@sroymdsol Thanks for the contribution! 👍

sroymdsol:
🎉

PR #893, opened 2017-01-17
Update ChartsAndGraphs-template.markdown

coreygans:
Needed to change the Chart1.png to chart1.png so that the link would resolve.


DanKeen:
Good catch, Corey. Yay case-sensitivity!

Issue #674, opened 2016-05-05
Starting a Signature Stroke in the Top Part of the View Fails


samfalconer:
Starting to stroke part of the signature from near the top of ORKSignatureView fails.

For example, if you wanted to draw a tall letter "I" starting from the top of the view, you would be unable. Also, drawing the cross on the top of a tall capital "T" is impossible.

Attached is a screenshot where I have drawn a horizontal stroke starting from the closest to the top of the view that was possible, to demonstrate how you can't start a stroke near the top. Additionally, I drew a vertical stroke **starting from the bottom** all the way up the view, to show how much further the view goes vertically.

I didn't notice this behavior myself while working on #653 because the signatures I tried started from near the bottom of the view, however there are many people who start their signatures from the top and go down.

I witnessed someone not being able to use the signature view because of this bug.

It looks like this bug existed before #653.

![img_2202](https://cloud.githubusercontent.com/assets/3956038/15048999/0ad05bb0-12ba-11e6-87f8-edfb53392e62.png)

YuanZhu-apple:
I can confirm this issue. 
To draw a vertical line, I have to start from bottom to top. Top to bottom fails.

umerkhan-apple:
PR #894 should fix this issue. Let me know if the issue still remains.

rsanchezsaez:
Actually, the already merged [PR #894](https://github.com/ResearchKit/ResearchKit/pull/894) didn't fix the issue.

I submitted [PR #898](https://github.com/ResearchKit/ResearchKit/pull/898) which does fix it.

umerkhan-apple:
#898 Fixes the issue! Thanks!

PR #898, opened 2017-01-25
[ORKSignatureStepViewController] Fix not being able to start traces on the top part of the signature view

rsanchezsaez:
Fixes not being able to start traces on the top part of the signature view ([Issue #674](https://github.com/ResearchKit/ResearchKit/issues/674)).

The problem was that the height of the `ORKSignatureView` was larger than the one of the containing `ORKSignatureWrapperView` (it was peeking out from the top), so the signing gesture recognizer was not being triggered if the initial touch began within the overextended part.

I removed the overextending part. As a result the layout is slightly different (and I also had to reduce the *iPhone 4* height by 10 pixels to avoid vertical scrolling), but I think this is more coherent from the functionality point of view. See screenshots below.

---

Another option would have been overriding `- (BOOL)pointInside:(CGPoint)point withEvent:(UIEvent *)event` to take into account the overextended layout, but I feel the proposed solution is simpler and less hackish.

---
Before (*iPhone 4s*)
simulator screen shot jan 24 2017 8 41 16 pm

---
After (*iPhone 4s*)
simulator screen shot jan 24 2017 8 42 05 pm

---
Before (*iPhone 7*)
simulator screen shot jan 24 2017 8 43 49 pm

---
After (*iPhone 7*)
simulator screen shot jan 24 2017 8 42 50 pm

### DIFF
--- a/ResearchKit/Common/ORKSignatureStepViewController.m
+++ b/ResearchKit/Common/ORKSignatureStepViewController.m
@@ -45,7 +45,7 @@
 #import "ORKSkin.h"
 
 
-@interface ORKConsentSignatureWrapperView : UIView
+@interface ORKSignatureWrapperView : UIView
 
 @property (nonatomic, strong) ORKSignatureView *signatureView;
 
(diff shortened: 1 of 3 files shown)


umerkhan-apple:
@rsanchezsaez Thank you for this fix! 👍

PR #900, opened 2017-01-26
Allow inheritance of navigation rules outside ResearchKit

syoung-smallwisdom:
This will still throw an exception if the developer attempts to initialize an instance of the base class without inheriting it but allows developers to create navigation rules that inherit directly from the base class.


syoung-smallwisdom:
Closing while I change this into a protocol. Will reopen shortly.

PR #902, opened 2017-01-28
Add method to ORKStepViewController to allow adding a result

syoung-smallwisdom:
Add a result to the step view controller's `ORKStepResult`. By default, the property for the step view controller's result will instantiate a copy of the result each time it is called. Therefore, the result cannot be mutated by adding a result to it's result array.

This method can be called by a delegate to add a result to a given step in a way that will be retained by the step, but will allow both the step and the ORKResult to be immutable when passed to the task view controller.

We have been finding that many of the researchers and developers we work with assume that this is possible and then become frustrated and confused when it doesn’t work.


syoung-smallwisdom:
@darrenlevy 
This PR also fixes your override of the result to return an `ORKStepResult` with the `ORKRangeOfMotionResult` as one of the results in the step result collection. Please ensure that fixing this will not break your app. Thanks!

darrenlevy:
Nice! Thanks, @syoung-smallwisdom 👍

syoung-smallwisdom:
Fixes https://github.com/ResearchKit/ResearchKit/issues/930

umerkhan-apple:
@syoung-smallwisdom This PR is great! Thank you for this great contribution! 👍

PR #903, opened 2017-01-28
Fix fashion typo

stevemoser:


kelseydedoshka-apple:
@stevemoser thanks for the fix!

@umerkhan-apple take a look.

PR #904, opened 2017-01-28
Fix ORKTimedWalkResult isEqual: and copyWithZone:

stevemoser:
It appears this was most likely the result of a copy and paste error. IMHO tests should be created for this class to catch small bugs like this one.


stevemoser:
If I had more time right now I would do it myself. Maybe another time 😄

kelseydedoshka-apple:
@stevemoser thanks for the fix!

@umerkhan-apple take a look.

umerkhan-apple:
@stevemoser Great catch! Thanks for the fix!

PR #908, opened 2017-01-30
Adding specific images for Tremor test left and right hand

joshbruhin:
Adding specific images for Tremor test left and right hand instead of simply flipping the right-handed image for use as the left. Bridge task 1513


kelseydedoshka-apple:
@joshbruhin thanks for adding this!

@umerkhan-apple take a look!

umerkhan-apple:
@joshbruhin Looks great! Thanks for this contribution! 👍

PR #913, opened 2017-02-07
Add missing height question

stevemoser:


kelseydedoshka-apple:
@stevemoser thanks for adding this in!

@umerkhan-apple - take a look.

PR #915, opened 2017-02-09
Call becomeFirstResponder with dispatch to main queue after delay

joshbruhin:
When user is prompted to use Touch ID and they tap 'cancel', the number pad keyboard is shown (via call to becomeFirstResponder). But, the keyboard is invisible. It is there and not hidden because the user can still tap the keys, but cannot see them. Something inexplicable is happening when switching from Touch ID context back to app context. Calling becomeFirstResponder with dispatch to main queue after delay fixes this.

### DIFF
--- a/ResearchKit/Common/ORKPasscodeStepViewController.m
+++ b/ResearchKit/Common/ORKPasscodeStepViewController.m
@@ -369,7 +369,17 @@ static CGFloat const kForgotPasscodeHeight              = 100.0f;
                                                             }]];
                     [strongSelf presentViewController:alert animated:YES completion:nil];
                 } else if (error.code == LAErrorUserCancel) {
-                    [strongSelf makePasscodeViewBecomeFirstResponder];
+
+                    // call becomeFirstResponder here to show the keyboard. dispatch to main queue with
+                    // delay because without it, the transition from the touch ID context back to the app
+                    // inexplicably causes the keyboard to be invisble. It's not hidden, as user can still
+                    // tap keys, but cannot see them
+                    
+                    double delayInSeconds = 0.3;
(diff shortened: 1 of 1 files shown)


ResearchKit/Common/ORKPasscodeStepViewController.m
```diff
@@ -369,7 +369,17 @@ - (void)promptTouchId {
...
                     [strongSelf presentViewController:alert animated:YES completion:nil];
                 } else if (error.code == LAErrorUserCancel) {
-                    [strongSelf makePasscodeViewBecomeFirstResponder];
+
+                    // call becomeFirstResponder here to show the keyboard. dispatch to main queue with
+                    // delay because without it, the transition from the touch ID context back to the app
+                    // inexplicably causes the keyboard to be invisble. It's not hidden, as user can still
+                    // tap keys, but cannot see them
+                    
+                    double delayInSeconds = 0.3;
```
syoung-smallwisdom: @joshbruhin 
Thanks for figuring this out. How did you choose this delay? I seem to recall (fuzzy on this one) that standard animation time for iOS presentation/dismissal is 0.2 seconds?

joshbruhin:
I've always thought it was .3 secs. I'll see if I can determine differently.

syoung-smallwisdom:
@joshbruhin No worries. Just curious. Thanks!

syoung-smallwisdom:
@joshbruhin 
Thanks for fixing. I've updated both https://github.com/Sage-Bionetworks/ResearchKit and mPower-AppStore.

ResearchKit/Common/ORKPasscodeStepViewController.m
umerkhan-apple: @joshbruhin It would be nice to know how you decided on 0.3 for the delay. Thanks!

PR #920, opened 2017-02-15
invert long and short investigator descriptions (logical)

Damien-B:
It seems more logical to have an investigatorShortDescription when we want to share only to one entity (share only), and an investigatorLongDescription when we want to share to multiple entities (share widely), right ?

Just a reminder of the strings for widely and only :
`"CONSENT_SHARE_WIDELY_%@" = "Share my data with %@ and qualified researchers worldwide";
"CONSENT_SHARE_ONLY_%@" = "Only share my data with %@";`

### DIFF
--- a/ResearchKit/Consent/ORKConsentSharingStep.m
+++ b/ResearchKit/Consent/ORKConsentSharingStep.m
@@ -70,8 +70,8 @@
         }
         
         self.answerFormat = [ORKAnswerFormat choiceAnswerFormatWithStyle:ORKChoiceAnswerStyleSingleChoice textChoices:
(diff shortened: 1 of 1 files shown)


Damien-B:
The more I look and think about it the less I get the logical way of associating these strings. I think the best would be to change the terms, something like investigatorDetailedDescription instead of investigatorLongDescription

syoung-smallwisdom:
@Damien-B I am not arguing that these descriptors are confusing. However, making this change would break every app that uses this step and is currently pointing the other way. If you would like to add a different `-init` method with better names, please do so, but deprecate the old one so existing apps won't break. Thanks.

Damien-B:
@syoung-smallwisdom That's what I'm ending up doing, maybe it could be great to clarify these descriptors in futur releases.
For the projects already using ResearchKit it's just a change they can handle in their update process. Anyway I'm closing this thread thanks for your answer.

PR #944, opened 2017-03-12
Fix app delegate for swift 3

stevemoser:
Because of the changes to Swift 3 the sample app’s delegate methods
were not being called which made the apps fail in ways that aren’t
apparent at first launch.


umerkhan-apple:
@stevemoser Great catch! Thanks! 👍

PR #952, opened 2017-03-20
[ORKTextAnswerFormat] Replace the validation regex pattern by a NSRegularExpression object

rsanchezsaez:
This PR replaces the *`validationRegex` pattern* initializer argument and property by an `NSRegularExpression` *validationRegularExpression` object. It also adds `NSRegularExpression` serialization utility functions and makes sure the corresponding tests pass.

This is a follow up to PR https://github.com/ResearchKit/ResearchKit/pull/567. I'm submitting it against `master` instead of against the original PR branch because it seems the original author has been inactive for some time.


ResearchKit/Common/ORKAnswerFormat.h
```diff
@@ -151,8 +151,8 @@ ORK_CLASS_AVAILABLE
 
 + (ORKTextAnswerFormat *)textAnswerFormatWithMaximumLength:(NSInteger)maximumLength;
 
-+ (ORKTextAnswerFormat *)textAnswerFormatWithValidationRegex:(NSString *)validationRegex
-                                              invalidMessage:(NSString *)invalidMessage;
++ (ORKTextAnswerFormat *)textAnswerFormatWithValidationRegularExpression:(NSRegularExpression *)validationRegularExpression
```
syoung-smallwisdom: Could you please leave the original method and deprecate it? That way, folks who are using this method aren't left trying to figure out why their code is suddenly failing.

rsanchezsaez: I don't feel bad breaking backwards compatibility on `ResearchKit` in pursuit of a cleaner API.

But if the consensus is that adding back the deprecated initializer is a good idea, I'll happily add it. :)

syoung-smallwisdom: Deprecation flags serve a valuable purpose in any API. It allows other developers to pin to master and use the latest build (which often includes bug fixes) without the worry that it's going to break their own code, and it serves as a document of the change you are making and how the developer can modify their code to conform to the new method.

umerkhan-apple: I agree with @syoung-smallwisdom that deprecation flags helps notify developers of API change, so they can eventually conform to it without having a broken build. However, it also encourages them to continue using deprecated API. The best way to encourage developers to conform to new API is to use the compiler to notify them of the API change via a build failure.

That being said, I'm indifferent about which approach we take here, as they both provide value in different ways.

rsanchezsaez: Yeah, I know what deprecation flags are intended for. :) 

However, in this case, I'm inclined to lean on Umer's side.

I see a more reasonable use for deprecation flags on frameworks that are bundled with the *iOS SDK*, where OS updates are greatly encouraged due to severity of security fixes and the scope of the improvements.

On a smaller standalone framework such a *ResearchKit*, developers can keep with their current version if they don't want the hassle of upgrading. This is such a small and trivial change from the API calling point that I continue to favor a shiny and lean API over causing some hassle to developers. :D

syoung-smallwisdom:
The deprecation flag is a warning that "This will be deleted in future versions" so anyone who ignores the warning does so at their own risk. In fact, I think it is even more valuable in an open-source framework like this because it allows developers to pull in code fixes without having to worry about breaking changes. Additionally, leaving breadcrumbs is a good idea if you want to encourage developers to continue contributing to a common framework, rather than just forking and going off on their own.

That said, go ahead and merge. This discussion documents the change which is equivalent, I suppose. ¯\_(ツ)_/¯

umerkhan-apple:
@syoung-smallwisdom I agree that this discussion can serve as the documented change. Thank you for the great explanation! 👍

PR #958, opened 2017-03-21
Fixed bug with ORKTextChoice not using the correct tint color

ninoguba:
This fixes issue #874


kelseydedoshka-apple:
@ninoguba - thanks for the fix!

@umerkhan-apple - take a look.

umerkhan-apple:
Looks good! Great catch! Thanks for submitting the PR! 👍

PR #964, opened 2017-03-29
[Misc] Fix Xcode 8.3 warnings and error

rsanchezsaez:
Please have a look at the new nullability annotations I added to some `ORKLocation` properties, as I'm not sure if such a location with these `nil` properties is still valid within the *ResearchKit* context. Also, review the change I made to `ORKSample`'s `ProfileViewController` regarding the default age text value.

Also, starting on Xcode 8.3, the `module.modulemap` file caused the compilation error described in https://github.com/ResearchKit/ResearchKit/issues/922. Renaming it to `ResearchKit.modulemap` seems to fix the issue.


moppymopperson:
This fixed #922 for me.

umerkhan-apple:
Thanks @rsanchezsaez! 👍

PR #965, opened 2017-03-31
Fixes #833. "Date Step - Mismatch between picker.date and what is shown in the UIDatePicker”.

### LINKED ISSUE #833: Date Step - Mismatch between picker.date and what is shown in the UIDatePicker
Creating the `ORKDateAnswerFormat` with a particular `NSCalendar` is not enough to ensure the time zone, provided with the calendar, is used by the picker (particularly in the _UI side of things_).

Check how `_pickerView.date` and the date displayed in the UI are different:
picker view bug

**The problem is** :
Inside `ORKDateTimePicker` the `_pickerView.calendar` is set with the calendar passed in the init method. But that is not enough, the `_pickerView.timeZone` property needs to be set too if we want to use a particular time zone. Sounds weird but that seems to be the way `UIPickerView` works.

Both of the following changes do the work for me:
1. Set the time zone in the method that creates the picker view.

``` Objective-C
- (UIDatePicker *)pickerView {
    if (_pickerView == nil) {
        _pickerView = [[UIDatePicker alloc] init];

        // mirlhaq ----
        _pickerView.timeZone = _calendar.timeZone;
        // mirlhaq ----

        [_pickerView addTarget:self action:@selector(valueChanged:) forControlEvents:UIControlEventValueChanged];
        self.answerFormat = _answerFormat;
        self.answer = _answer;
    }
    return _pickerView;
}
```
1. Set the time zone in the method that sets the answer format.

``` Objective-C
- (void)setAnswerFormat:(ORKAnswerFormat *)answerFormat {
...
    if ([self isTimeOfDay]) {
        [self setDate:[(ORKTimeOfDayAnswerFormat *)answerFormat pickerDefaultDate]];
    } else {
        ORKDateAnswerFormat *dateAnswerFormat = (ORKDateAnswerFormat *)answerFormat;
        [self setDate:[dateAnswerFormat pickerDefaultDate]];
        _pickerView.calendar = [dateAnswerFormat currentCalendar];

        // mirlhaq -----
        _pickerView.timeZone = _pickerView.calendar.timeZone;
        // mirlhaq -----

        _calendar = [dateAnswerFormat currentCalendar];     
...
}
```

mirlhaq:
#833

### DIFF
--- a/ResearchKit/Common/ORKDateTimePicker.m
+++ b/ResearchKit/Common/ORKDateTimePicker.m
@@ -135,7 +135,10 @@
     } else {
         ORKDateAnswerFormat *dateAnswerFormat = (ORKDateAnswerFormat *)answerFormat;
         [self setDate:[dateAnswerFormat pickerDefaultDate]];
+        
         _pickerView.calendar = [dateAnswerFormat currentCalendar];
+        _pickerView.timeZone = _pickerView.calendar.timeZone;
+        
         _calendar = [dateAnswerFormat currentCalendar];
         
         [_pickerView setMinimumDate:[dateAnswerFormat pickerMinimumDate]];


kelseydedoshka-apple:
@mirlhaq - looks great! 
@Pariecemckinney-apple - please take a look for merging.

PR #966, opened 2017-04-03
Fix class visibility for new types of Steps, fixes #649

### LINKED ISSUE #649: ORKESerialization runs on simulator but fails to archive

cheif:
It seems like issue #649 has started to appear again. This fixes it by changing the class-visibility for two new steps that have been added since the last fix was merged.

### DIFF
--- a/ResearchKit/ActiveTasks/ORKShoulderRangeOfMotionStep.h
+++ b/ResearchKit/ActiveTasks/ORKShoulderRangeOfMotionStep.h
@@ -36,6 +36,7 @@
  The `ORKShoulderRangeOfMotionStep` class represents a step that takes a range of motion measurement
  for either a left or right shoulder.
  */
+ORK_CLASS_AVAILABLE
 @interface ORKShoulderRangeOfMotionStep : ORKRangeOfMotionStep
 
 @end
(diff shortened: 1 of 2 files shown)


kelseydedoshka-apple:
@cheif - looks good, thanks for adding this!

PR #970, opened 2017-04-06
Remove deprecation handling intended to support iOS 8

syoung-smallwisdom:
When Xcode 8 beta (iOS 10) and Xcode 7 (iOS 8) were still both in use by the development community, support for iOS 8 facilitated development by allowing developers to develop against a code base that included support for both iOS and Xcode versions. Since ResearchKit has been pinned to iOS 9, this work-around is no longer required.


syoung-smallwisdom:
Closing due to inactivity

PR #971, opened 2017-04-08
Update README.md Swift example code to latest Swift syntax

benjamin-42:
The `HelloResearchWorld` Swift code example from the README isn't up-to-date with latest Swift syntax and does not compile with Swift 3.x.


kelseydedoshka-apple:
@benjamin-42 - apologies for the delay here, this looks great!

Issue #960, opened 2017-03-23
The Create Passcode View Controller is not fully laid out when presented


rsanchezsaez:
The Create Passcode View Controller is not fully laid out when presented. Run the `Create Passcode` item in `ORKTest` and observe how the view controller is being laid out at the same time as it presents.

kelseydedoshka-apple:
@rsanchezsaez - thanks for bringing this up. Closing this as I have merged @benjamin-42 's PR to resolve this.

PR #972, opened 2017-04-08
[ORKPasscodeStepViewController] Make sure the view has been laid out before showing the keyboard

benjamin-42:
This is a fix for issue #960, `-viewDidAppear:` is called but the view isn't on screen yet when the view controller is set in a `UIPageViewController`. Might be a bug from UIKit? https://forums.developer.apple.com/thread/15441


kelseydedoshka-apple:
@benjamin-42 - looks great thanks for the fix here!

Issue #376, opened 2015-08-17
Heart rate measurement using back camera


pawisoon:
Hi all!
I'm pretty sure I saw some announcements that ResearchKit will support heart rate measurement but now when I Google it again I can't see such information.. 

In docs there's also a clue saying(under the second sceeenshot):
"The rest step. This step is skipped if heart rate data is unavailable." 

![screen shot 2015-08-17 at 12 48 26](https://cloud.githubusercontent.com/assets/6951295/9304366/3ce92ed2-44df-11e5-88b6-18fd6e619831.png)

Is it implemented in ResearchKit or not yet? Any info? 

Cheers 
P

p2:
You would typically read heart rate and other health information from _HealthKit_. Whether implementing heart rate detection using camera+light or something more advanced like [Cardiio](cardiovascular disease) into ResearchKit would be desirable is a good question.

pawisoon:
I'll wait for more replies.. maybe somebody knows something about future updates on it. 

My plan to do this anyway so if I will have enough time I will try to push it to ResearchKit..

YuanZhu-apple:
@p2 is correct. ResearchKit itself does not measure heart rate, it only reads heart rate from HealthKit.
@pawisoon Can you share with us what do you plan to implement?

pawisoon:
My project demands from me to implement heart rate measurement right after some activity so I got the basic idea of what I want to do. 
First of all I'll try to get the camera and led(optional) working on demand and then I will focus on function for counting the bpm.
I found some useful old projects online but I'm not into Obj-C so I will rewrite it using Swift. 
If someone wants to help with it or something I'd really appreciate it..

scdi:
When an item is labeled "new feature" is it implemented? Or, is it just a labeled to distinguish it from a bug?
jude

YuanZhu-apple:
@scdi Just distinguish it from a bug. And it is not on the road map.

scdi:
Thank you. Any news on when approximately the next release will take place?

dallinja:
@pawisoon I found those old camera/LED projects in Obj-C online too. Any luck converting to Swift?

AbidSana:
@pawisoon , I want to implement the detection of heart rate using the iPhone's cam, but I found just some old tutorials in Obj-C. I want to know if you found any solution to implement this feature using Swift ?
Any help will be appreciated :)

kashifmehmood9:
[...]

PR #981, opened 2017-05-17
Now HR can be obtained trough the device's back camera. Feature #376

rochoSC:
I used the VideoCaptureStep as a guideline and created a new step that uses the device's back camera and measures user's hearth rate. The step uses an FFT based approach to estimates the HR of the user every second once the measurement is started and enough samples have been collected (64 samples just to start showing information, supposing we are sampling at 30FPS we are talking about 2 seconds). The step retrieves a list of estimated HR values during the given session that can last no more than 2 minutes and no less than 20 seconds.

### DIFF
--- a/ResearchKit.xcodeproj/project.pbxproj
+++ b/ResearchKit.xcodeproj/project.pbxproj
@@ -557,6 +557,15 @@
 		D442397E1AF17F7600559D96 /* ORKImageCaptureStepViewController.m in Sources */ = {isa = PBXBuildFile; fileRef = D442397C1AF17F7600559D96 /* ORKImageCaptureStepViewController.m */; };
 		D458520A1AF6CCFA00A2DE13 /* ORKImageCaptureCameraPreviewView.h in Headers */ = {isa = PBXBuildFile; fileRef = D45852081AF6CCFA00A2DE13 /* ORKImageCaptureCameraPreviewView.h */; };
 		D458520B1AF6CCFA00A2DE13 /* ORKImageCaptureCameraPreviewView.m in Sources */ = {isa = PBXBuildFile; fileRef = D45852091AF6CCFA00A2DE13 /* ORKImageCaptureCameraPreviewView.m */; };
+		DBBF16A41ECBC72700602B86 /* ORKHrCaptureStepViewController.m in Sources */ = {isa = PBXBuildFile; fileRef = DBE4AF0D1ECBBC3100C9EE78 /* ORKHrCaptureStepViewController.m */; };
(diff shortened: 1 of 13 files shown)


kelseydedoshka-apple:
@rochoSC - this is a really interesting idea. I am curious if you have had the opportunity to compare accuracy against the HR taken via the watch/HealthKit. This would be interesting as a way to gather Heart Rate for users without a watch however I do have some concerns about multiple Heart Rate sources and potentially differing data.

rochoSC:
@kelseydedoshka-apple Thanks for taking a look at it! I totally agree, the HR estimates provided from difference sources should be congruent. I could not compare the results obtained from this against those of the smartwatch just because I don't have one, however, I informally compared the results I was getting with those of a Empatica E4 device, a clinical quality wearable device designed specifically for HR obtention. The results were very close with an average error of about 2 to 3 BPM.

kelseydedoshka-apple:
@rochoSC - thats great, and exciting to hear the results were so close. Do you envision a user utilizing this step and then writing that data to HealthKit for analysis later? I would love to understand what your use case for this was and how effective you found it? Again I have a few reservations about the secondary source of HR data but if used appropriately I think this could be really valuable to users.

rochoSC:
@kelseydedoshka-apple Being completely honest I didn't have a use case that motivated me to do this. The reason why I decided to contribute its because I have been working with this PPG stuff on smartphone cameras (specifically with an Ipod touch) for my Master thesis. When I read about the Research Kit library, I though that since the only way a user can get their HR is trough the smart watch, it may be a good idea to incorporate another way to get such information for people who don't own a SmartWatch. So, I would say that I envision the usage of this step as an alternative of measuring the HR after certain tasks or activities as required. 

On the other hand, about the reservations you have about the secondary source, I totally understand. I would say that there are two ways to ensure that this measurements are coherent among them. The first one is to validate the HR obtained from the camera with that of the smartwatch and do whatever its required to match the HR of the smartwatch. The second and easiest in terms of development its simply match the algorithm that is being used to get HR from the smartwatch PPG signal and apply such algorithm to the PPG signal obtained from the camera. However, I don't think the code for such HR is available.

I hope I addressed all your questions, if I did't please let me know and I will try to explain myself better.

kelseydedoshka-apple:
@rochoSC - I love to hear that you considered contributing this while doing your thesis. This use case you have mentioned for gathering HR from non smart watch users is a really interesting one to consider and I appreciate your thoughts on how to handle multiple sources of data. I am interested to throw this idea around to some other folks on our side to get feedback from them as well, and would be interested to keep this discussion open as I gather more input.

rochoSC:
@kelseydedoshka-apple That sounds awesome, thanks for considering this contribution. I'm looking forward hearing back from you guys!

davwillev:
This is a great feature, particularly for the health sciences! Why has it not yet been merged into the master?
@srinathtm-apple

rodelt-medopad:
Hi @srinathtm-apple , is Apple still considering this feature to be available in ResearchKit?

PR #988, opened 2017-06-29
Feature/plus minus button accessory

Garrettv11:
Added a plus and a minus keyboard accessory to allow the input of negative values for decimal and integer type questions.


Pariecemckinney-apple:
Hello @Garrettv11 ,

Thanks so much for the PR. We will review this on our end and get back to you soon. Thanks again!

Issue #1012, opened 2017-09-14
Height response Issue


arun14999:
**Description**:-

**Steps to replicate**:-
In height QuestionStep, enter your height as 5'2" OR 4'3" in response to a question that uses the standard RK height response type
tap next to go to the next question
tap back to go back to the height question

**Actual result**:-

height has been reduced by 1 inch
If 5'2" was entered - height shown is now 5'1" 
If 4'3" was entered - height shown is now 4'2"

**Expected result**:-
Heights entered should not change unless changed by the user

ninoguba:
The fix to this is a minor change to the ORKInchesToFeetAndInches() function in ORKHelpers_Internal.h

```
ORK_INLINE void ORKInchesToFeetAndInches(double inches, double *outFeet, double *outInches) {
    if (outFeet == NULL || outInches == NULL) {
        return;
    }
    *outFeet = floor(inches / 12);
    *outInches = round(fmod(inches, 12));
}
```
Just need to round the inches @rsanchezsaez

ninoguba:
Submitted PR #1013 to fix this

arun14999:
Thanks a lot.

Pariecemckinney-apple:
@arun14999 Thanks for posting the issue and thanks @ninoguba for the quick fix and PR. Will merge it now! Closing

PR #1013, opened 2017-09-16
Fix for #1012 to prevent inch value from decrementing unexpectedly

ninoguba:


Pariecemckinney-apple:
@ninoguba Thank you for the PR!

PR #1025, opened 2017-10-31
WIP: ORKInstructionStep detailText as NSAttributedString

tobiasjungnickel:
feature to set an attributed text as detailText of ORKInstructionStep

setting NSString as detailText of ORKInstructionStep for existing classes have to be migrated to
`step.detailText = [[NSAttributedString alloc] initWithString:@"myString"];`


tobiasjungnickel:
Does anyone has a good architectural idea to not implement defaultFont of ORKDefaultFont in ORKSubheadlineLabel only for this case to also enable setting bold and italic fonts this way?

kelseydedoshka-apple:
@tobiasjungnickel - I really like this idea, but have a quick question on your implementation choice. From a use case perspective, do you often find that you want to use an attributed string for the detailText? I want to ensure that this is the most common use case so we don't force users to migrate to what you mentioned above, if the common use case is just for an NSString. In that case we could add an additional property like attributedDetailText, which if set overrides the detailText and if not set still allows users to use detailText as is.

Thoughts on this?

DanKeen:
Second @kelseydedoshka-apple 's comment here - please mirror other UIKit APIs like `UILabel` here and add a new `attributedDetailText` property instead of overriding `detailText`.

tobiasjungnickel:
@kelseydedoshka-apple i think NSString is still the commen use case and that is why i was asking about a "good architectural idea" so that other users don't have to migrate. I agree with both - you and @DanKeen - that introducing an additional `attributedDetailText` looks like the best solution.

kelseydedoshka-apple:
@tobiasjungnickel - great to hear. Looking forward to an update to the PR here if you get a chance!

tobiasjungnickel:
I added an additional attributed detail text that overrides detail text if set

tobiasjungnickel:
now it's possible to set an `attributedDetailText`, but unfortunately i can't set another font attributes (family, weight, ...), because `instructionLabel` in `ORKStepHeaderView` is an `ORKSubheadlineLabel` which implement `ORKDefaultFont` to override the font. Any suggestions how to custom override the font?

kelseydedoshka-apple:
@tobiasjungnickel - initial thoughts on this would be to do something similar to what you just implemented and have a font property for the instruction label that would override the default font like you mentioned above.

However it seems as though we are starting to go down the path of too many customizations and potentially starting to defeat the purpose of using the standardized labels and styles. Perhaps not using the instruction label and using a custom view to almost replicate the instruction label with whatever text and style you want might be a better approach. Thoughts on something like this?

tobiasjungnickel:
@kelseydedoshka-apple even a font property wouldn't be enough for me, because I need different fonts in the instruction label (e. g. parts of bold text).

I agree that it seems too much customization, because i also need `title` and `text` labels in `ORKStep` to be styled. I just wanted to use and integrate it easily, but it seems my Active Task needs more to be designed by myself as custom views.

kelseydedoshka-apple:
@tobiasjungnickel - would be interested to see what you come up with here if you go down the more customized route.

PR #1026, opened 2017-11-02
Add Info.plist keys for microphone and camera access

corduroy:
Adds the necessary Info.plist keys for requesting access to camera and microphone, as required in the Image Capture Step and Video Capture Step from ORKCatalog.app


kelseydedoshka-apple:
@corduroy -  Great catch, thanks for the quick fix.
@Pariecemckinney-apple - please take a look!

PR #1029, opened 2017-11-14
[Projects] Update projects to recommended Xcode 9.1 settings and fix all the warnings

rsanchezsaez:
I've updated the projects to the recommended Xcode 9.1 and fixed all the warnings. Note that I've removed some deprecated `ORKTextChoice` code, please have a look. You can no longer pass raw `NSString`s to those, but I updated all the sample apps accordingly.

---

Also, this warning remains for all three sample apps:

screenshot 2017-11-14 00 51 01

Could you go to the original `ORKCatalog` and `ORKTest` design sources and upload the corresponding hi-res icons? I will then be able to update the `ORKSample` one (I originally created it and have its PSD source, but in order to update it I need the hi-res `ORKCatalog`'s one as a template).


srinathtm-apple:
Hey @rsanchezsaez - the PR looks good and yes we will try to secure the asset for ORKCatalog before merging this.

rsanchezsaez:
@srinathtm-apple Actually, if you guys want to merge this we can get rid of most of the warnings now to avoid noise, and we can update the assets in a separate PRs. :-)

Issue #873, opened 2016-12-02
Proposal: ResearchKit Scoring Module


thediary:
# Overview
ResearchKit’s Surveys module is very useful for collecting structured data from an end user, however ResearchKit has limited functionalities in utilizing collected data.

We would like to propose a way to associate a numerical score to an ORKStep and to compute a total score base on step scores for a given ORKTask.

A scoring module proves useful when dynamically creating assessments using ResearchKit (ie. assessments that are not created at compile time, but at runtime—either by the application or an end user.) The goal of the proposed module is to allow a developer to easily associate data collected through ResearchKit’s Surveys module with a numerical score, and to access this score for persistence or for display in an app’s front-end.

An example of a scored assessment is the standardize PHQ-9 assessment. Patients are asked a series of questions centered around their mental health, to which one answers with one of four options on a scale (not at all, several days, more than half the days, nearly every day.) Each one of these options is assigned a score (from 0 to 3 “points.”) After all questions in the assessment are completed, a medical professional can use the total score to make a tentative diagnosis of depression.

Other examples include pain scoring, as well as more advanced implementations such as Risk Stratification (which often also include use of dynamic step scores.)

The following proposal is a draft for implementing such a scoring module on ResearchKit. While many details are to be discussed on the actual implementation, the central goal of having a fast and easy way to evaluate data collected from ResearchKit remains the same.

## Basics of the proposed scoring module
- Association of static step score
- Association of dynamic step score
- Computation of task total score (DSL)

### Association of static step score
Each `ORKStep` can have a optional static score value associated, if a task is associated with a static score value, the value will be used to calculate the whole score of the parent `ORKTask`.

`ORKStep` may have positive or negative static score value associated.

```
step1.staticScoreValue = 1
step2.staticScoreValue = -3 
step3.staticScoreValue = 2
```

`ORKTaskResult` can have a method for retrieving the total final score of the task, based on the individual values and outcome of its steps and step results.

```
taskResult.totalScore()
```

### Association of dynamic step score
While a step may have a static base value associated, to support more complex scenarios, e.g. a step’s value may depends on outcome of another step,  we need to able to define a way for a step to return dynamic values.
A way to associate dynamic scoring logic is to attacked a block to a `ORKStep`.

For example:
```
step1.dynamicScoreValue = { task, step, stepResult in 
    if 
	let booleanResult == stepResult as? ORKBooleanQuestionResult, 
	let answer = booleanResult.booleanAnswer {
[...]

Issue #767, opened 2016-07-26
Proposal ORKDrawingStep


oliverschaefer:
## Motivation
Implement `ORKDrawingStep` which lets users draw any type of line-art with their finger/apple pencil
## Proposed Solution
Port [WWDC example] (https://developer.apple.com/library/prerelease/content/samplecode/SpeedSketch/Introduction/Intro.html) to ObjC and implement the corresponding `ORKDrawingStepViewController` and `ORKDrawingView`classes
## Impact on existing code
ORKSignatureStep could become a subclass of ORKDrawingStep. New implementation could fix #674 

@YuanZhu-apple @syoung-smallwisdom What do you think?

syoung-smallwisdom:
I think it's a great idea.

YuanZhu-apple:
@oliverschaefer Sounds good. My 2 cents, currently `ORKSignatureStep` carries no parameters, not sure it actually need to be the subclass of `ORKDrawingStep`.

oliverschaefer:
@YuanZhu-apple Thanks for your input. Turns out, some aspects of SpeedSketch are not quite ready for real-word usage, see `CGDrawingEngine.swift` lines 137-144 for example. Could you help me do the port to ObjC and resolve these issues?

YuanZhu-apple:
@oliverschaefer I think it basically equals to 

```
[array enumerateObjectsUsingBlock:^(id object, NSUInteger idx, BOOL *stop) {
    // do something with object
}];
```

dillan:
@oliverschaefer I would love to see this happen as well. I think it would be particularly useful if the user was able to annotate an image. For example, allowing the user to annotate an anatomical model to indicate where they feel pain.

oliverschaefer:
@dillan Yep, this was also the use case that I was thinking of. However, I'm currently working on printing support and the video instruction step and will only be able to work on the drawing step once my existing PRs are merged. This my take some time, since there currently isn't much activity going on. I guess the engineers in Cupertino are busy preparing for the macOS release.

dillan:
It looks like the SpeedSketch sample code link above is no longer working. I believe it's the sample code for the [WWDC 2016 Session 220 - Leveraging Touch Input on iOS](https://developer.apple.com/videos/play/wwdc2016/220/). I'd be happy to help port it over to ObjC if anyone can link me to the code.

oliverschaefer:
[...]

Issue #367, opened 2015-08-07
How to add a body shader step?


rugheid:
I would like to add a step for the user to shade certain spots on a human body. I already have the code for this from an application I developed earlier. I can enhance that code and make the layout a bit better. Here's a screenshot from that application:
![drawing](https://cloud.githubusercontent.com/assets/9199556/9133550/da46526c-3cfc-11e5-8443-aca58f040ce3.png)
As you can see the controls have to be more intuitive...
How would I go about adding such a step? Should I add it as an active step? Do I just have to add a new answer format and provide the needed control views?

Thanks

umerkhan-apple:
@rugheid This looks like a very interesting step to add into RK. I think it would be best to add it as an answer format. 

In terms of the UI, could you explain exactly what the controls do? What is the purpose of the slider? What does the stop button do?

Also in terms of the design, you would probably want to consider how to adopt this for an iPhone screen. One approach could be to just show one side of the body and provide a button to flip the body and show the back side. It would also be nice to have cleaner/smoother outlines on the body graphic to give it a more flat and consistent look like other RK assets.

rugheid:
@umerkhan-apple The slider controlled the pencil thickness, the draw and erase buttons switched to draw or erase mode... Since it was incorporated in a vertical scroll view, the scrolling had to be disabled to draw. The stop button re-enabled the scrolling and stopped the drawing...
As I mentioned in my original text, I will improve these controls to be more intuitive.

The idea to flip the body on an iPhone screen is great, I'll try to implement that.
I was already planning to redraw the body so it would look better as well.

There is one issue I need to ask something about. The drawing view is implemented using a custom view, so the dimensions of the exported image depend on the device. A retina screen yields a bigger image. In the app, I always reduced the graphics content to non-retina quality. I did this for performance issues as well. Should I change the drawing to use the retina quality and resize the image afterwards?

Thanks for your help

YuanZhu-apple:
@rugheid For the body shape, I suggest you using Core Graphics to draw it in `drawRect:` in code.
Please refer to `ORKSpatialSpanTargetView` as an example :
https://github.com/ResearchKit/ResearchKit/blob/master/ResearchKit/ActiveTasks/ORKSpatialSpanTargetView.m

umerkhan-apple:
@rugheid The idea behind the step is very neat, but I agree that the controls need to be more intuitive.
[...]

PR #1030, opened 2017-11-14
Add "ORKWebViewStep"

chrisnowak:
This adds a simple ORKWebView step which allows you to embed a WKWebView in a task, using either a url or embedded html.  The embedded html can pass results back to the native ResearchKit code using Javascript.

## Background
While full native implementations of steps are in most cases preferable (and the only option in some cases to be able to leverage things like HealthKit) - we have encountered a few use cases where the ability to embed a WebView in a step really could be useful:
1.  Research instruments which either are too specific to a particular application to warrant a first-class native implementation (and contribution to ResearchKit), or they are still in the prototyping stages so a full native implementation would be premature (but there's still desire to try them out)
2.  We have had requests for the ability to embed highly custom content - which would be satisfied if we could just embed a webview
3.  Using a webview also allows for these instruments to be used in cross platform - the same (or similar) code for a step/instrument could be leveraged for Android or Web based surveys

For instance - here's a body site map developed as HTML/CSS/Javascript, embedded in a ResearchKit survey:

screen shot 2017-11-14 at 10 03 12 am

Everything below the "Step 1 of 3" is HTML/CSS/Javascript.  

## Implementation

I attempted to keep this as simple as possible - but would very much be open to feedback / ideas for improvement.  

Currently the WebViewStep can take either embedded HTML or an external url.  The WebView itself is set up using a WKUserContentController / script message hander to respond to "ResearchKit" messages from javascript.  In order for the user to proceed, this line of javascript must be executed:
```
window.webkit.messageHandlers.ResearchKit.postMessage(answer);  //answer is the string result of the step
```

There is currently no support for restoring answers - so if the user goes back to the step, they will have to fill it out again before proceeding.  It might be possible to accomplish restoring answers by establishing that there must be a javascript function called "restoreAnswers(string)" or something - and once the webview loads we could call that function natively with the answer they had previously supplied.  I opted to leave that for the future, but could look into it further if necessary.

Could be a solution for the folks who proposed:
https://github.com/ResearchKit/ResearchKit/issues/767
https://github.com/ResearchKit/ResearchKit/issues/367

It could also be interesting in the future if an interstitial webview step with javascript could be used to accomplish: https://github.com/ResearchKit/ResearchKit/issues/873 - by passing the current survey results to the embedded view, which could evaluate a "score" via javascript.  (Though again - not possible from this PR - would require the code to pass the results to the webview).

As part of this PR I added this to both ORKTest and ORKCatalog, with a simple embedded html example.


kelseydedoshka-apple:
@chrisnowak - this is a great idea. I think having the ability to add custom content via HTML/CSS/JavaScript would be really beneficial to a lot of developers. I would love to chat a bit further with you about the URL approach however.

I have some high level concerns about allowing developers to point their apps to a URL as the content contained in the URL can be dynamic and unfortunately means that it could be manipulated post app creation/consent/etc. In theory this might allow developers to abuse the system. I am curious your thoughts on this? On the flip side, I think the html approach is fantastic as it is static and will be baked into the application.

chrisnowak:
@kelseydedoshka-apple Thanks for taking a look! I'm glad this resonates.

Could you expand on the situation you are imagining occurring - how do you imagine a developer might abuse the system?   I had pondered that for a while but I couldn't actually see a (plausible) scenario by which it could be "abused"...many, many, apps out there use embedded WebViews to render content.  

While having the externalized content definitely introduces another moving part - that's also part of what makes this powerful.  A great example could be patient education resources - one could introduce a highly dynamic step which incorporates access to entire databases of educational materials, without having to actually have that content packaged with the app (meaning - no app store release needed to update that content).  

Absolutely agree that (when possible) the embedded html approach is preferable - as it doesn't require an internet connection and it is nice to have the survey be entirely self-contained.  However - we see (and have encountered) some profound use cases where that might not be possible...

kelseydedoshka-apple:
@chrisnowak - thanks for the additional information and the use cases you described help provide some additional understanding!

To try and clarify my concern a little bit more (and the case of this happening is very slim, but unfortunately still plausible). A developer could initially use a web view to display informational or educational content like you mentioned above (which I think is a great use case by the way), to then later change that URL to display something like advertisements or irrelevant content, which can lead to a poor user experience or even user misrtust. It is more of a precautionary concern, similar to how the App Store Review requires a higher age rating if you have a view that can display anything from the web (as we have no way to determine or screen what that content may be). Let me know if the helped clarify at all, and as we discuss this I am going to get some additional opinions on this as well as I could be over thinking it here.

As I mentioned before I am a big fan of this idea and would love to see it as a part of the framework, I just want to make sure all use cases are considered. Would you be interested in contributing this step with just the HTML property and not the URL?

chrisnowak:
@kelseydedoshka-apple sure - if the externalurl will hold this up, I am ok with refactoring to just have it support the embeddedhtml property.  I will go ahead and make that change and fix the current conflicts with master as well.

kelseydedoshka-apple:
@chrisnowak - great, I appreciate your flexibility here and looking forward to seeing the refactor so we can get this into the framework!

chrisnowak:
@kelseydedoshka-apple no problem - I have updated the code to remove the url property on WebViewStep, and fixed the conflicts with master.  Thanks!

kelseydedoshka-apple:
@chrisnowak - thanks for getting this updated so quickly. Great contribution!

PR #1031, opened 2017-11-16
[Misc] Fix unit tests

rsanchezsaez:
This PR fixes three separate unit test failures:

- Localization recorder failure. The `ork_doubleEqual()` method had too high of a precision, and failed for the *latitude* double floating point rounding errors when boxing and unboxing into `NSNumber`. Original value:  *37.3131699999999995043*, compared value: *37.313169999999978188*, difference: *2.13162820728030055761e-14*, minimumDifference: *1.65703761823010543768e-14*. I increased the K value (units of errors in the last place) from *1* to *2*.
  
screenshot 2017-11-15 23 10 38

- `ORKESerializationTests` failure due to not handling the new `scale` property of `ORKNumericAnswerFormat`. I took the chance to rename `scale` to `maximumFractionDigits` which is coherent with the `NSNumberFormatter` property.
- `ORKESerializationTests` failure due to not handling the new `style` and `vertical` property of `ORKImageChoiceAnswerFormat`.

### DIFF
--- a/ResearchKit/Common/ORKAnswerFormat.h
+++ b/ResearchKit/Common/ORKAnswerFormat.h
@@ -729,7 +729,7 @@ ORK_CLASS_AVAILABLE
  
  @return An initialized image choice answer format.
  */
-- (instancetype)initWithImageChoices:(NSArray<ORKImageChoice *> *)imageChoices NS_DESIGNATED_INITIALIZER;
+- (instancetype)initWithImageChoices:(NSArray<ORKImageChoice *> *)imageChoices;
 
 /**
  Returns an initialized image choice answer format using the specified array of images.
--- a/ResearchKit/Common/ORKAnswerFormat.h
+++ b/ResearchKit/Common/ORKAnswerFormat.h
@@ -1080,17 +1080,37 @@ Returns an initialized numeric answer format using the specified style, unit des
  
  This method is the designated initializer.
  
- @param style       The style of the numeric answer (decimal or integer).
- @param unit        A string that displays a localized version of the unit designation.
- @param minimum     The minimum value to apply, or `nil` if none is specified.
- @param maximum     The maximum value to apply, or `nil` if none is specified.
- 
+ @param style                   The style of the numeric answer (decimal or integer).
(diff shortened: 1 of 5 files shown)


kelseydedoshka-apple:
@rsanchezsaez - thanks so much for fixing these, overall looks great. Out of curiosity what was your reasoning for renaming the scale property to maximumFractionDigits?

rsanchezsaez:
@kelseydedoshka-apple The idea was to use a similar name as the `maximumFractionDigits ` property on `NSNumberFormatter` [(docs)](https://developer.apple.com/documentation/foundation/nsnumberformatter?language=objc#dash_1412008). The meaning of this property within the `ORKNumericAnswerFormat` context is "*the maximum number of fraction digits displayed on the label*". I though `scale` was a bit ambiguous (it also can be confused with the unrelated scale answer format).

---

Thinking more about it, and given that this doesn't alter the actual answer but only how it's displayed, I'd further rename to `maximumDisplayedFractionDigits`, if you are ok with this I'll do it tonight.

kelseydedoshka-apple:
@rsanchezsaez - definitely appreciate wanting to be less ambiguous with our property names, just also like to be cautious of sometimes unnecessary name changes as those will require developers already using it to update it. But in this case I like your proposal! Would love to see the updated version you proposed.

rsanchezsaez:
@kelseydedoshka-apple Actually, I now see that `ORKContinuousScaleAnswerFormat` also has a `maximumFractionDigits` parameter which behaves similarly to `ORKNumericAnswerFormat`'s one, so let's keep this name.

I agree that breaking backwards compatibility is not ideal, but in this case, `scale` has not been part of any "official" ResearchKit releases yet (it was not included in https://github.com/ResearchKit/ResearchKit/releases/tag/1.5.3), so I think it's fine to rename. Also, I kind of follow the Swift-team philosophy of breaking backwards compatibility in the spirit of a cleaner API interface. ;-)

kelseydedoshka-apple:
@rsanchezsaez - all great points, thanks for getting these fixed up!

PR #1035, opened 2017-11-28
Fix for ORKFormStepViewController getting stuck in a "skipped" state

mmertsock:
When the user taps the Skip button on a form step, ORKFormStepViewController [sets a private `_skipped` field to `YES`](https://github.com/ResearchKit/ResearchKit/blob/master/ResearchKit/Common/ORKFormStepViewController.m#L739). It [checks the value](https://github.com/ResearchKit/ResearchKit/blob/master/ResearchKit/Common/ORKFormStepViewController.m#L702) of `_skipped` whenever it needs to produce an ORKStepResult.

This can cause the form step to get into a bad state in the following scenario:

1. User taps Skip
2. Forward navigation is prevented, for example by an `ORKTaskViewControllerDelegate` instance implementing `taskViewController:shouldPresentStep:` and returning `NO`

At this point, the user is still viewing the current form step, but `_skipped` is still true. Thus, all of the results for that step will remain empty, even if the user attempts to change their answers (because the `result` property getter ignores any entered answers if `_skipped` is true).

In this pull request, I added some code to reset `_skipped` to NO when the answers are updated in the form step. I also added a sample form step to the `ORKTest` app to demonstrate this.

### DIFF
--- a/ResearchKit/Common/ORKFormStepViewController.m
+++ b/ResearchKit/Common/ORKFormStepViewController.m
@@ -392,6 +392,7 @@
         }
     }
     
+    _skipped = NO;
     [self updateButtonStates];
     [self notifyDelegateOnResultChange];
 }
@@ -914,6 +915,7 @@
             [self removeAnswerForIdentifier:formItemIdentifier];
         }
         
+        _skipped = NO;
         [self updateButtonStates];
         [self notifyDelegateOnResultChange];
     }
@@ -981,6 +983,7 @@
         [self removeAnswerForIdentifier:cell.formItem.identifier];
     }
     
+    _skipped = NO;
     [self updateButtonStates];
     [self notifyDelegateOnResultChange];
 }
--- a/Testing/ORKTest/ORKTest/Tasks/TaskFactory+Forms.m
+++ b/Testing/ORKTest/ORKTest/Tasks/TaskFactory+Forms.m
@@ -686,7 +686,31 @@
     }
     
     {
-        ORKFormStep *step = [[ORKFormStep alloc] initWithIdentifier:@"step5" title:@"Optional Form Items" text:@"Required form with no required items"];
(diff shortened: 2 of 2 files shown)


kelseydedoshka-apple:
@mmertsock - this is a great find, and thanks for adding an example to the Test App. Your solution looks great I am curious however if you have thought at all about how it seems odd from a use case perspective to allow the user to skip but then block them from proceeding? I am curious if you came across this bug as this was functionality you required? And if so I would be curious what that is?
 
There is something that bothers me about that specific user experience and trying to think of a solution similar to what you have that might help improve that a bit. Thoughts?

mmertsock:
@kelseydedoshka-apple You're right that this does look a little odd, especially with the (somewhat contrived) example. We have a system for setting up arbitrarily complex validation rules, managed by the survey author, that are executed via the `taskViewController:shouldPresentStep:` delegate method. We came across this behavior when testing the various edge cases of these validation rules.

chrisnowak:
@kelseydedoshka-apple Yeah, as @mmertsock mentioned, this just came up in testing.  Agree that it's strange to allow skip and then bar the user from hitting the button - we actually debated whether the right move would be to just always allow "Skip" or to fix this bug in ResearchKit.

One example of where this *might* make sense is that you may want to validate that the user has at least answered just *one* of the 5 previous questions before they are allowed to proceed.  So in this case - it would be valid for each of the questions to have a "skip" button, but the user would be barred from skipping the last question if they haven't answered at least one of the previous questions.  That being said, it's not a real world use case we've actually encountered.

kelseydedoshka-apple:
@mmertsock, @chrisnowak - thanks for the additional information here. Interesting example Chris and I could see how that might find its way into a use case. I appreciate you both taking the time to fix this via the framework in case others come across it at some point!

PR #1060, opened 2018-02-06
Add Go-no-go active task

syoung-smallwisdom:
Developed by Roland Rabien and the BiAffect research team. Submitted by [Sage Bionetworks](http://www.sagebase.org) in collaboration with [BiAffect](http://www.biaffect.com).

In a go-no-go task, the participant is asked to move the device sharply in any direction in response to a visual cue. You can use this task to accurately assess the participant's simple reaction time. Randomly they will be given a cue to not shake the device; this tests impulsiveness.

A go-no-go time task finishes when the participant has completed the required number of attempts successfully. For a "go", an attempt is successful when the participant exerts acceleration greater than `thresholdAcceleration` to the device after the stimulus has been delivered and before `timeout` has elapsed. An attempt is unsuccessful if acceleration greater than `thresholdAcceleration` is applied to the device before the stimulus or if this does not occur before `timeout` has elapsed. If unsuccessful, the result is not reported and the participant must try again to proceed with the task. For a "no go", the sucess condition is reversed.

Data collected by the task is in the form of `ORKGoNoGoResult` objects. The researcher can use these results to evaluate the response to the stimulus, and calculate the reaction time.

Note: Also included is a fix for the `ORKTest` serialization unit tests for the `attributedDetailText` property on `ORKInstructionStep`.


srinathtm-apple:
Hey @syoung-smallwisdom thank you so much for adding this!

A few questions, so I tried running the app and using the go-no-go active -
1) Does the user need to know parameters like "last reaction time" and "success/failure" ? Won't this influence their behavior? Can't we hide these from the user and return them in ORKGoNoGoResult?
2) It took me a while to understand how this works, I kept shaking the phone even after the check mark appeared in the blue circle which lead to a failure. Can we think about adding a different countdown timer-like animation?

syoung-smallwisdom:
@srinathtm-apple  Thanks for taking a look.

Unfortunately, I don't have an immediate answer for you. This is a task module being used by BiAffect in one of their studies. It was developed over a year ago. (I had previously submitted this on their behalf and then closed the PR last Fall due to inactivity and assorted merge issues.) Since the research scientist who developed this was a graduate student, I am unsure if he is still working with the group. I will confirm who is the current owner and whether or not the module has been validated by their research team (which could affect their willingness to make UI changes).

I will ping you again once I have contacted the BiAffect team.

srinathtm-apple:
@syoung-smallwisdom sounds good!

I also had a follow up question - the user will start shaking the device based on a visual cue, but if they are moving the phone too vigorously - how will they be able to stop shaking based on another visual cue? Is it okay to provide the 'stop shaking' cue using count down UI + haptics + sound ?

syoung-smallwisdom:
Closing. The team that was working on this task has opted not to update it.

Issue #1056, opened 2018-01-30
Picker Views doesnt support Dynamic Type fonts


ninoguba:
I found that question types that use the UIPickerView doesnt support Dynamic Type fonts, unlike the rest.

In iOS Settings > General > Accessibility > Larger Text, one can change the size of the font based on their preference, i.e for the vision impaired.

<img width="551" alt="screen shot 2018-01-29 at 3 40 51 pm" src="https://user-images.githubusercontent.com/4361804/35540594-77f9a81e-050b-11e8-88a9-030fae3a8b2d.png">

This is how it looks when using larger text.

<img width="551" alt="screen shot 2018-01-29 at 3 40 27 pm" src="https://user-images.githubusercontent.com/4361804/35540715-172ac4d6-050c-11e8-87ea-d1b7e1f4fee8.png">

This is how it looks when using the normal text size.

<img width="551" alt="screen shot 2018-01-29 at 3 46 20 pm" src="https://user-images.githubusercontent.com/4361804/35540734-33ea2cce-050c-11e8-9ae2-3fab58bfdd40.png">

As you can tell, the text of the UIPickerView didnt change size at all.  I expect it to be similar to the adjusted size of the ORKTextChoices below:

<img width="551" alt="screen shot 2018-01-29 at 3 40 09 pm" src="https://user-images.githubusercontent.com/4361804/35541064-c8d1585c-050d-11e8-9a0f-808476bc3e67.png">

Anyone have dealt with this issue before?

ninoguba:
This is a start for the solution.  In ORKValuePicker.m, I had to implement viewForRow of the UIPickerViewDelegate like below:

```
- (UIView *)pickerView:(UIPickerView *)pickerView viewForRow:(NSInteger)row forComponent:(NSInteger)component reusingView:(UIView *)view
{
    UILabel* valueLabel = (UILabel*)view;
    if (!valueLabel)
    {
        valueLabel = [[UILabel alloc] init];
        [valueLabel setFont:[self defaultFont]];
        [valueLabel setTextAlignment:NSTextAlignmentCenter];
    }
    // Fill the label text here
    valueLabel.text = [self pickerView:pickerView titleForRow:row forComponent:component];
    return valueLabel;
}

- (UIFont *)defaultFont {
    UIFontDescriptor *descriptor = [UIFontDescriptor preferredFontDescriptorWithTextStyle:UIFontTextStyleBody];
[...]

PR #1062, opened 2018-02-10
Added support for Dynamic Types to answer formats that use picker views

ninoguba:
This is the corresponding PR to issue #1056.

### DIFF
--- a/ResearchKit/Common/ORKMultipleValuePicker.m
+++ b/ResearchKit/Common/ORKMultipleValuePicker.m
@@ -44,6 +44,8 @@
 
 @end
 
+static const CGFloat PickerSpacerHeight = 15.0;
+static const CGFloat PickerMinimumHeight = 34.0;
 
 @implementation ORKMultipleValuePicker {
     UIPickerView *_pickerView;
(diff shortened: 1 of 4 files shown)


ResearchKit/Common/ORKMultipleValuePicker.m
```diff
@@ -251,6 +253,30 @@ - (NSString *)pickerView:(UIPickerView *)pickerView titleForRow:(NSInteger)row f
...
+{
+    UILabel* valueLabel = (UILabel*)view;
+    if (!valueLabel)
+    {
+        valueLabel = [[UILabel alloc] init];
+        [valueLabel setFont:[self defaultFont]];
+        [valueLabel setTextAlignment:NSTextAlignmentCenter];
+    }
+    valueLabel.text = [self pickerView:pickerView titleForRow:row forComponent:component];
+    return valueLabel;
```
syoung-smallwisdom: I just did this for Sage's [shared research framework](https://github.com/ResearchKit/SageResearch) and discovered that centering all the rows will result in really weird looking UI under some circumstances (eg. blood pressure, duration formatted for hours/minutes/seconds, month/year date entry, etc.). What we ended up doing was computing the size of all the labels, caching the sizes, and then special-casing the first and last components. Then the first component is formatted to right alignment and the last component is formatted to left alignment. Only the components in the center are center-aligned. We used calculated widths b/c if the picker is ever used for something like month/year (which is not supported by UIDatePicker), then the widths will not be even. We considered optimizing the code by using character count, etc (because calculated text bounds are expensive) but decided that doing so would be premature optimization.

https://github.com/ResearchKit/SageResearch/blob/master/ResearchSuiteUI/ResearchSuiteUI/iOS/Views/RSDPickerViewProtocol.swift

ninoguba: @syoung-smallwisdom the default alignment, prior to this PR, is center so I just implemented the same.

I agree with you that on some use cases changing the alignments will become necessary which can be done by the developer as they see fit for their exact application.

ninoguba:
I added the In Review label on behalf of @syoung-smallwisdom =)

ninoguba:
What should be the label now that this is merged?

Issue #1021, opened 2017-10-12
ORKImageCaptureStep camera viewport does not fill the width of the view frame


obianachebe:
When I create an ORKImageCaptureStep and display the ORKStep, the viewport is very tiny and does not fill the width of the view frame. Currently testing on an iPhone 6s+ running iOS 11.0.2

**What did you do?**

I created an ORKImageCaptureStep, then an ORKOrderedTask, and finally presented a taskViewController with the task as shown below:

```
let imageStep = ORKImageCaptureStep(identifier: "imageCapture")
imageStep.title = "Image Capture"

let task = ORKOrderedTask(identifier: "SurveyTask", steps: [imageStep])
let taskVC = ORKTaskViewController(task: task, taskRun: nil)

present(taskVC, animated: true, completion: nil)
```

**What did you expect to happen?**

I expected for the viewport to take up most of the frame

**What did you see instead?**

The viewport is very small

![img_3501](https://user-images.githubusercontent.com/13955058/31510933-f0a779f2-af3a-11e7-8f29-2905fd94f080.PNG)

Any help is much appreciated!

ninoguba:
You'll want to set your imageStep's templateImage and templateImageInsets to control the size of the camera viewport.  I know the API documentation says setting this is optional but I learned this is necessary to size the viewport to expectations.

obianachebe:
@ninoguba thanks for the reply. I've set the templateImageInsets as so:

`imageStep.templateImageInsets = UIEdgeInsets(top: 10, left: 10, bottom: 10, right: 10)`

This allows the viewport to better fill the frame, but the capture button is no longer visible (see below). Am I missing something?

![img_3503](https://user-images.githubusercontent.com/13955058/31514242-cf648694-af45-11e7-94b3-9dabb135c629.PNG)

ninoguba:
Try setting a templateImage that is transparent and sized to the frame size and aspect ratio you desire.

obianachebe:
@ninoguba I've set a templateImage and set the templateImageInsets as shown:

`imageStep.templateImageInsets = UIEdgeInsets(top: 0, left: 10, bottom: 0, right: 10)`

[...]

PR #1064, opened 2018-02-12
Make the ORKImageCaptureCameraPreviewView take the full width of the screen.

overcyn:
Previously the size of the camera capture view was dependent on the templateImage size. This is somewhat unintuitive https://github.com/ResearchKit/ResearchKit/issues/1021. Additionally this setup messed up the constraints in the ORKNavigationContainerView, causing it to hide the skip button, and take up most of the screen in landscape.

Previously:
screen shot 2018-02-12 at 11 42 47 am
screen shot 2018-02-12 at 2 32 04 pm

Now:
screen shot 2018-02-12 at 12 38 47 pm
screen shot 2018-02-12 at 2 33 01 pm

### DIFF
--- a/ResearchKit/Common/ORKImageCaptureCameraPreviewView.m
+++ b/ResearchKit/Common/ORKImageCaptureCameraPreviewView.m
@@ -78,6 +78,9 @@
 - (void)setUpConstraints {
     NSMutableArray *constraints = [NSMutableArray new];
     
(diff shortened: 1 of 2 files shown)


overcyn:
Oops did not mean to create this against actual ResearchKit.

ZaidSA:
@srinathtm-apple Can we have same fix for `ORKVideoCaptureStep`?
It looks like this:
img_4175

PR #1067, opened 2018-02-13
Show time-left-label for audio task

danielemden:
Probably due to a typo. The formatter which is used to format the string showing the remaining time for the audio task is always nil.


srinathtm-apple:
@danielemden thanks for submitting this. We'll get this merged once we have 2 approved reviews.

PR #1075, opened 2018-02-20
Timed walk task cleanup

srinathtm-apple:
1) deprecate an initializer to avoid duplicate code paths
2) use ORKText choice for picker view instead of NSString which was causing a run time crash

### DIFF
--- a/ResearchKit/Common/ORKOrderedTask+ORKPredefinedActiveTask.m
+++ b/ResearchKit/Common/ORKOrderedTask+ORKPredefinedActiveTask.m
@@ -1529,7 +1529,7 @@ NSString *const ORKTimedWalkTrial2StepIdentifier = @"timed.walk.trial2";
         }
 
         {
-            if (turnAroundTimeLimit >= 0) {
+            if (turnAroundTimeLimit > 0) {
                 ORKTimedWalkStep *step = [[ORKTimedWalkStep alloc] initWithIdentifier:ORKTimedWalkTurnAroundStepIdentifier];
                 step.title = ORKLocalizedString(@"TIMED_WALK_INSTRUCTION_TURN", nil);
                 step.text = ORKLocalizedString(@"TIMED_WALK_INSTRUCTION_TEXT", nil);
--- a/ResearchKit/Common/ORKOrderedTask+ORKPredefinedActiveTask.m
+++ b/ResearchKit/Common/ORKOrderedTask+ORKPredefinedActiveTask.m
@@ -1458,12 +1458,12 @@ NSString *const ORKTimedWalkTrial2StepIdentifier = @"timed.walk.trial2";
                                                             answerFormat:answerFormat1];
         formItem1.optional = NO;
 
-        NSArray *textChoices = @[ORKLocalizedString(@"TIMED_WALK_QUESTION_2_CHOICE", nil),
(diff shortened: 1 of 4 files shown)


ResearchKit/Common/ORKOrderedTask+ORKPredefinedActiveTask.m
```diff
@@ -1594,12 +1458,12 @@ + (ORKOrderedTask *)timedWalkTaskWithIdentifier:(NSString *)identifier
                                                             answerFormat:answerFormat1];
         formItem1.optional = NO;
 
-        NSArray *textChoices = @[ORKLocalizedString(@"TIMED_WALK_QUESTION_2_CHOICE", nil),
-                                 ORKLocalizedString(@"TIMED_WALK_QUESTION_2_CHOICE_2", nil),
-                                 ORKLocalizedString(@"TIMED_WALK_QUESTION_2_CHOICE_3", nil),
-                                 ORKLocalizedString(@"TIMED_WALK_QUESTION_2_CHOICE_4", nil),
-                                 ORKLocalizedString(@"TIMED_WALK_QUESTION_2_CHOICE_5", nil),
-                                 ORKLocalizedString(@"TIMED_WALK_QUESTION_2_CHOICE_6", nil)];
+        NSArray *textChoices = @[ [ORKTextChoice choiceWithText:ORKLocalizedString(@"TIMED_WALK_QUESTION_2_CHOICE", nil) value:ORKLocalizedString(@"TIMED_WALK_QUESTION_2_CHOICE", nil)],
```
rsanchezsaez: Would it make sense to introduce some hardcoded English identifiers for the `value` fields, so the values in the results don't vary between user localizations? For example, if the same study is presented to users with different locales, you'd want the results to be consistent among users, rather than being localized.

srinathtm-apple: Good catch! Absolutely, I will make a change to reflect just the key value.

rsanchezsaez: Not to be annoying, but would it make sense to change the keys to something more descriptive in the result context such as
```
"TIMED_WALK_QUESTION_2_MODE_NONE" = "None";
"TIMED_WALK_QUESTION_2_MODE_UNILATERAL_CANE" = "Unilateral Cane";
"TIMED_WALK_QUESTION_2_MODE_UNILATERAL_CRUTCH" = "Unilateral Crutch";
"TIMED_WALK_QUESTION_2_MODE_BILATERAL_CANE" = "Bilateral Cane";
"TIMED_WALK_QUESTION_2_MODE_BILATERAL_CRUTCH" = "Bilateral Crutch";
"TIMED_WALK_QUESTION_2__MODE_WALKER" = "Walker/Rollator";
```
?

If you think this is unnecessary, my previous approval stands. :-)

srinathtm-apple: No, I completely agree. Do you think changing this might affect existing existing apps? What if someone is using the same keys in some other task? Do we keep the old keys and add new ones ? (do we care about the duplication involved?) 

Would love to hear your thoughts on this...

srinathtm-apple: @rsanchezsaez - any further thoughts about this?

PR #1078, opened 2018-02-21
convenience initializers for active tasks

srinathtm-apple:
** Request for initial review **

As we keep adding more active tasks, it might be helpful to provide convenience initializers that accept NSData (read - JSON payload) or NSDictionary objects with specific key-value pairings that can help create instances of the corresponding active tasks.

For example, the following blob when passed to the convenience initializer will return an instance of trailmaking active task -

{
    "taskType": "trailmaking",
    "identifier": "trailmaking_task_sample_0202",
    "intendedUseDescription": "Tests visual attention and task switching",
    "trailType": "B"
}

### DIFF
--- a/ResearchKit/Common/ORKOrderedTask+ORKPredefinedActiveTask.h
+++ b/ResearchKit/Common/ORKOrderedTask+ORKPredefinedActiveTask.h
@@ -35,11 +35,10 @@
 
 @class ORKNavigableOrderedTask;
 
-
+NS_ASSUME_NONNULL_BEGIN
 
 @interface ORKOrderedTask (ORKPredefinedActiveTask)
 
-NS_ASSUME_NONNULL_BEGIN
 
 /**
  Returns a predefined task that measures the upper extremity function.
@@ -635,11 +634,10 @@ NS_ASSUME_NONNULL_BEGIN
                                         trailType:(ORKTrailMakingTypeIdentifier)trailType
                                           options:(ORKPredefinedTaskOption)options;
 
-NS_ASSUME_NONNULL_END
-
 + (nullable ORKOrderedTask *)initializeActiveTaskFromData:(NSData *)data;
 
 + (nullable ORKOrderedTask *)initializeActiveTaskFromDictionary:(NSDictionary *)dictionary;
 
 @end
 
+NS_ASSUME_NONNULL_END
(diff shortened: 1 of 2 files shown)


ResearchKit/Common/ORKOrderedTask+ORKPredefinedActiveTask.h
```diff
@@ -635,6 +635,11 @@ NS_ASSUME_NONNULL_BEGIN
                                         trailType:(ORKTrailMakingTypeIdentifier)trailType
                                           options:(ORKPredefinedTaskOption)options;
 
+NS_ASSUME_NONNULL_END
```
syoung-smallwisdom: Why move the `NS_ASSUME_NONNULL_END` ? You can also just define your methods as nullable and it will be fine without moving the wrapper macro.

srinathtm-apple:
@syoung-smallwisdom agreed, I can definitely move all hard-coded string to constants (I don't favor macros a lot since they can't be referenced in the debugger)

What do you think is the best way to establish a schema? In my current approach, I have simply leveraged the function signature and a top level key to denote the activeTask type which is basically the function name. 

Yup, unit tests to follow!

ResearchKit/Common/ORKOrderedTask+ORKPredefinedActiveTask.h
srinathtm-apple: My bad, I didn't start off by explicitly specifying the return type as nullable. I will revert this.

rsanchezsaez:
I kind of agree with @syoung-smallwisdom that this introduces type unsafely and the possibility of typos. What's the use case for adding a general purpose initializer like this? Is some kind of general purpose research app which can dynamically support new task types in mind? Is there something similar to this anywhere else on Apple's iOS API?

Personally, I prefer to stick to typed initializers when possible, they seem more straight forward. :-)

syoung-smallwisdom:
Here's how we were doing it:

https://github.com/Sage-Bionetworks/ResearchUXFactory-iOS/blob/master/ResearchUXFactory/SBAActiveTask.swift

The factory framework was designed to allow us to create tasks from a protocol that uses either a dictionary (JSON file) or objects defined by our services, but the general idea is the same. The code pre-dates swift `Codable` protocol but the idea is the same.

This is an older Obj-c example that uses KVO to create objects from dictionaries. https://github.com/Sage-Bionetworks/ResearchUXFactory-iOS/blob/master/ResearchUXFactory/SBADataObject.m

Issue #1071, opened 2018-02-16
Consent signature/date layout issues


mirlhaq:
There are some issues with the consent signature and dates layout, in both, the review step and the generated PDF.

Here are some screenshots:

![signature issue screenshot 1](https://user-images.githubusercontent.com/570932/36326215-20c7e4a4-130f-11e8-9afb-3eda87d7d155.png)

iPhone X:
![signature issue screenshot 2](https://user-images.githubusercontent.com/570932/36326216-20e88f06-130f-11e8-8a21-de66bd309f28.jpeg)

The fix is super simple. Just an escaped `%` (`%%`) that doesn't need to be escaped in the CSS.

windenchanter:
Just a comment. I tried your fix, and it made the Participant name and Date sunk into the "Participant's Name (printed)" and "Date". So they now overlap partially.

mirlhaq:
Hi @windenchanter thanks for checking! What device are you on? This worked for an iPhone 7 plus over here.

windenchanter:
I was using an iPad 5.
The pdf file is generated so i suppose it would not be device sensitive?

Fernando Waigandt <notifications@github.com> 于 2018年2月21日周三 06:42写道：

mirlhaq:
The pdf is generated from an HTML `NSString`, and that part is dealing with relative css layout/style. It is screen size sensitive. We need to find a better way to layout the signatures and dates.

srinathtm-apple:
#1081 @windenchanter @mirlhaq can you folks please review this to see if it works? Seems to work as expected on iPhone and iPad simulators for me.

mirlhaq:
@srinathtm-apple I can confirm this works ok on an iPhone 7 plus device.

srinathtm-apple:
Closing this issue since the fix has been merged into master.

majeedyaseen:
After applying the fix the added signatures are not showing up and name and dates are aligned correctly, 

[css appendString:@".inbox { position: **relative**; bottom: 0; left:0; top: 100%%; transform: translateY(-100%%); -webkit-transform: translateY(-100%%);  }\n"];

making position relative and keeping everything else same for the merged CSS lines. The signature showed up but the name again went up

PR #1081, opened 2018-02-21
fix for #1071

srinathtm-apple:


majeedyaseen:
After applying the fix the added signatures are not showing up and name and dates are aligned correctly, 

[css appendString:@".inbox { position: **relative**; bottom: 0; left:0; top: 100%%; transform: translateY(-100%%); -webkit-transform: translateY(-100%%);  }\n"];

making position relative and keeping everything else same for the merged CSS lines. The signature showed up but the name and date again went up.

PR #1088, opened 2018-03-05
Changes to angle calculations in Range of Motion active tasks

davwillev:
- Added quaternion to Euler definitions for both Roll and Yaw. These will be needed for future Range of Motion tasks, beyond sagittal rotation.
- Replaced the attitude.roll implementation for when the device in landscape orientation with the quaternion for Roll.
- Modified the calculation for 'flexed' and 'extended' angles.


srinathtm-apple:
@davwillev thanks a bunch for submitting this!

The code looks fine to me, but I am totally gonna trust you with the math. Will merge this in once we have another reviewer approve this change.

davwillev:
@srinathtm-apple I have managed to implement the above equations in some tasks that I am developing. Having tested the angles produced, I would like to amend one of the equations in the above pull requestvery slightly (change a '+' sign for a '-' sign) to this:

`#define allOrientationsForYaw(x, w, y, z) (asin(2.0 * (x*y - w*z)))`

ResearchKit/ActiveTasks/ORKRangeOfMotionStepViewController.m
```diff
@@ -44,8 +44,14 @@ OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 
 
 #define radiansToDegrees(radians) ((radians) * 180.0 / M_PI)
+
 #define allOrientationsForPitch(x, w, y, z) (atan2(2.0 * (x*w + y*z), 1.0 - 2.0 * (x*x + z*z)))
 
+#define allOrientationsForRoll(x, w, y, z) (atan2(2.0 * (y*w - x*z), 1.0 - 2.0 * (y*y + z*z)))
+
+#define allOrientationsForYaw(x, w, y, z) (asin(2.0 * (x*y + w*z)))
```
rsanchezsaez: Could you change all of this defines to inline functions? It's safer.

davwillev: Hi @rsanchezsaez 
Sorry, I'm not sure what you mean... could you make these changes so that I can see... ?
Best wishes

ResearchKit/ActiveTasks/ORKRangeOfMotionStepViewController.m
```diff
@@ -207,12 +218,15 @@ - (ORKResult *)result {
     ORKStepResult *stepResult = [super result];
     
     ORKRangeOfMotionResult *result = [[ORKRangeOfMotionResult alloc] initWithIdentifier:self.step.identifier];
-    result.flexed = _flexedAngle;
-    result.extended = result.flexed - _rangeOfMotionAngle;
     
+    result.flexed = 90.0 - _flexedAngle;
+    result.extended = _rangeOfMotionAngle - result.flexed;
+
```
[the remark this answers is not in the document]
davwillev: I think I've now done this successfully

ResearchKit/ActiveTasks/ORKRangeOfMotionStepViewController.m
davwillev: Hi @rsanchezsaez, how can we get these requested changes finalised so that this pull request can be approved? Thanks

ninoguba:
These are the screenshots I find redundant...

**Addition to Knee RoM (suggesting removal):**

screen shot 2018-07-30 at 1 23 01 pm

**Original to Knee RoM (keep as is):**

screen shot 2018-07-30 at 1 23 19 pm
screen shot 2018-07-30 at 1 23 23 pm

Similarly the same for the Shoulder RoM:

**Addition to Shoulder RoM (suggesting removal):**

screen shot 2018-07-30 at 1 23 59 pm

**Original to Shoulder RoM (keep as is):**

screen shot 2018-07-30 at 1 24 03 pm
screen shot 2018-07-30 at 1 24 07 pm

Thanks.

davwillev:
Hi @ninoguba
I added the spoken instruction to the 'TouchAnywhereStep' because when we were testing these tasks with volunteers, they kept removing the phone from their shoulder/knee to read the instructions and look at the picture. The additional spoken instructions helped reduce this.
By the same token, I added the 'phonesoundwave_inverted' image to improve the look of the instruction that asks the participant to ensure their sound is switched on. In the original version of the tasks, this is the only instruction without an image.
These were considered enhancements to the task instructions, but happy to remove them from this PR if it helps focus the improvements to angle recording/results.
Just so you know, I'm not creating my own app, but wanted to improve these tasks for wider research use. I also want to add new ROM tasks for other movements to the RK framework at a later date, and these enhancements would be part of these.

Regarding angles, I agree - the minimum angle should not always be zero, but instead should record the 'largest' angle in the direction opposite to the maximum (and should therefore be negative in value). Any ideas what might be causing this?

Thanks!

ninoguba:
Now I understand.  Perhaps the sample image in this screenshot is more representative of "turn the sound on."  But I would defer the design decision to the folks at Apple. @srinathtm-apple please review.

screen shot 2018-07-30 at 2 34 37 pm


To fix the minimum, change the code in ORKRangeOfMotionStepViewController.m to the following:

```
    if (_minAngle == 0.0 || _rangeOfMotionAngle < _minAngle) {
        _minAngle = _rangeOfMotionAngle;
    }
```

davwillev:
Thank you, @ninoguba I have updated your fix for the minimum angle.

Looking through other active tasks, the norm is to accompany instructions with images. The reason that I chose the 'turn the sound on' image was to keep consistency to [existing images in the RK framework](https://github.com/ResearchKit/ResearchKit/tree/master/ResearchKit/Artwork.xcassets/Active/phonesoundwaves.imageset), but I do agree that the image you chose is intuitive.

Thanks again!

ninoguba:
I suggest that you remove the "waveform" image modification in this PR so we can streamline the review by focusing just on the code-related enhancements.

davwillev:
Hi @ninoguba 
I've removed the 'phonesoundwaves' image modification, as requested

davwillev:
@ninoguba 
Just tested out the fix for minimum, and unfortunately the result still doesn't quite record properly. Here's a screenshot where I started and finished the knee task with the phone upright in portrait (approx. 0 degrees), and tried to get a maximum of +90, and a minimum of -90:

img_4616

I tried a few other tests and the minimum always seems to result as a number between 0 and 1. Any ideas where the problem could lie, since maximum is working perfectly?

ninoguba:
Hi @davwillev currently the minimum is the lowest  _rangeOfMotionAngle value recorded so it is highly dependent on how the _rangeOfMotionAngle calculation is implemented in this function.  If it's not correct then the issue is in here:

```
- (void)calculateAndSetAngles {
    _startAngle = fabs([self getDeviceAngleInDegreesFromAttitude:_referenceAttitude]);
    
    BOOL rangeOfMotionMoreThan180Degrees = _highestAngle > 180 && _lowestAngle <= 180;
    if (rangeOfMotionMoreThan180Degrees) {
        _rangeOfMotionAngle = 360 - fabs(_lastAngle);
    } else {
        _rangeOfMotionAngle = fabs(_lastAngle);
    }
    if (_rangeOfMotionAngle > _maxAngle) {
        _maxAngle = _rangeOfMotionAngle;
    }
    if (_minAngle == 0.0 || _rangeOfMotionAngle < _minAngle) {
        _minAngle = _rangeOfMotionAngle;
    }
}
```

Or possibly in how the _highestAngle, _lowestAngle, _lastAngle are calculated in the corresponding functions.   Please look into those.

davwillev:
OK, @ninoguba 
I'll try a few iterative tweaks and chat to a couple of colleagues about the various functions. This might take a few weeks due to vacations though.
@srinathtm-apple is there any chance of getting some Core Motion folks to take a look in the meantime, considering we are nearly there with these changes?
Thanks to all you guys for spending time on this.

davwillev:
After plenty of tests, there are a few problems:

**1).** When device motion is either max > +180 or min < -180, then BOTH maximum and minimum results default to +179.9 and -179.9 respectively. I think this is simply due to the default range of attitude data from Core Motion (i.e. from 0 to +/- pi radians, or +/-180 degrees). For example, if the angle moved during the ROM task is beyond +180, the device will next register -179.9 etc, and this is immediately recorded by our new code as the minimum result (as no smaller output can be measured).

I think the existing Boolean function (below) is supposed to deal with this default range with `360 - fabs(_lastAngle)`, but there are a couple of problems in using this original version for our new calculations. Firstly, the `_lastAngle` Boolean will currently only modify the angle values if BOTH `_highestAngle` AND `_lowestAngle` reach their respective thresholds (originally >175 and <175). To adjust the negative minimum value, I think the latter should have been <-175.

```
BOOL rangeOfMotionMoreThan180Degrees = _highestAngle > 175 && _lowestAngle < 175;
    if (rangeOfMotionMoreThan180Degrees) {
        _rangeOfMotionAngle = 360 - fabs(_lastAngle);
    } else {
        _rangeOfMotionAngle = _lastAngle;
    }
```

We can easily change the logic to an OR function but then we will all too often have a correct maximum but an incorrect minimum (or vice versa), since `_lastAngle` is used to calculate both results. The finish angle is also calculated using _rangeOfMotionAngle, so if the Boolean is triggered, then the finish angle will also be adjusted. Hence, I tried to separate the `360 - fabs(_lastAngle)` adjustments out into two Boolean functions (and I changed +/-175 to +/-179.):

```
    BOOL rangeOfMotionMoreThanPlus180Degrees = _highestAngle > 179;
    if (rangeOfMotionMoreThanPlus180Degrees) {
        _rangeOfMotionAngle = 360 - fabs(_lastAngle);
    } else {
        _rangeOfMotionAngle = _lastAngle;
    }
    BOOL rangeOfMotionLessThanMinus180Degrees = _lowestAngle < -179;
    if (rangeOfMotionLessThanMinus180Degrees) {
        _rangeOfMotionAngle = fabs(_lastAngle) - 360;
    } else {
        _rangeOfMotionAngle = _lastAngle;
    }
```

Unfortunately, this is still producing unexpected results, which I think occur because _lastAngle is included in both Booleans. I'm not sure how to deal with this.


**2).** The current knee and shoulder range of motion tasks allow for the device to be rotated into either positive or negative pitch or roll. Our min / max calculations currently don't allow for this flexibility and will report different results depending on the orientation of the device (see below). We need to somehow adjust the angle of the device before calculation of max or min values; otherwise they will be the wrong way around.

axes of rotation


**3).** Since @ninoguba has pointed out to me how to calculate min / max values, I can't help notice that the original highest/lowest angle values (below) used the same approach. Do we still need this function, and if so, what role does it perform now, given the new max/min calculations?

```
double angle = [self getDeviceAngleInDegreesFromAttitude:currentAttitude];

    if (angle > _highestAngle) {
        _highestAngle = angle;
    }
    if (angle < _lowestAngle) {
        _lowestAngle = angle;
    }
    _lastAngle = angle;
```

davwillev:
Hi All,
A quick update - using the above changes, the module is now working perfectly with only the one problem remaining:
The minimum/maximum results will not adjust via the Boolean function if they pass beyond -180/180 degrees respectively. Interestingly, the finish (second screen tap) result does adjust correctly. Hence, we are very close to getting this working and I will try some options to see if I can get the min/max results to adjust.
Any help in the meantime will be greatly appreciated.
Thanks!

davwillev:
Hi All,
I've tried several ways to amend this, without success. The strategy that I have tried to implement is to shift the output angle range to -90 to +270 (from the original +/- 180), which should be able to cover all knee and shoulder ranges of motion.

```
if (angle < -90 && angle >= -180) {
   _newAngle = 360 - fabs(angle);
}
_newAngle = angle;
```

However, every time I try to implement this, I cannot seem to limit the function for the intended angle range; once triggered by any angle, it applies to all output measurements. Is there a way of limiting the function just to the specified angles (-90 to -180)?

davwillev:
Hi All,
@srinathtm-apple 
@rsanchezsaez 
@syoung-smallwisdom 
@ninoguba 
After spending far too much time on this, I have finally got this to work and have solved all of the problems that I highlighted previously; I believe that the module is now ready for the PR!

In summary:
- The CoreMotion range of +/- 180 degrees has been adjusted to -90 to +180 degrees, which will cover the range of motion for all knee and shoulder ranges of motion.
- The original Boolean function that was designed for adjusting the angle turned out to be unnecessary. This has been replaced by another Boolean that works for all calculated results: start, finish, minimum and maximum.
- The inbuilt CoreMotion attitude settings record positive and negative angles (via a quaternion) using the right-hand rule, whioch is in the opposite direction to that when knee and shoulder motion is recorded with the creen facing out. Hence, miminum and maximum results have had to be reversed. This reversal also applied to the angle range adjustment Boolean.
- By calculating the range of motion as the absolute difference between calculated maximum and minimum results, the tasks can be measured with the device in any orientation (as long as the same orientation is used throughout the task. Without doing so, the range could be reported as a negative value.
- Finally, as I explained previously the image labels 'flexed' and 'extended' had to be changed (to 'start' and 'maximum'), primarily because they were anatomically wrong for the shoulder (flexion and extension are the opposite in knees and shoulders). Apologies that this led to a large number of commits.

Thank you all for your patience with my many alterations and requests throughout this process. Particular thanks go to @ninoguba for showing me how to calculate min and max results. I look forward to feedback from you all!

ninoguba:
@davwillev sorry didnt realized I was the one blocking the PR. We'll just have to trust you with the calculations since you know it best.   Great job on this contribution to the RoM tasks!

ninoguba:
@srinathtm-apple please see if you merge this PR the soonest for @davwillev. thanks!

srinathtm-apple:
Merging into master. Thank you so much @davwillev !

Issue #1022, opened 2017-10-13
ORKNumericAnswerFormat doesn't show unitString on iOS 11


AlucarDWeb:
This is the code I use to create an ORKNumericAnswerFormat:

```SWIFT
let answerFormat = ORKNumericAnswerFormat(style: .integer, unit: (self.unitString ?? ""))
ORKFormItem(identifier: (self.answerId ?? "\(self.objectId)"), text: self.placeholder, answerFormat: answerFormat)
```
The screens attached shows the difference between iOS 10 and iOS 11. As you can see, the unitString doesn't appear on iOS 11. 
![simulator screen shot - iphone se - 2017-10-13 at 10 04 23](https://user-images.githubusercontent.com/6224848/31536689-d5138ac6-afff-11e7-8df9-f4403c3ec266.png)
![simulator screen shot 13 ott 2017 10 08 27](https://user-images.githubusercontent.com/6224848/31536690-d53039f0-afff-11e7-8ee1-534b4911e963.png)

Pariecemckinney-apple:
@AlucarDWeb Thank you for pointing this out. We will begin investigating today.

kelseydedoshka-apple:
@AlucarDWeb - I am still working through a final fix for this but I wanted to update you with progress to see if I can help unblock you here. It seems like the label displaying the suffix doesn't have the proper height and y values set on it's position and therefore isn't being displayed.

I have added the following two lines of code to the end of the **ork_suffixFrame** function in **ORKTextField.m**

suffixFrame.size.height = CGRectGetHeight(self.bounds);
suffixFrame.origin.y = 0;

So far it seems to be working for me. Let me know if this seems to work for you as well.

srinathtm-apple:
Closing this since the fix has been merged into master.

PR #1090, opened 2018-03-07
fix for #1022

srinathtm-apple:
fix for #1022


dukemike:
This worked great for me. Thanks!

PR #1134, opened 2018-07-13
Fixed position of ORK_CLASS_AVAILABLE

anaglik:
When building project with iOS SDK lower than 12, it causes compilation error because `ORK_CLASS_AVAILABLE` was expecting class which in fact was omitted by `#if defined(__IPHONE_12_0)...`

It was breaking compilation mostly for CocoaPods, which reads all headers specified in .podspec file.
Additionally `ResearchKit_Private.h` was simplified because validation is done directly in `ORKHealthClinicalTypeRecorder.h` file.

Thanks for help to @alistra


srinathtm-apple:
Thank for submitting this fix! Merging this PR.

PR #1144, opened 2018-07-25
Use fixed accessibility label for the completion view.

conorhughes-apple:
The completion view looks like it's now just always a check in a blue
circle. Update the label to a fixed string. This avoids interpolating
nil, which was happening because captionLabel is not visible here (and
as such moving focus to it with a screen change is also inappropriate).

Resolves rdar://problem/42294298. Tested on iPhone 8.


srinathtm-apple:
Merging changes by Accessibility team into master.

PR #1147, opened 2018-07-30
Use UIAccessibilityTraitStartsMediaSession on the Start Recording button, so that VoiceOver doesn't speak the button title when you activate it.

kasemset-apple:
Resolves rdar://problem/42294456. Tested on iPhone 7.


srinathtm-apple:

PR #1149, opened 2018-07-30
Make the Amsler Grid a direct touch area, so the task can be completed easily by low vision users using VoiceOver.

kasemset-apple:
Also give it an appropriate label and hint.

Resolves rdar://problem/42293953. Tested on iPhone 7.


srinathtm-apple:

PR #1150, opened 2018-07-30
Make the waveform accessible so that low vision users are aware that it is present.

kasemset-apple:
Resolves rdar://problem/42294396. Tested on iPhone 7.


srinathtm-apple:

PR #1152, opened 2018-07-31
Accessibility fixes for the transcript view in the Speech Recognition task.

kasemset-apple:
- Expose only one element for the text view.
- Expose a custom action to allow dismissing the keyboard.

Resolves rdar://problem/42294608. Tested on iPhone 7 with VoiceOver and Switch Control.


srinathtm-apple:

PR #1156, opened 2018-08-01
Improve usability of tests that use ORKRoundTappingButton.

kasemset-apple:
- In the tone audiometry test, the user has a chance to determine the location of the buttons during the practice step. Once the test begins, the buttons become directly tappable.
- In the two finger tapping interval test, the user can also determine the location of the buttons first. Then on the first activation of either button, the buttons become directly tappable, but the tap is not registered yet.
- In the dBHL tone audiometry test, the user is given instructions about the button and it is directly tappable.
- In the trailmaking test, the user is given instructions about the buttons and they are directly tappable.


srinathtm-apple:

PR #1161, opened 2018-08-16
Allow autoplay and inline media playback in ORKWebViewStep

overcyn:
Updates the webview configuration to allow inline media playback and autoplay.

### DIFF
--- a/ResearchKit/Common/ORKWebViewStepViewController.m
+++ b/ResearchKit/Common/ORKWebViewStepViewController.m
@@ -55,6 +55,11 @@
     
     if (self.step && [self isViewLoaded]) {
         WKWebViewConfiguration *config = [[WKWebViewConfiguration alloc] init];
+        config.allowsInlineMediaPlayback = true;
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
+        config.mediaPlaybackRequiresUserAction = false;
+#pragma clang diagnostic pop
         WKUserContentController *controller = [[WKUserContentController alloc] init];
         [controller addScriptMessageHandler:self name:@"ResearchKit"];
         config.userContentController = controller;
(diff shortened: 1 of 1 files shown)


ResearchKit/Common/ORKWebViewStepViewController.m
```diff
@@ -55,6 +55,11 @@ - (void)stepDidChange {
     
     if (self.step && [self isViewLoaded]) {
         WKWebViewConfiguration *config = [[WKWebViewConfiguration alloc] init];
+        config.allowsInlineMediaPlayback = true;
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
+        config.mediaPlaybackRequiresUserAction = false;
```
srinathtm-apple: Can we avoid using a deprecated API?

overcyn: Were you thinking something like the below, which would work on iOS 8+?

```
#pragma clang diagnostic push
#pragma clang diagnostic ignored "-Wdeprecated-declarations"
        if ([config respondsToSelector:@selector(mediaTypesRequiringUserActionForPlayback)]) {
            config.mediaTypesRequiringUserActionForPlayback = WKAudiovisualMediaTypeNone;
        } else if ([config respondsToSelector:@selector(requiresUserActionForMediaPlayback)]) {
            config.requiresUserActionForMediaPlayback = false;
        } else {
            config.mediaPlaybackRequiresUserAction = false;
        }
#pragma clang diagnostic pop
```

Or to just use the latest and support iOS 10+?

```
        if ([config respondsToSelector:@selector(mediaTypesRequiringUserActionForPlayback)]) {
            config.mediaTypesRequiringUserActionForPlayback = WKAudiovisualMediaTypeNone;
        }
```

srinathtm-apple: I would say use the latest since ResearchKit 2.0 sets the base SDK to iOS 11.0.

overcyn: Fixed

srinathtm-apple:
Merging into master.

Issue #1142, opened 2018-07-24
ORKHealthClinicalTypeRecorder should not be using NSJSONSerialization


srinathtm-apple:
We are using NSJSONSerialization to deserialize the FHIR resource for each clinical record, create an array, and then use NSJSONSerialization to re-serialize this array. This means we'll suffer from JSON precision issues.

srinathtm-apple:
This has been fixed in #1162

PR #1162, opened 2018-08-20
Fix for issue #1142

srinathtm-apple:


srinathtm-apple:
#1142

srinathtm-apple:
Merging into master.

PR #1169, opened 2018-09-04
Fix using deprecated declaration.

karimhm:
Fix using deprecated declaration.
Remove compiler warnings about the use of deprecated declaration on earlier versions.


srinathtm-apple:
Hey @karimhm , given that ResearchKit 2.0 has a minimum base iOS SDK version as 11.0, do we really need to support UIAccessibilityVoiceOverStatusChanged?

srinathtm-apple:
Closing out this PR since the changes are no longer valid, please file an issue if we should continue to supported the deprecated UIAccessibilityVoiceOverStatusChanged notification.

Issue #657, opened 2016-04-11
optionally hide value in ORKSliderScaleView


dukemike:
There are use cases where displaying the value of a scale may affect the user's response. Could a `valueLabelEnabled` attribute be added?

YuanZhu-apple:
https://github.com/ResearchKit/ResearchKit/pull/446/ has been tried to work on this.
A quick workaround would be create a subclass of `ORKScaleAnswerFormat`/`ORKContinuousScaleAnswerFormat` to override `numberFormatter`, make formatter return an empty string for current value.

bgannin:
I'm currently recreating #446 and updating the changes for completeness before submitting a new PR to address this ticket.

bgannin:
FYI: there's a pull request for this now 🤞

simon-apple:
Thanks for bringing this up! This issue has since been addressed in ResearchKit. We recommend upgrading to the latest release, which includes this fix along with many improvements and new features. If you encounter any issues after upgrading, feel free to open a new issue. Thanks for your support and for contributing to ResearchKit!

PR #446, opened 2015-09-09
Scale hide value label

rugheid:
This pull request adds an optional boolean to all scale answer formats to hide the value label of the scale. The boolean value is not included in the initialisers, but has to be set explicitly if one wants to override the default behaviour. Two steps are added to ORKCatalog to the scale task. The first added step contains a discrete horizontal scale with a hidden value label. The second added step contains a continuous vertical scale with a hidden value label.

PR #1181, opened 2018-09-27
Add option to hide selected value in slider scale (#657)

bgannin:
Fixes issue #657. 

Bring in and update the original changes from PR #446. Utilize in ORKTest - the warmth sliders, both horizontal and vertical, and a copy of the text choice horizontal slider have it enabled - as well as the JSON serialization unit test.

No new warnings or failures introduced. No localization impact. Whitespace verified 👍 

Warmth Vertical
Warmth Horizontal
Feeling Horizontal


ResearchKit/Common/ORKAnswerFormat.h
```diff
@@ -109,12 +111,14 @@ ORK_CLASS_AVAILABLE
                                                                    defaultValue:(double)defaultValue
                                                           maximumFractionDigits:(NSInteger)maximumFractionDigits
                                                                        vertical:(BOOL)vertical
+                                                              hideSelectedValue:(BOOL)shouldHideValue
```
srinathtm-apple: Hey @bgannin , do you think it is necessary to add this parameter to the designated initializer? My concern is that this may not be backwards compatible..

Can't this simply remain as a bool property that defaults to NO? Would love to hear your thoughts about this..

bgannin: Hi @srinathtm-apple, I went back and forth on this one - for completeness of configuration in the default initializer it seems to fit there. That said, a convenience initializer to match the prior signature would ease the transition and ensure backward compatibility. I'll add one 👍

bgannin: To add a bit more…

I agree that these class methods can be reverted to not include the value for compatibility and stick with this change as a property essentially. That said, it makes sense to me to have the designated initializer expand to include with a convenience equivalent to the prior one (in the same way that min/max descriptions are optional but included for completeness of setup.)

srinathtm-apple: @bgannin thanks for the consideration! This might be part of a bigger discussion, but we need to establish some trade-offs between completeness of configuration and overhead of adding convenience initializers.

For this particular use case, I can totally see developers wanting to leverage this in their implementation immediately by adding a new line which sets the property hideSelectedValue to YES. And for folks implementing a scale answer format for the very first time, does adding a fourth initializer option help them or will it just end up cluttering Xcode's auto complete scroll window?

Would appreciate your thoughts/feedback with regards to this.

bgannin: @srinathtm-apple That's a good point and I'll defer to the larger discussion for the framework. Will roll back the initializer changes and do it as an editable property later today.

ResearchKit/Common/ORKAnswerFormat.h
```diff
@@ -329,6 +329,11 @@ ORK_CLASS_AVAILABLE
  */
 @property (readonly, getter=isVertical) BOOL vertical;
 
+/**
```
srinathtm-apple: minor nit - remove (read-only) from comment

ResearchKit/Common/ORKAnswerFormat.h
```diff
@@ -496,6 +501,11 @@ ORK_CLASS_AVAILABLE
  */
 @property (readonly) NSNumberFormatter *numberFormatter;
 
+/**
```
srinathtm-apple: minor nit - remove (read-only) from comment

ResearchKit/Common/ORKAnswerFormat.h
```diff
@@ -609,6 +618,11 @@ ORK_CLASS_AVAILABLE
  */
 @property (readonly, getter=isVertical) BOOL vertical;
 
+/**
```
srinathtm-apple: remove the setter here to remain consistent with the previous two property declarations?

ResearchKit/Common/ORKAnswerFormat.h
```diff
@@ -609,6 +618,11 @@ ORK_CLASS_AVAILABLE
...
+ A Boolean value indicating whether the selected value should be hidden.
+ */
+@property (assign, getter=shouldHideSelectedValueLabel,setter=setHideSelectedValue:) BOOL hideSelectedValue;
+
 /**
  The colors to use when drawing a color gradient above the slider. Colors are drawn such that
```
srinathtm-apple: Also, do you think it is worth explicitly setting _hideSelectedValue to NO as part of the initializers?

ResearchKit/Common/ORKAnswerFormat.h
bgannin: I added the getter with this format here to match the precedent of -isVertical and and -isSecureTextEntry in ORKAnswerFormat.

ResearchKit/Common/ORKAnswerFormat.h
bgannin: Yes, given the architectural differences per platform for BOOL. Will update momentarily.

ResearchKit/Common/ORKAnswerFormat.h
bgannin: This wasn't a part of the PR. AFAICS it's meant to be read-only so that subclasses override it rather than assign to a property managed by ORKAnswerFormat itself.

ResearchKit/Common/ORKAnswerFormat.h
srinathtm-apple: I was referring to the comment that is part of the PR for hideSelectedValue - "A Boolean value indicating whether the selected value should be hidden. (read-only)"

ResearchKit/Common/ORKAnswerFormat.h
srinathtm-apple: I was referring to the comment that is part of the PR for hideSelectedValue - "A Boolean value indicating whether the selected value should be hidden. (read-only)"

ResearchKit/Common/ORKAnswerFormat.h
srinathtm-apple: Here are the two different signatures that is part of this PR -

1. @property (assign, getter=shouldHideSelectedValueLabel) BOOL hideSelectedValue;
2. @property (assign, getter=shouldHideSelectedValueLabel,setter=setHideSelectedValue:) BOOL hideSelectedValue;

Can we just stick with the signature in (1) ?

bgannin:
@srinathtm-apple Thanks for the corrections, it was a bit of cruft from switching from initializer to property. I think we're close now 👍

Testing/ORKTest/ORKTest/Tasks/TaskFactory+QuestionSteps.m
```diff
@@ -677,6 +683,35 @@ @implementation TaskFactory (QuestionSteps)
...
+        ORKTextChoice *textChoice5 = [ORKTextChoice choiceWithText:@"Excellent" value:@(5)];
+        
+        NSArray *textChoices = @[textChoice1, textChoice2, textChoice3, textChoice4, textChoice5];
+        
+        ORKTextScaleAnswerFormat *scaleAnswerFormat = [ORKAnswerFormat textScaleAnswerFormatWithTextChoices:textChoices
+                                                                                               defaultIndex:3
+                                                                                                   vertical:NO];
+        
+        scaleAnswerFormat.hideSelectedValue = YES;
+
```
srinathtm-apple: This is not consistent with the comment at the beginning of the block -         
          /*
          Horizontal text choice scale with visible selected value.
          */

Update the comment?

bgannin: Sorry about that.

srinathtm-apple:
Merging into master. Thanks a bunch for this submission @bgannin !

PR #1182, opened 2018-09-28
Allow task view controller to be canceled directly

MacroYau:
Setting an `ORKTaskViewController` to cancelable omits showing the "End Task" action sheet, and dismisses the view controller directly, which can improve the UX for non-data-collecting tasks that are not necessary to deal with user cancellation.

### DIFF
--- a/ResearchKit/Common/ORKTaskViewController.h
+++ b/ResearchKit/Common/ORKTaskViewController.h
@@ -487,6 +487,13 @@ ORK_CLASS_AVAILABLE
  */
 @property (nonatomic, readonly) UINavigationBar *navigationBar;
 
+/**
+ A Boolean value indicating whether the task can be canceled directly, without prompting the "End Task" action sheet.
+ 
+ The default value of this property is `NO`. To allow the task to be canceled directly, set the value to `YES`.
+ */
+@property (nonatomic, assign) BOOL cancelable;
+
 @end
 
 NS_ASSUME_NONNULL_END
--- a/ResearchKit/Common/ORKTaskViewController.h
+++ b/ResearchKit/Common/ORKTaskViewController.h
@@ -488,11 +488,13 @@ ORK_CLASS_AVAILABLE
 @property (nonatomic, readonly) UINavigationBar *navigationBar;
 
 /**
- A Boolean value indicating whether the task can be canceled directly, without prompting the "End Task" action sheet.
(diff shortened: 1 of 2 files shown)


ResearchKit/Common/ORKTaskViewController.h
```diff
@@ -487,6 +487,13 @@ ORK_CLASS_AVAILABLE
  */
 @property (nonatomic, readonly) UINavigationBar *navigationBar;
 
+/**
+ A Boolean value indicating whether the task can be canceled directly, without prompting the "End Task" action sheet.
+ 
+ The default value of this property is `NO`. To allow the task to be canceled directly, set the value to `YES`.
+ */
+@property (nonatomic, assign) BOOL cancelable;
+
```
srinathtm-apple: worth renaming this to "discardable"?

MacroYau: I am open to either option, but my concern is whether naming it "discardable" would confuse this with "discarding task results"; wherein this property is for dismissing a `ORKTaskViewController`?

Maybe something like `dismissWithoutPrompt`...?

srinathtm-apple: Per your current code path, when the user hits "Cancel" the developer will receive the taskVC: didFinishWithReason: callback with value ORKTaskViewControllerFinishReasonDiscarded, which is the same as "discard results". Which is why I suggested "discardable".

Let me know if that makes sense..

MacroYau: Oh good point! That makes perfect sense to me now.

ResearchKit/Common/ORKTaskViewController.m
```diff
@@ -1513,6 +1514,7 @@ - (void)reviewStepViewController:(ORKReviewStepViewController *)reviewStepViewCo
 
 static NSString *const _ORKTaskRunUUIDRestoreKey = @"taskRunUUID";
 static NSString *const _ORKShowsProgressInNavigationBarRestoreKey = @"showsProgressInNavigationBar";
+static NSString *const _ORKDiscardableTaskResotreKey = @"discardableTask";
```
srinathtm-apple: Typo - ResotreKey -> RestoreKey

srinathtm-apple:
Merging into master.

PR #1185, opened 2018-10-11
ScreenType iPhone Max & XR

tobiasjungnickel:
adds iPhone Max and iPhone XR (same points resolution)

### DIFF
--- a/ResearchKit/Common/ORKSkin.h
+++ b/ResearchKit/Common/ORKSkin.h
@@ -187,7 +187,7 @@ typedef NS_ENUM(NSInteger, ORKScreenMetric) {
 
 typedef NS_ENUM(NSInteger, ORKScreenType) {
     ORKScreenTypeiPhoneX,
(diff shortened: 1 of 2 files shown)


ResearchKit/Common/ORKSkin.h
```diff
@@ -187,6 +187,7 @@ typedef NS_ENUM(NSInteger, ORKScreenMetric) {
 
 typedef NS_ENUM(NSInteger, ORKScreenType) {
     ORKScreenTypeiPhoneX,
+    ORKScreenTypeiPhoneMax,
```
srinathtm-apple:
```suggestion
    ORKScreenTypeiPhoneXSMax,
```

srinathtm-apple: Super-duper minor nit - but can we have the name spelled out a iPhoneXSMax, just to stay consistent with the naming format for all other device screen type names?

srinathtm-apple:
Merging into master.

Issue #1184, opened 2018-10-05
Carthage Install Issue


kcarriedo:
Install failed when running `carthage update` command

Xcode Version: `10.0 (10A255)`

Cartfile: 
`github "ResearchKit/ResearchKit" "stable"`

Issue:
`Checking out ResearchKit at "b46a5f4d8350549918c8aae7d16b67dcb73fbff9"
A shell task (/usr/bin/env git checkout --quiet --force b46a5f4d8350549918c8aae7d16b67dcb73fbff9 (launched in /Users/user_name/Library/Caches/org.carthage.CarthageKit/dependencies/ResearchKit)) failed with exit code 128:
git-lfs smudge -- 'LFS-Files/speechInNoise/Apple/English/V0/Apple-English-V0-Sentence1.wav': git-lfs: command not found
error: external filter git-lfs smudge -- %f failed -1
error: external filter git-lfs smudge -- %f failed
fatal: LFS-Files/speechInNoise/Apple/English/V0/Apple-English-V0-Sentence1.wav: smudge filter lfs failed`

bgannin:
If I'm reading correctly it seems like this is a [blocking issue](https://github.com/Carthage/Carthage/issues/2071) in Carthage that ResearchKit isn't going to be able to work around now that it's configured with Git-LFS.

kcarriedo:
Correct, was able to resolve the issue by installing Git-LFS but you may want to update the docs to make Git-LFS a requirement when install using [dependency manager](https://github.com/ResearchKit/ResearchKit/blob/master/docs-standalone/dependency-management.md).

PR #1192, opened 2018-10-20
Documentation tweaks

bgannin:
Per suggestion in issue #1184 I've updated the dependency management documentation to indicate what tools are necessary. I also added a few badges to the README as found in many GH projects for glanceable information. 

readmeflair


srinathtm-apple:
Merging into master.

Issue #1214, opened 2018-11-24
Project hardening


bgannin:
I've enabled a large number of analyzer and compiler flags to harden the overall project's code including running the analyzer by default.

CLANG_ANALYZER_GCD_PERFORMANCE = YES;
CLANG_ANALYZER_LOCALIZABILITY_EMPTY_CONTEXT = YES;
CLANG_ANALYZER_SECURITY_FLOATLOOPCOUNTER = YES;
CLANG_ANALYZER_SECURITY_INSECUREAPI_RAND = YES;
CLANG_ANALYZER_SECURITY_INSECUREAPI_STRCPY = YES;
CLANG_UNDEFINED_BEHAVIOR_SANITIZER_INTEGER = YES;
CLANG_UNDEFINED_BEHAVIOR_SANITIZER_NULLABILITY = YES;
CLANG_WARN_ASSIGN_ENUM = YES;
CLANG_WARN_OBJC_IMPLICIT_RETAIN_SELF = YES;
CLANG_WARN_SEMICOLON_BEFORE_METHOD_BODY = YES;
GCC_TREAT_IMPLICIT_FUNCTION_DECLARATIONS_AS_ERRORS = YES;
GCC_WARN_ABOUT_MISSING_NEWLINE = YES;
GCC_WARN_ABOUT_MISSING_PROTOTYPES = YES;
GCC_WARN_SHADOW = YES;
GCC_WARN_UNKNOWN_PRAGMAS = YES;
GCC_WARN_UNUSED_LABEL = YES;
RUN_CLANG_STATIC_ANALYZER = YES;

I'm going to address reported issues and submit these all in a PR.

bgannin:
Priority inversion and implicit self retains in blocks were substantial tasks and I've disabled them again for a separate PR.

PR #1216, opened 2018-11-28
Project hardening

bgannin:
Turns on additional compiler warnings and analyzer checks per #1214. Shallow analyzer is enabled by default now.

Addresses all reported issues so the project builds with the same number of deprecation warnings and passes both shallow and deep analysis phases. The specific changes are enumerated in the individual commits.


ResearchKit/Common/ORKWeightPicker.m
```diff
@@ -322,10 +322,11 @@ - (NSArray *)kilogramValues {
     _canonicalMinimumValue = minimumValue;
     _canonicalMaximumValue = maximumValue;
     
-    for (double i = minimumValue; i <= maximumValue; i++) {
+    for (NSUInteger i = minimumValue; i <= maximumValue; i++) {
         [mutableWholeValues addObject:@(i)];
```
srinathtm-apple: since minimumValue is a double, I think it's risky to declare "i" as a NSUInteger...

bgannin: Iteration on float values is flagged by the compiler under security IIRC, it needs to be converted to a safer scalar format.

bgannin: Yep: CLANG_ANALYZER_SECURITY_FLOATLOOPCOUNTER. Solid discussion here: https://wiki.sei.cmu.edu/confluence/display/c/FLP30-C.+Do+not+use+floating-point+variables+as+loop+counters

srinathtm-apple: Ah thanks for sharing that article!

PR #1236, opened 2019-02-04
Fixed email regex to enforce start and end of string.

gustafnilklint:
Previous implementation could accidentally accept white space in email string
te st@mail.ex would be accepted because the substring "st@mail.ex" would be valid.


srinathtm-apple:
Thanks for submitting this fix @gustafnilklint , merging into master.

PR #1271, opened 2019-05-08
Conditionally Hide an ORKFormItem based on an NSPredicate

eschramm:
Implements an enhancement [discussed here recently](https://github.com/ResearchKit/ResearchKit/pull/1264/files#diff-f953f90e65bcefc4e8aefb2a8a587851) where an `ORKFormItem` can be hidden via an `NSPredicate`. This can be useful for a showing/hiding later `ORKFormItem`s based on answers in the same `ORKFormStep`. This is demonstrated below. It can also be used to simply provide conditional display of various items in a lengthy `ORKFormStep` as well based on previous responses.

In this implementation, a `ORKFormItem` where `isOptional` is `false` will NOT be required to move to the next step. This prevents an unsatisfiable condition in validation where a required step is not visible to the user.

Furthermore, results for hidden form items at the time the step is completed are removed, similarly to skipped steps (via rules) in a task.

showHide2

### DIFF
--- a/ResearchKit/Common/ORKFormStep.h
+++ b/ResearchKit/Common/ORKFormStep.h
@@ -203,7 +203,7 @@ ORK_CLASS_AVAILABLE
 /**
  A predicate that when true, hides the item from display.
  */
-@property (nonatomic, nullable) NSPredicate *hideItemPredicate;
+@property (nonatomic, nullable) NSPredicate *hidePredicate;
 
 /**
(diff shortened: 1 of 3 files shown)


srinathtm-apple:
@eschramm - this is a great addition to the framework! Merging into master, thank you!

eschramm:
@srinathtm-apple I noticed that this functionality was reverted with the ResearchKit 2.1 updates (#1321 ). Was this intentional?

PR #1280, opened 2019-06-03
Missing embedded summary in ORKReviewStep

gabriel-blanco-apple:
Bug introduced here:

commit 9ef1876c45f874ba04d67dd9c5ba7ae5d88a52c3
Author: Brian Ganninger <coder@infinitenexus.com>
Date:   Sun Nov 25 20:44:22 2018 -0800

    [code cleanup] Review mismatched nil returns (issue #1214)

——

Impact:
Causes an embedded ORKReviewStep to return no steps

### DIFF
--- a/ResearchKit/Common/ORKReviewStep.m
+++ b/ResearchKit/Common/ORKReviewStep.m
@@ -116,7 +116,7 @@
 }
 
 - (BOOL)isStandalone {
-    return _steps != nil;
+    return _steps.count > 0;
 }
 
 @end


gabriel-blanco-apple:
Previously, initializing an embedded ORKReviewStep

```objc
+ (instancetype)embeddedReviewStepWithIdentifier:(NSString *)identifier {
```

Would leave the steps nullable. After the above change, we now initialize steps as an empty array, which is resulting in the wrong value being returned for the `isStandalone` method (steps is not nil, but is an empty array). 

Sugested fix is to return `steps.count > 0`

PR #1293, opened 2019-07-23
Fix the ORKValuePickerAnswerFormat can not popup the picker controller in the Form

fansaien:
The ORKValuePickerAnswerFormat in the form should be in an individual section. 
Because it can not popup the picker controller when it is in the same section with others.


fansaien:
The latest version solved this issue

PR #1329, opened 2019-12-11
remove tests from archive

xinsight:
This is preventing carthage from building ResearchKit with:

```
carthage update ResearchKit --platform ios
```

`Cartfile` contains:
```
github "ResearchKit/ResearchKit" "master"
```


srinathtm-apple:
thanks for the fix! @xinsight

Issue #1336, opened 2020-02-10
dB HL Tone Audiometry - 'NSDecimalNumber overflow exception'


rogelj:
I am trying to use the `dBHLToneAudiometryTask` but when I try to run the application I get the following error:

`*** Terminating app due to uncaught exception 'NSDecimalNumberOverflowException', reason: 'NSDecimalNumber overflow exception'`

![Screenshot 2020-02-10 at 17 58 19](https://user-images.githubusercontent.com/5548677/74176352-6df26f00-4c2f-11ea-91d3-fcb1ec41352c.jpg)

I have also tried running the ORKCatalog example in the repo and I get the same error. Any hints would be most welcome.

rogelj:
I have narrowed it down to this line (345) in `ORKdBHLToneAudiometryAudioGenerator.m` 

```swift
NSDecimalNumber *updated_dBSPLForVolumeCurve = [dBSPL decimalNumberByAdding:offsetDueToVolume];
```

Will continue investigating, but any hints are more than welcome!

rogelj:
OK, it seems that the `_sensitivityPerFrequency` and `_retspl` are not being set up as the `headphones` variable is not set. Appropriate code has been written for the `_volumeCurve`. There appropriate `.plists` for the two Dictionaries may need to be revised. Will try to open a pull request to sort this issue.

PR #1337, opened 2020-02-11
dbHL Tone Audiometry requires valid values for "Sensitivity per frequency" and `_retspl` to work

rogelj:
The generation of tones to be used in the dbHL Tone Audiometry Active Task requires valid dictionaries for `_sensitivityPerFrequency` and `_retspl` to work. 
Currently the value of the `headphones` is checked to set these values. 
Appropriate values for "WIRED" headphones may be required, and at least the changes proposed in this pull request make the task run again.


rogelj:
This pull request addresses the Issue #1336

PR #1366, opened 2020-04-02
Handle new UITouchPhase enumerations.

gfiumara:
Three new `UITouchPhase`s were added in iOS 13.4.


srinathtm-apple:
@gfiumara - thanks for submitting this!

PR #1383, opened 2020-06-24
Fix "Create Survey" document formatting

vldmrkl:
I noticed, that there were a few problems with titles on a "Create Survey" document markdown, so here is a small fix 🙌


srinathtm-apple:
thanks @vldmrkl !

PR #1384, opened 2020-06-24
RK 2.1 updates

srinathtm-apple:


ResearchKit/Common/ORKIconButton.m
```diff
@@ -0,0 +1,138 @@
...
+    _imageView = [UIImageView new];
+    _imageView.image = _buttonIcon;
+    _imageView.translatesAutoresizingMaskIntoConstraints = NO;
+    
+    [_customView addSubview:_imageView];
+}
+
+- (void)setupConstraints {
+    [[_textLabel.centerYAnchor constraintEqualToAnchor:_customView.centerYAnchor] setActive:YES];
+    [[_textLabel.leadingAnchor constraintEqualToAnchor:_customView.leadingAnchor constant:15.0] setActive:YES];
```
pmckinney8, line 111: @srinathtm-apple forgot to add constants for the values here. Is that worth changing/adding right now?

Pariecemckinney-apple, line 111: @srinathtm-apple forgot to add constants for the values here. Is that worth changing/adding right now?

Issue #1406, opened 2020-08-25
ORKTextChoiceOther.choice Seemingly not working with iOS 14


scdi:
It seems that the ORKTextChoiceOther.choice is working in iOS 13 but NOT with iOS 14. 
A click in the TextBox that opened when the other choice is selected closes the TextBox rather than allowing the user to enter a text.
![Screen Recording 2020-08-25 at 10 08 43 AM](https://user-images.githubusercontent.com/13545391/91185394-c49b3400-e6bb-11ea-8d66-c4e73f85968a.gif)

zinn-claus:
I can confirm this bug for iOS 14 as well. It also happens when textViewStartsHidden is set to false. This is very annoying. Since the issue was raised a couple of months ago, what are the chances that it will be fixed soon?

smclelland:
Yes, can confirm that I'm seeing the same thing. Had to disable for now.

marius-serban:
There's a fix on a branch:  #1429. If you want, you can merge that into your fork.

PR #1429, opened 2020-11-23
Fix ORKTextChoiceOther.choice not working on iOS 14

marius-serban:
Fixes issue https://github.com/ResearchKit/ResearchKit/issues/1406 by setting up the subviews of a `ORKChoiceViewCell` as children of `contentView` rather than the cell directly.

Tested on both iOS 14 and iOS 13.

### DIFF
--- a/ResearchKit/Common/ORKChoiceViewCell.m
+++ b/ResearchKit/Common/ORKChoiceViewCell.m
@@ -179,7 +179,7 @@ static const CGFloat LabelCheckViewPadding = 10.0;
     if (!_containerView) {
         _containerView = [UIView new];
     }
-    [self addSubview:_containerView];
+    [self.contentView addSubview:_containerView];
 }
 
 - (void)addContainerViewToSelfConstraints {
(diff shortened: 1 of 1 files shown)


scdi:
Thank you for the fix. I think that it is a partial fix. It works for single choice item but not for multiple choice item.
When used with multiple choice items, the other selected choices get deselected.
Screen Recording 2020-12-14 at 3 05 30 PM

OmarAlhammadi1996 (approved):

jscubida (approved):

scdi:
@jscubida Do you have a demonstration where the fix works for multipleChoice formStep. It seems that "other choice" deselect all other choices.

marius-serban:
@scdi I think it'd be better to raise a separate issue for the multiple choice problem. This PR focuses on only fixing #1406 as it was described.
> It seems that the ORKTextChoiceOther.choice is working in iOS 13 but NOT with iOS 14.
> A click in the TextBox that opened when the other choice is selected closes the TextBox rather than allowing the user to enter a text.

I think that raising it as a separate issue will increase its chances of getting fixed. I don't have the capacity to look into it, but others might.

scdi:
Thank you for your work on this. I am sorry that the problem submitted cannot be fixed at this time. The example shown above, submitted in December, last year, is about the ORKChoice not working for multipleChoice formStep.

jscubida:
@scdi I think the behavior you illustrated is due to the value for `exclusive:`. If you have the code that was run to produce the GIF above, I can take a look. Regardless, I have a sample project [here](https://github.com/TympaHealth/ResearchKit/pull/12) that allows you to view the difference in behavior when `exclusive:` is true or false. Would this suffice or were you looking for something else?

scdi:
@jscubida, You are a life saver. I have needed for so long to use this functionality and could not. You are right, the exclusive:. should be false for the Other to work when .multipleChoice is being used. Thank you so much.

PR #1461, opened 2021-06-21
Update deprecated method call

erik-apple:
This PR updates a call to a deprecated method that generated warnings in projects that import ResearchKit


joeylabarck-apple (approved):
Thanks, Eric!

wbj:
Thank you everybody!

PR #1486, opened 2022-02-11
IOS 15 Fixes

Pariecemckinney-apple:


ResearchKit/Common/ORKSkin.m
```diff
@@ -236,7 +236,13 @@ static ORKScreenType ORKGetHorizontalScreenTypeForBounds(CGRect bounds) {
         // Use this method instead of UIApplication's keyWindow or UIApplication's delegate's window
         // because we may need the window before the keyWindow is set (e.g., if a view controller
         // loads programmatically on the app delegate to be assigned as the root view controller)
-        window = [UIApplication sharedApplication].windows.firstObject;
+        
+        for (UIScene *scene in UIApplication.sharedApplication.connectedScenes) {
+            if ([scene.delegate conformsToProtocol:@protocol(UIWindowSceneDelegate)]) {
+                window = [(id<UIWindowSceneDelegate>)scene.delegate window];
```
ronzilla-apple, line 242: Do you want to add a break  if you get into this condition?
Rather than relying on there only being one connected scene.

ResearchKit/Common/ORKPasscodeStepViewController.m
```diff
@@ -499,7 +499,7 @@ + (void)savePasscode:(NSString *)passcode withTouchIdEnabled:(BOOL)touchIdEnable
 
 - (void)removePasscodeFromKeychain {
     NSError *error;
-    [ORKKeychainWrapper objectForKey:PasscodeKey error:&error];
+    [ORKKeychainWrapper objectOfClass:NSDictionary.self forKey:PasscodeKey error:&error];
```
ronzilla-apple, line 502: This is technically unsafe. The contract for this kind of API is that if the result is nil, you can check whether there's more info in the error, but if the result is non-nil, there's no guarantee about what's in error. 
```suggestion
    id anyValue = [ORKKeychainWrapper objectOfClass:NSDictionary.self forKey:PasscodeKey error:&error];
    if (anyValue != nil) {
```

ResearchKit/Consent/ORKEAGLMoviePlayerView.m
```diff
@@ -123,12 +123,15 @@ - (BOOL)validateProgram:(GLuint)prog;
...
 
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-implementations"
 @implementation ORKEAGLMoviePlayerView
 
 const GLfloat DefaultPreferredRotation = 0;
 
 + (Class)layerClass {
     return [CAEAGLLayer class];
+//    return [CALayer class];
```
ronzilla-apple, line 134:
```suggestion
```

ResearchKit/Consent/ORKEAGLMoviePlayerView.m
```diff
@@ -490,10 +493,10 @@ Vertex data formed using (-1,-1) and (1,1) as the bottom left and top right coor
     [self saveGLContext];
     
     glBindVertexArrayOES(_vertexArrayHandle);
-
```
ronzilla-apple, line 493: We could make this PR smaller by getting rid of these whitespace changes

PR #1490, opened 2022-02-22
Add spacing below ORKTableContainerView header

jackcook:
The new iOS 15 changes that were pushed a few days ago caused the spacing below form headers to be removed. See screenshots for before (current main branch), and after (my PR). I tested before and after on a couple different devices and the change appears to hold up well. The example shown here is the "Form Survey Example" in the ORKCatalog app.

| Current Main Branch | My PR |
| --------------------- | ------ |
| Simulator Screen Shot - iPhone 13 Pro Max - 2022-02-22 at 17 56 40 | Simulator Screen Shot - iPhone 13 Pro Max - 2022-02-22 at 17 57 08 |


jackcook:
Just added a check to make sure this code only compiles on iOS 15+, as I just noticed the repository's target version has been updated since I made this PR.

jackcook:
It looks like this has been addressed in the recent 2.1.1 update, thank you.

PR #1491, opened 2022-02-22
Fix ORKTableContainerView footer in fullScreen modals

jackcook:
My second PR for the day! I figured it'd be better to split them up into two separate requests. I've noticed a bug where the continue and skip buttons are not displayed correctly when an `ORKTaskViewController` is presented with `modalPresentationStyle = .fullscreen`. Namely, the following code produces the visual issue seen below:

```
let taskViewController = ORKTaskViewController(task: task, taskRun: nil)
taskViewController.modalPresentationStyle = .fullScreen
present(taskViewController, animated: true, completion: nil)
```

The form shown in the screenshots below is the "Form Survey Example" in the ORKCatalog app. You can test this issue by adding `taskViewController.modalPresentationStyle = .fullScreen` below line 117 in `ORKCatalog/Presenting Tasks/TaskListViewController.swift`, right after the `taskViewController` is instantiated.

| Current Main Branch | My PR |
| ---- | --- |
| Simulator Screen Shot - iPhone 13 - 2022-02-22 at 18 16 17 | Simulator Screen Shot - iPhone 13 - 2022-02-22 at 18 16 37 |
| Simulator Screen Shot - iPad Pro (9 7-inch) - 2022-02-22 at 18 04 04 | Simulator Screen Shot - iPad Pro (9 7-inch) - 2022-02-22 at 18 04 22 |

When the devices are rotated, this change does not prevent the buttons from adjusting their size, and it does not appear to cause issues for how the footer view was displayed with other modal presentation styles. However, if there is a better solution for this issue, please let me know. Thank you!

### DIFF
--- a/ResearchKit/Common/ORKTableContainerView.m
+++ b/ResearchKit/Common/ORKTableContainerView.m
@@ -222,7 +222,7 @@ static const CGFloat FooterViewHeightOffset = 20.0;
         CGFloat newHeight = tableViewHeight - self.tableView.contentSize.height + FooterViewHeightOffset;
         CGRect footerBounds = newHeight < minHeight ? CGRectMake(0.0, 0.0, _tableView.bounds.size.width, minHeight) : CGRectMake(0.0, 0.0, _tableView.bounds.size.width, newHeight);
 
-        [_footerView setBounds:footerBounds];
+        [_footerView setFrame:footerBounds];
         _tableView.tableFooterView = _footerView;
     }
 }


louie-apple:
Hi Jack
Your PR looks good, but we believe this issue is just fixed in the main branch. Please pull main and let us know if it fixes your issue.
Thank you!

jackcook:
Yes, it looks like this has been fixed with the recent update. Thank you!

PR #1505, opened 2022-06-09
Prevent isPasscodeStoredInKeychain crash, correct warnings

### LINKED ISSUE #1504: ORKPasscodeViewController's isPasscodeStoredInKeychain causes an exit
Appears to have been introduced in #1487.

If `ORKPasscodeViewController.isPasscodeStoredInKeychain()` is called when the dictionary does not exist, e.g. at first launch of the application, the application will terminate with uncaught exception of type `NSException`: `"There is no passcode stored in the keychain."`.

I am unsure as to why this was changed - I believe the main reason for this function is to return `YES` if a passcode is stored in the keychain, else return `NO`. I think, therefore, that this function should not throw an exception in this case, but instead return `NO`.

SwiftArtery:
This PR updates `isPasscodeStoredInKeychain` in `ResearchKit/Common/ORKPasscodeViewController.m` to remove the `@throw NSException` called when the dictionary is `nil`, and instead just return `NO` here.

This prevents a crash when `isPasscodeStoredInKeychain` is called in certain circumstances, such as when called at the very first launch of the application.

---

It also updates the invocation of
```
(id<NSSecureCoding>)objectOfClass:(Class)objectClass forKey:(NSString *)key error:(NSError **)errorOut
```
to be
```
(id<NSSecureCoding>)objectsOfClasses:objectClasses forKey:(NSString *)key error:(NSError **)errorOut
```
\- thus taking an `NSSet<Class>` input and ensuring that `NSString` and `NSNumber` are included in the allowed classes set. This prevents warnings in the console, such as
```
[general] *** -[NSKeyedUnarchiver validateAllowedClass:forKey:] allowed unarchiving safe plist type ''NSString' (0x1f697e4f0) [/System/Library/Frameworks/Foundation.framework]' for key 'NS.objects', even though it was not explicitly included in the client allowed classes set: '{(
    "'NSDictionary' (0x1f6973850) [/System/Library/Frameworks/CoreFoundation.framework]"
)}'. This will be disallowed in the future.
```
and
```
[general] *** -[NSKeyedUnarchiver validateAllowedClass:forKey:] allowed unarchiving safe plist type ''NSNumber' (0x1f697ed38) [/System/Library/Frameworks/Foundation.framework]' for key 'NS.objects', even though it was not explicitly included in the client allowed classes set: '{(
    "'NSDictionary' (0x1f6973850) [/System/Library/Frameworks/CoreFoundation.framework]"
)}'. This will be disallowed in the future.
```

---

All unit tests still pass as expected.

---

Closes #1504

### DIFF
--- a/ResearchKit/Common/ORKKeychainWrapper.h
+++ b/ResearchKit/Common/ORKKeychainWrapper.h
@@ -66,7 +66,7 @@ ORK_CLASS_AVAILABLE
  
  @return An object or `nil` if key is not valid.
  */
-+ (nullable id<NSSecureCoding>)objectOfClass:(Class)objectClass forKey:(NSString *)key error:(NSError * __autoreleasing _Nullable *)errorOut;
++ (nullable id<NSSecureCoding>)objectsOfClasses:objectClass forKey:(NSString *)key error:(NSError * __autoreleasing _Nullable *)errorOut;
 
 /**
  Removes the object in the keychain for the provided key.
--- a/ResearchKit/Common/ORKKeychainWrapper.m
+++ b/ResearchKit/Common/ORKKeychainWrapper.m
@@ -60,14 +60,14 @@ static NSString *ORKKeychainWrapperDefaultService() {
                    error:errorOut];
 }
 
-+ (id<NSSecureCoding>)objectOfClass:(Class)objectClass forKey:(NSString *)key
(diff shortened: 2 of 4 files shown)


Pariecemckinney-apple (approved):
Thanks @SwiftArtery! This looks great. Approved.

Issue #1503, opened 2022-06-03
ORKAnswerFormat.textAnswerFormat UI is treated in the same question and multiple choice


tyne-admin:
I currently have a ORKFormStep and append my questions as different ORKFormItems depending on whether they are a multiple choice question or a text-based question. In the 2.0.0 cocoa pod version, it used to work fine (The questions were separated). When I use the most-recent (non-stable) version, it seems like the text-based question isn't given its own question box. 

here is a snippet of my code:

```
        let formStep = ORKFormStep.init(identifier: "form", title: "Questions", text: "Complete the survey with the best answer for each question.")
        var formItems = [ORKFormItem]()
        
            for (index, question) in questionList.questionList.enumerated() {
                if question.questionType == "yesNo" {
                    let item = ORKFormItem(identifier: "q\(index)", text: question.question, answerFormat: ORKAnswerFormat.booleanAnswerFormat(), optional: false)
                    formItems.append(item)
                } else if question.questionType == "textBased" {
                
                    let item = ORKFormItem(identifier: "q\(index)", text: question.question, answerFormat:   ORKAnswerFormat.textAnswerFormat(withMaximumLength: 100), optional: false)
                    formItems.append(item)
          }
```

And here is an example image:
![Screen Shot 2022-06-02 at 5 24 57 PM](https://user-images.githubusercontent.com/92927021/171761809-8f705b9e-8ee1-48bf-a3d0-21cf4671a409.png)

The  "If yes..." question isn't being treated as question 12, but rather, embedded inside question 11

Pariecemckinney-apple:
Hello @tyne-admin, thanks for posting this issue. We’ve tracked the issue internally and will update you as soon a fix is identified!

kwanhlee:
Hi @Pariecemckinney-apple, this is my personal GitHub account for the Original Poster of this issue.

I have fixed the issue and made a pull request. Please take a look. It seems like there have been a bug in the code where a boolean value was set to NO instead of YES (which is specified in the header file)

kwanhlee:
@Pariecemckinney-apple,
Sure! But I think it may be a good idea to update the header file to mention that the default value is a NO. 
I won't make a PR for that since it's so trivial, but it would be nice to be included in the next version for people reading the documentation.

PR #1508, opened 2022-07-05
Default value for multipleLines in ORKAnswerFormat.m fixed to YES from NO

kwanhlee:
Fix for Issue #1503 

In ORKAnswerFormat.h, the default value for _multipleLines is stated as YES. However, in ORKAnswerFormat.m, this default value is set as NO. This causes the text answer format to not have its own section and be merged into previous sections. 

The fix includes changing the default value from NO -> YES

Multiple_lines_default_val


kwanhlee:
Closed, because changing default value may break base case for users using this default value. 
#1503

Issue #1509, opened 2022-08-18
iOS 16 Compatibility


Brett-Best:
Is it expected that any updates will be made to support iOS 16?
It seems this project is a bit abandoned?

jstemmle-apple:
Hi.  Thanks for reaching out.

ResearchKit is still an active project. We are committed to ensuring it works with  iOS 16.  If you see any problems or issues, please do report them here by filing issues on GitHub.

Brett-Best:
https://github.com/ResearchKit/ResearchKit/pull/1491

I’m seeing an issue with this although not in fullscreen mode, the fix in that PR works

louie-apple:
@Brett-Best  Not sure what issue you had, but did the PR we merged (https://github.com/ResearchKit/ResearchKit/pull/1516) (which also addressed issue 1491) fix your issue too?

Brett-Best:
I can’t seem to get it to compile now using cocoapods.

Brett-Best:
<img width="517" alt="Screen Shot 2022-11-10 at 5 39 28 am" src="https://user-images.githubusercontent.com/6104381/200913768-78bf9588-f381-4fb2-b974-aaede72f14fa.png">
@louie-apple

Brett-Best:
CC: @Pariecemckinney-apple

louie-apple:
> <img alt="Screen Shot 2022-11-10 at 5 39 28 am" width="517" src="https://user-images.githubusercontent.com/6104381/200913768-78bf9588-f381-4fb2-b974-aaede72f14fa.png">
> 
> @louie-apple

Hi @Brett-Best, 

Here are our latest CocoaPods Installation guidelines:

 ------------
 For latest stable release

 ```
 pod 'ResearchKit'
 ```
 The latest early development release based off of main is
 ```
pod 'ResearchKit',  :git => 'https://github.com/ResearchKit/ResearchKit.git', :tag => '2.1.1-dev'

 ```
 or  for the latest early development release based off of main

 ```
   pod 'ResearchKit', :git => 'https://github.com/ResearchKit/ResearchKit.git', :branch => 'main'
 ```

Brett-Best:
Thanks the last few commits made a couple of days ago fixed this for me.
I’m still getting the warnings about the asset catalog.

louie-apple:
No problem! Are these warnings around `The Image set name "" is used by multiple image sets` ? 
[...]

PR #1511, opened 2022-09-10
Fixed compilation when OpenGLES is unavailable

Brett-Best:


louie-apple:
Thank you for submitting your PR. We are curious in what cases do you see that `OpenGLES is unavailable` ?

Brett-Best:
When compiling for macOS (Catalyst)
> OpenGLES is deprecated and is not available when building for Mac Catalyst.

louie-apple:
Thanks for the clarification.  Since this PR is only partially addressing catalyst support, we can't merge at this time.
Have you done the work to fully support macOS (Catalyst)?

Brett-Best:
I guess in principle this PR doesn’t claim to add Catalyst support, e.g. if/when iOS removes OpenGLES then ResearchKit would still compile. I might be wrong, but it seemed like the usage of OpenGLES was for a “deprecated” component of ResearchKit that was replaced with a newer component?

Nonetheless, the portion of RK that I tested out so far in Catalyst was the surveys which seemed to work well after applying the fix linked to in this comment https://github.com/ResearchKit/ResearchKit/issues/1509#issuecomment-1242568509

louie-apple:
Thank you for sharing. You're correct: The OpengGLES components in ResearchKit have been deprecated which means they'll be removed in a future release.

louie-apple:
Glad to hear that you are able to compile for Mac Catalyst now. You are correct that `ORKEAGLMoviePlayerView.h/m`  was still left in the codebase even though the class is deprecated. Those files will be removed in a future release.

PR #1564, opened 2024-03-01
fix: Building ResearchKit for iPad

cbaker6:
- [x] Add back building ResearchKit for the iPad which I believe was accidentally removed by a previous commit


Pariecemckinney-apple (approved):
LGTM! Thanks @cbaker6!

PR #1566, opened 2024-03-01
fix: Deprecate ORKTaskViewControllerFinishReason instead of renaming it

cbaker6:
`ORKTaskViewControllerFinishReason` was renamed to `ORKTaskFinishReason` instead of being deprecated which will break projects dependent on `ORKTaskViewControllerFinishReason`. This PR adds a simple typedef and deprecation warning.

### DIFF
--- a/ResearchKit/Common/ORKTask.h
+++ b/ResearchKit/Common/ORKTask.h
@@ -64,7 +64,7 @@ typedef NS_ENUM(NSInteger, ORKTaskFinishReason) {
     ORKTaskFinishReasonEarlyTermination
 };
 
-ORK_TO_BE_DEPRECATED("Use the ORKTaskFinishReason instead")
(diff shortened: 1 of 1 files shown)


Pariecemckinney-apple:
Hello @cbaker6,

Thank you for submitting this PR! The reason we decided to avoid creating a “typedef” was due to other breaking changes that can’t be avoided when adopting the 3.0 update (e.g updating imports to include ResearchKitUI/ResearchKitActiveTasks). So instead of adding the typedef to prolong the use of ORKTaskViewControllerFinishReason, we’re allowing the build error to show so that all updates will be supplied at the same time.

Generally, this symbol would only appear in an app a few times at most, so it shouldn’t be a large lift for developers.

cbaker6:
Makes sense, I'll close this

PR #1569, opened 2024-03-29
ci: Add code integration testing to repo

### LINKED ISSUE #1470: Some unit tests do not pass when trying to build
I was able to successfully build with many warnings. I tried running the unit tests during build and not all tests passed,
I am on an intel processor
Xcode 13
macOS Big Sur

Screen Shot 2021-10-10 at 6 20 12 PM
Screen Shot 2021-10-10 at 6 20 24 PM

cbaker6:
- [x] Adds CI using GitHub actions to repo. Close #1470 

I believe you have to approve the PR for it to run the CI for the first time.

You can see the CI results on my fork: https://github.com/cbaker6/ResearchKit/pull/3


.github/workflows/ci.yml
```diff
@@ -0,0 +1,27 @@
+name: CI
```
Pariecemckinney-apple, line 1: It would probably be helpful to make this name a bit more descriptive if possible.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,27 @@
...
+
+on:
+  push:
+    branches: [ 'main' ]
```
Pariecemckinney-apple, line 5: Could we include stable as well for push and pull requests?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,27 @@
...
+  pull_request:
+    branches: [ 'main' ]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:        
+  test:
+    runs-on: macos-13
```
Pariecemckinney-apple, line 15: Could we update this to macos-latest?

Pariecemckinney-apple (commented):
Thank you for submitting this PR @cbaker6! 
This could definitely be helpful moving forward. Just a few comments.

.github/workflows/ci.yml
cbaker6: Do you have something in mind? I think the default action name GitHub provides for this is `swift` and it usually is the same name of the file, `swift.yml`.

If there was a special `build.yml` and `release.yml` then we could name them `build` and `release`. Do you think `build` works better here?

cbaker6: Because of the current name, this will show as `ci / test (platform=iOS\ Simulator,OS=17.2,name=iPhone\ 15\ Pro\ Max, ResearchKit) (pull_request)` in the settings which is more specific

.github/workflows/ci.yml
cbaker6: Github action still defaults `macos-latest` to version `12.x.x` which is missing the latest Xcode https://github.com/actions/runner-images/blob/macOS-12/20240329.1/images/macos/macos-12-Readme.md#xcode

I've switched this to `macos-14` instead. Let me know if this works

.github/workflows/ci.yml
cbaker6: Done!

.github/workflows/ci.yml
```diff
@@ -0,0 +1,25 @@
...
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:        
+  test:
+    runs-on: macos-14
+    strategy:
+      matrix:
+        destination: ['platform=iOS\ Simulator,OS=17.4,name=iPhone\ 15\ Pro']
+        scheme: ['ResearchKit', 'ORKCatalog']
```
Pariecemckinney-apple, line 19: Lets keep this specific for testing ResearchKit only for now.

.github/workflows/ci.yml
Pariecemckinney-apple: I think having it say something along the lines of **Build / RK Unit Tests** or something close would be fine. 

If we could avoid displaying the destination information here (platform, simulator, etc) that would also be great.

Pariecemckinney-apple (changes requested):
Remove catalog changes.

.github/workflows/ci.yml
Pariecemckinney-apple: It's saying it defaults to 14 here. But specifying macOS-14 is fine! Thanks for updating.

https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories

cbaker6: I think they haven't updated the `latest` images to point to macOS 14, when I tried latest, it should it's macOS 12 https://github.com/cbaker6/ResearchKit/actions/runs/8620787466/job/23628313753#step:1:8

.github/workflows/ci.yml
cbaker6: Do you want me to move all of the `ORKCatalog` test updates out of this PR? I've tested the changes locally and they all pass now

Pariecemckinney-apple: Yes, keeping the catalog app as is for now would be best.

.github/workflows/ci.yml
cbaker6: These now show up as `build / ResearchKit Unit Tests`:

image

.github/workflows/ci.yml
cbaker6: Reverted

Pariecemckinney-apple (approved):

PR #1589, opened 2024-11-16
fix: Remove iOS support for ResearchKit to fix watchOS build

cbaker6:
The ResearchKitSwiftUI is looking great so far!


simon-apple:
Hi Corey and Luis - thank you for trying out `ResearchKitSwiftUI`! We look forward to any feedback you both may have.

The addition of watchOS to the ResearchKit target as a supported destination is an intentional decision that is not yet reflected in the main branch. Would you or @lmillan1 be able to upload a test project that includes the bare minimum to reproduce the issue? Thanks!

PR #1590, opened 2024-12-06
fix: Use accent color instead of hardcoded blue

cbaker6:
The buttons are hardcoded to blue, preventing the ability to change the color via a modifier. In addition, navigation buttons at the top (e.g. "Cancel") are not hard coded leading to views that don't match when accentColor is passed.

## Before the fix
image

## After the fix
image


simon-apple (approved):
Thank you for the change!

PR #1596, opened 2025-04-13
feat: Make types easier to use by conforming to Codable, Hashable

cbaker6:


jessi-apple (approved):
Thank you for this change @cbaker6!

PR #1599, opened 2025-07-08
fix crash when answering form question

lukaskollmer:
fixes a crash in the `ORKFormStepViewController` when answering the last question in a form, but there's subsequent informational (non-question) sections

the issue here is caused by the fact that the informational sections in the UITableView consist of only a header (which displays the informational content), but no actual cells.
meaning that when you attempt to scroll to the section, it'll try to scroll to a non-existent index path, which'll cause the table view to crash.

we work around this by explicitly handling this situation and adjusting the destination index path to scroll to the section as a whole instead.

### DIFF
--- a/ResearchKitUI/Common/Step/Form Step/ORKFormStepViewController.m
+++ b/ResearchKitUI/Common/Step/Form Step/ORKFormStepViewController.m
@@ -1387,7 +1387,7 @@ NSString * const ORKFormStepViewAccessibilityIdentifier = @"ORKFormStepView";
 
 /// Determines if we should auto-scroll to the next section
 /// @returns The destination index path that should be used when scrolling to the next question, or `nil` if no auto-scroll should occur.
-- (NSIndexPath *_Nullable)shouldAutoScrollToNextSection:(NSIndexPath *)indexPath {
+- (NSIndexPath *_Nullable)indexPathForAutoScrollingToNextSectionAfter:(NSIndexPath *)indexPath {
     if (![self _isAutoScrollEnabled] || _autoScrollCancelled) {
         return nil;
     }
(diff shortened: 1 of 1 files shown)


lukaskollmer:
@simon-apple this is all it takes to reproduce this:
```swift
let formStep = ORKFormStep(identifier: "formStep")
formStep.formItems = [
    ORKFormItem(
        identifier: "question",
        text: "Question1 text",
        answerFormat: .some(.booleanAnswerFormat())
    ),
    ORKFormItem(sectionTitle: "Section Title")
]

let task = ORKNavigableOrderedTask(identifier: "task", steps: [formStep])
let VC = ORKTaskViewController(task: task, taskRun: UUID())
// present the VC, and select an option in the question
```

lukaskollmer:
@simon-apple i incorporated your feedback, and also added an `@available` check since i only just now noticed that ResearchKit deploys all the way back to iOS 13. (section identifiers in the diffable data source are only available starting with iOS 15...)

i reworked the `NSIndexPath`-returning `shouldAutoScroll` method into
- `- (NSIndexPath *_Nullable)indexPathForAutoScrollingToNextSectionAfter:(NSIndexPath *)indexPath`, and
- `- (BOOL)shouldAutoScrollToNextSectionAfter:(NSIndexPath *)indexPath`,

and also renamed the autoScroll method itself from `-autoScrollToNextSection:` to `-autoScrollToNextSectionAfter:`, to bring it in line with your suggested naming changes

simon-apple:
Awesome, thanks for the updates. We'll be qualifying these changes internally alongside other changes. Once everything checks out, we can go ahead and merge this PR. I'll keep you posted.

simon-apple:
@lukaskollmer 

We have qualified these changes and will go ahead and merge this PR. Thanks again for the changes!

If you're interested in chatting with our team, we'd be happy to hear about the projects you've been using the ResearchKit framework for. Feel free to email researchkit@apple.com if so and we can connect.

Issue #1601, opened 2025-08-28
Using localized keys in that are not already translated in Researchkit's localized files in other projects


adonawa:
I am localizing an app that uses the ResearchKit framework to Spanish and Simplified Chinese, however I am having issues with some of the strings that are originally called in the Researchkit framework. 

For example, during the onboarding process, I am calling different `ORKPermissionType`s to display on an `ORKRequestPermissionStep`. I am able to change the title and text property on an `ORKRequestPermissionStep`; however, for the` ORKPermissionType`s that I display on that step, I do not have direct access to change the `localizedTitle` or `localizedDetailText`, as it is a get-only property. 

When I further go into an `ORKPermissionType`, such as `ORKHealthKitPermissionType`, I see that the `localizedTitle` and `localizedDetailText` are created using `ORKLocalizedString(@"REQUEST_HEALTH_DATA_STEP_VIEW_TITLE", nil)`, which does the following:

```
  NSString *ORKLocalizedHiddenString(NSString *key) {
    NSString *value = [ORKBundle() localizedStringForKey:key value:key table:@"ResearchKit"];
    if ([value isEqualToString:key]) {
        // If it fails try to find on default table
        value = [ORKDefaultLocaleBundle() localizedStringForKey:key value:key table:@"ResearchKit"];
    }
    return value;
}
```

So my question is, how would I be able to get this string to be localized in my project? I have added the key in my xcstrings file, but it doesn’t get picked up since this part of ResearchKit is using a custom function for those keys. Is there a way I can override this function or make it call my localized strings? I have added the exact same keys that ResearchKit uses to my project. I am also using the new strings catalog to localize, versus the old strings dictionary, if that makes a difference.

Another note, coincidentally, the keys that I am trying to localize are in Researchkit’s English dictionary, but they are not in the dictionaries localized to the other languages.

Pariecemckinney-apple:
Hello @adonawa, thank you for filing the issue. At the moment, there isn't a path to update ResearchKit's localized strings via an API. We're taking a look at a possible solution and will get back to you.

PR #1602, opened 2025-09-08
feat: Use main bundle localized key/values first if available

adonawa:
Updated the internal helper method ORKLocalizedString() to check if a string key exists in the main bundle localization tables before checking for the key in the ResearchKit localization tables.

Checks for keys in the following priority:
1. "ResearchKit" table in main bundle
2. checks the default localization table in the main bundle 
3. checks the ResearchKit bundle tables

Close: #1601


adonawa:
@Pariecemckinney-apple The potential for key collisions makes sense, I overlooked that possibility. Thank you for your response and for looking into this!

PR #1617, opened 2026-02-12
Add answer validation with NSError

beltradini:
Improved answer validation for question steps in the ResearchKit UI, ensuring that invalid answers are caught and users receive helpful feedback. The main change is the addition of a new validation method in `ORKAnswerFormat`, which is now used throughout the question step view controller to check answers before saving or continuing.
**Answer validation improvements:**
* Added a new `- (BOOL)validateAnswer:error:` method to `ORKAnswerFormat`, which checks if an answer is valid and provides an `NSError` with a descriptive message if not. [[1]](diffhunk://#diff-b6f69cc029a45409bf773da4a5d9c17a48912532da41e88221e3264ea11f689bR448-R455) [[2]](diffhunk://#diff-dc9a385b259c55b36a3afcf8528871a28b8b8a406b4ea51ef5b4b470467d1126R634-R642)
* Updated `ORKQuestionStepViewController` methods (`saveAnswer`, custom view delegate, `goForward`, and `continueAction`) to use the new validation method, showing an alert with the error message when an invalid answer is detected.

### DIFF
--- a/ResearchKit/Common/ORKAnswerFormat.m
+++ b/ResearchKit/Common/ORKAnswerFormat.m
@@ -631,6 +631,15 @@ static NSNumberFormatterStyle ORKNumberFormattingStyleConvert(ORKNumberFormattin
 - (BOOL)isAnswerValid:(id)answer {
     ORKAnswerFormat *impliedFormat = [self impliedAnswerFormat];
     return impliedFormat == self ? YES : [impliedFormat isAnswerValid:answer];
+    return YES;
+}
+
+- (BOOL)validateAnswer:(id)answer error:(NSError **)error {
+    BOOL isValid = [self isAnswerValid:answer];
+    if (!isValid && error != NULL) {
+        *error = [NSError errorWithDomain:@"ORKAnswerValid" code:1001 userInfo:@{NSLocalizedDescriptionKey: NSLocalizedString(@"The answer is not valid.", nil)}];
+    }
+    return isValid;
 }
 
 - (BOOL)isAnswerValidWithString:(NSString *)text {
--- a/ResearchKitUI/Stale/ORKQuestionStepViewController.m
+++ b/ResearchKitUI/Stale/ORKQuestionStepViewController.m
@@ -626,6 +626,12 @@ static const NSTimeInterval DelayBeforeAutoScroll = 0.25;
 }
 
 - (void)saveAnswer:(id)answer {
+    NSError *validationError = nil;
+    if (![self.answerFormat validateAnswer:answer error:&validationError]) {
+        [self showValidityAlertWithMessage:validationError.localizedDescription];
+        return;
+    }
+    
     self.answer = answer;
     _savedSystemCalendar = [NSCalendar currentCalendar];
     _savedSystemTimeZone = [NSTimeZone systemTimeZone];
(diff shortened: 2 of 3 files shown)


ResearchKit/Common/ORKAnswerFormat.m
```diff
@@ -631,6 +631,15 @@ - (ORKQuestionResult *)resultWithIdentifier:(NSString *)identifier answer:(id)an
 - (BOOL)isAnswerValid:(id)answer {
     ORKAnswerFormat *impliedFormat = [self impliedAnswerFormat];
     return impliedFormat == self ? YES : [impliedFormat isAnswerValid:answer];
+    return YES;
```
jessi-apple, line 634: This introduces unreachable code (`return YES;`) after an existing return statement. This line will never execute.

ResearchKit/Common/ORKAnswerFormat.m
```diff
@@ -631,6 +631,15 @@ - (ORKQuestionResult *)resultWithIdentifier:(NSString *)identifier answer:(id)an
...
+}
+
+- (BOOL)validateAnswer:(id)answer error:(NSError **)error {
+    BOOL isValid = [self isAnswerValid:answer];
+    if (!isValid && error != NULL) {
+        *error = [NSError errorWithDomain:@"ORKAnswerValid" code:1001 userInfo:@{NSLocalizedDescriptionKey: NSLocalizedString(@"The answer is not valid.", nil)}];
+    }
```
jessi-apple, line 641: ResearchKit already provides context-specific error messages via `localizedInvalidValueStringWithAnswerString:`, which subclasses override (see `ORKTextAnswerFormat`, `ORKNumericAnswerFormat`). This generic message doesn't leverage that existing infrastructure and provides less helpful feedback than what's already available.

Also: the error domain should be a defined constant rather than a hardcoded string, and the localized string should include a translator comment.

ResearchKitUI/Stale/ORKQuestionStepViewController.m
```diff
@@ -626,6 +626,12 @@ - (BOOL)hasAnswer {
 }
 
 - (void)saveAnswer:(id)answer {
+    NSError *validationError = nil;
+    if (![self.answerFormat validateAnswer:answer error:&validationError]) {
+        [self showValidityAlertWithMessage:validationError.localizedDescription];
+        return;
+    }
```
jessi-apple, line 633: This will cause duplicate validation when called from `customQuestionStepView:didChangeAnswer:` below, which also validates before calling `saveAnswer:`.

jessi-apple (changes requested):
Thank you for your interest in improving answer validation in ResearchKit. Unfortunately, we can't merge this PR in its current state.

 Beyond the specific issues noted inline, the main concern is that this doesn't integrate with ResearchKit's existing validation patterns. The codebase already has:
  - `isAnswerValid:` / `isAnswerValidWithString:` for checking validity
  - `localizedInvalidValueStringWithAnswerString:` for context-specific error messages
  - `shouldContinue` in answer cells for validating at navigation time

If you'd like to revisit this, I'd suggest studying how `ORKSurveyAnswerCellForText.m` and `ORKFormItemCell.m` handle validation to understand the existing patterns before proposing changes.

